PR #52, opened 2025-08-30
Ancestry Compendium Update

DoubleKing-Prime:
Adds all Ancestry options


NekroDarkmoon:
Main issue with this pr is the use of uppercase chars for folder names. These need to all be lowercase. While windows doesn't care about upper or lower case chars linux does and it might end causing some issues with some of foundry's functions.

PR #76, opened 2025-09-27
Update slot requirement calculation and remove table layout of mundane items

### LINKED ISSUE #50: Inventory Slot calculation
### Is
Inventory used slot for PCs needs an update.
Currently items can have two properties:
- stackable: boolean
- slotsRequired: number

Used Slots (apart from currency which is correctly calculated) are calculated as follows:
Add all slotsRequired, where slotsRequired is an integer (e.g. 1, 2, 3, ...).
If at least one decimal slotsRequired was found (e.g. 0.5) an additional used slot is added to the sum.
Quantity is ignored.

### Should
However here is the rules on Inventory Slots (Core Rules p. 21)
- 1 slot can hold a one-handed weapon, shield, worn armor, stack of javelins, 500 gp, or 2 potions.
- 2 slots can hold a 2-handed weapon, unworn armor, or similarly bulky items.
- Small, related items can be grouped into one slot for simplicity (e.g., Camping Supplies: soap, blanket, brush, rations)

On items it says (Core Rules p. 35)
- Javelins: Stack of 4
- Throwing Hammers: Stack of 3

### Proposed Change
To better model those rules i propose the following new item properties:
- stackable: boolean
- stackSize: number
- slotsRequired: number
- smallSize: boolean

I propose to disable quantity config for non-stackable, non-small items.

This would allow for the following calculation:
Add slotsRequired of all non-small non-stackable items (ignoring quantity).
For stackable items: quantity / stackSize than rounded up.
If at least one small item is present, add 1 slot used.

SeseFeuerherz:
Fixes #50 

Objects now have properties objectSizeType, slotsRequired and stackSize. stackable is removed.

Objects in sheets and in the item sidebar of foundry need their slots config to be updated. All new items from the compendium work out of the box.


src/models/item/ObjectDataModel.ts
```diff
@@ -22,13 +22,25 @@ const schema = () => ({
...
+		nullable: false,
+		options: ['slots', 'stackable', 'smallSized'],
+	}),
 	slotsRequired: new fields.NumberField({
 		required: true,
 		initial: 0,
 		min: 0,
 		nullable: false,
 	}),
 	stackable: new fields.BooleanField({ required: true, initial: false, nullable: false }),
```
NekroDarkmoon: This needs to be removed if it's no longer a used property.

PR #81, opened 2025-10-14
Updates to the Nimble System settings config

trevlar:
Nimble Settings

UI Settings

Optional Rules

In-Game Languages


7D7D:
@trevlar this is great stuff! Excited for these options eventually

PR #100, opened 2025-10-19
Enable dialog window resizing across app

trevlar:
This PR makes ALL dialog windows adjustable so they can be resized.

## Scene Dialog

Screenshot 2025-10-23 at 4 24 13 AM

## Actors

Screenshot 2025-10-23 at 4 35 13 AM

Screenshot 2025-10-23 at 4 23 55 AM

## Items

Screenshot 2025-10-23 at 4 22 22 AM

## Journals

Screenshot 2025-10-23 at 4 14 33 AM

Screenshot 2025-10-23 at 4 14 21 AM

## Rollable Tables

Screenshot 2025-10-23 at 4 14 03 AM

## Card Stacks

Screenshot 2025-10-23 at 4 13 49 AM

## Macros

Screenshot 2025-10-23 at 4 13 32 AM

## Playlists

Screenshot 2025-10-23 at 4 13 12 AM

## Compendium Packs

Screenshot 2025-10-19 at 11 52 58 AM

## Game Settings

Screenshot 2025-10-23 at 4 11 57 AM

### DIFF
--- a/src/settings/SystemSettings.ts
+++ b/src/settings/SystemSettings.ts
@@ -4,20 +4,6 @@ import SystemSettingsDialog from '../view/dialogs/SystemSettingsDialog.svelte';
 const { ApplicationV2 } = foundry.applications.api;
 
 export class SystemSettings extends SvelteApplicationMixin(ApplicationV2) {
-	constructor(options = {} as Configuration) {
-		super(
-			foundry.utils.mergeObject(options, {
-				svelte: {
-					component: SystemSettingsDialog,
-				},
-			}),
-		);
-
-		this.props = {
-			dialog: this,
-		};
-	}
-
 	static override DEFAULT_OPTIONS = foundry.utils.mergeObject(
 		super.DEFAULT_OPTIONS,
 		{


src/hooks/setup.ts
```diff
@@ -8,4 +8,236 @@ export default function setup() {
 	registerSystemSettings();
 
 	game.nimble.conditions.initialize();
+
+	if (foundry.applications?.sheets) {
```
NekroDarkmoon: Is there a reason why we're making sheets we don't touch resizable? Mostly because it adds more overhead and maintenance when foundry changes things.

src/scss/components/_image.scss
trevlar: I have concerns about this file because I think it breaks the image location override. I'm going to come back to this tomorrow and get it resolved.

src/hooks/setup.ts
trevlar: Additionally, I modified the code so that it only changes the DEFAULT_OPTIONS to include resizable property.

src/view/chat/components/RollSummary.svelte
```diff
@@ -39,65 +39,65 @@
...
-        width: 2.25rem;
-        height: 2.25rem;
-        padding: 0;
-        font-size: var(--nimble-lg-text);
-        color: var(--nimble-primary-color);
-        background-color: transparent;
-        border-radius: 4px;
-        border: 1px solid var(--nimble-card-border-color);
-    }
+	// .nimble-button--apply-damage {
```
trevlar: This is not used anywhere and has an error on compilation.

Issue #36, opened 2025-07-14
Cannot manually edit max HP or hit dice


Reisenkodie:
Despite increasing my max HP on the character sheet, the attributes.hp.max value does not actually change. When I close the sheet and re-open it, it's back down to its true value. I also can't even click on the hit-dice to change them, the only way to change them is to Field Rest.

<img width="805" height="248" alt="Image" src="https://github.com/user-attachments/assets/ea84032d-eb32-47b1-92bc-1c27ae93cef3" />

PR #101, opened 2025-10-19
custom hp

uiuiu-kit:
added the option to customise HP as discussed in Issue #36


src/documents/actor/character.ts
```diff
@@ -410,6 +412,32 @@ export class NimbleCharacter extends NimbleBaseActor {
...
+		const dialog = new GenericDialog(
+			`${this.name}: Configure Hit Points`,
+			EditHitPointsDialog,
+			{ document: this },
+			{ icon: 'fa-solid fa-heart', width: 250 },
+		);
+
+		await dialog.render(true);
+		const result = await dialog.promise;
+
```
NekroDarkmoon: Can just return early if no result. That way don't have to nest everything in an if block

PR #102, opened 2025-10-19
always show the feature headline

uiuiu-kit:
always show the feature headline so you can add new features without having an initial featur


uiuiu-kit:
done

PR #105, opened 2025-10-24
Fix missing subclass and wrong assignment 37

trevlar:
## Dialog when Leveling Up to 3

Screenshot 2025-10-24 at 1 06 01 PM

## Accordion is available to transition view of each subclass

Screenshot 2025-10-24 at 1 06 10 PM
Screenshot 2025-10-24 at 1 06 19 PM

## Subclass selected in view hides the other options

Screenshot 2025-10-24 at 1 06 41 PM

## Dragging subclass from compendium when subclass is already selected.

Screenshot 2025-10-24 at 1 05 27 PM

### Confirming will replace the current subclass.

Screenshot 2025-10-24 at 1 05 14 PM

### Dragging over a subclass when below level 3 results in error

Screenshot 2025-10-24 at 1 04 40 PM

### Dragging a subclass to a character doesn't have that class results in an error 

Screenshot 2025-10-24 at 1 04 29 PM

### Dragging a subclass to a character that already has that subclass results in an error.

Screenshot 2025-10-24 at 1 04 24 PM

### DIFF
--- a/src/view/dialogs/CharacterLevelUpDialog.svelte
+++ b/src/view/dialogs/CharacterLevelUpDialog.svelte
@@ -3,7 +3,6 @@
 	import getChoicesFromCompendium from '../../utils/getChoicesFromCompendium.js';
 
 	import getSubclassChoices from '../../utils/getSubclassChoices.js';
-	import replaceHyphenWithMinusSign from '../dataPreparationHelpers/replaceHyphenWithMinusSign.js';
 
 	import AbilityScoreIncrease from './components/levelUpHelper/AbilityScoreIncrease.svelte';
 	import HitPointSelection from './components/levelUpHelper/HitPointSelection.svelte';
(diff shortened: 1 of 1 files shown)


src/documents/sheets/PlayerCharacterSheet.svelte.ts
```diff
@@ -1,59 +1,206 @@
...
+		if (this.document.uuid === item.actor?.uuid) return this._onSortItem(event, itemData);
+
+		// Create the owned item
+		return this._onDropItemCreate(itemData, event);
+	}
+
+	/**
+	 * Handle sorting items within the same actor
+	 */
+	_onSortItem(event: DragEvent, itemData: any) {
```
NekroDarkmoon: Don't need to override this.

NekroDarkmoon (changes requested):

PR #108, opened 2025-10-25
Update ancestry size, added size selection on character sheet

trevlar:
Modify ancestry character size from character sheet. Hover over the metadata to reveal edit button. Clicking on the edit button will bring up the dialog where size can be changed.

When there is no size listed on the ancestry then display the playable sizes (small, medium, large)

Screenshot 2025-10-26 at 11 53 45 AM

Screenshot 2025-10-26 at 11 58 08 AM


NekroDarkmoon:
Can you take a look at the BioTab conflict and see which one you need.

PR #140, opened 2025-11-13
(fix) scrollable spells and PC inventory

trevlar:


src/view/sheets/pages/PlayerCharacterInventoryTab.svelte
trevlar: Only formatting changes in this file.

PR #143, opened 2025-11-14
(chore) Fixed all linter errors, warnings and info results. Lint check passes!

trevlar:
Fix all linter errors, warnings, and info results.

Screenshot 2025-11-14 at 10 16 19 AM

Screenshot 2025-11-14 at 10 25 48 AM


eslint.config.js
trevlar: If a variable has an underscore in front of it then it is intentionally unused.

PR #192, opened 2025-12-29
tokenizer fixes

MrTheBino:
Fixes for:

- opening tokenizer dialog for every actor type
- don't overwrite the token image if the tokenizer module is active


trevlar (approved):
LGTM

trevlar (changes requested):
Create the same changes from the dev branch instead of the main branch.

PR #205, opened 2026-01-03
Tc fix field rest dialog

trevlar:


src/view/dialogs/FieldRestDialog.svelte
```diff
@@ -40,145 +34,274 @@
...
 					type="radio"
 					name="{actor.uuid}-field-rest-type"
 					value={false}
 					bind:group={makeCamp}
 				/>
+				<div class="rest-type-card__header">
+					<i class="rest-type-card__icon fa-solid fa-wind"></i>
+					<span class="rest-type-card__title">Catch Breath</span>
+				</div>
+				<p class="rest-type-card__description">10 min rest. Roll Hit Dice + STR to heal.</p>
```
trevlar: We need to internationalize these.

Issue #48, opened 2025-08-17
actions get reset when starting combat


Reisenkodie:
Nimble has a unique initiative system where you only roll initiative to determine how many actions you have in the first round of combat, and once your turn ends, you get your actions reset.

When opening the combat tracker, players already have the option of rolling their Initiative, which sets their actions. However, if they do this before Start Combat has been clicked, then clicking Start Combat resets their actions to 3 as if a new round has begun.

PR #213, opened 2026-01-09
Fix actions being reset on combat start

Fronix:
- Modified `_onEndRound()` to skip actions reset during the first turn of the first round, fixing #48 
- Added `startCombat()` override in `NimbleCombat` to roll initiative for character combatants lacking it at combat start
- Ensure `CombatTracker` updates correctly when state changes

Comment on the package-lock update, this happened as i did `npm i`, I can remove the commit if it's unwanted but I needed it for tsconfig not to give me errors about `fvtt-types`. It seems to only have added `dev: true` to packages that are only dev dependancies.

### DIFF
--- a/src/view/ui/CombatTracker.svelte
+++ b/src/view/ui/CombatTracker.svelte
@@ -190,16 +190,16 @@
 	});
 </script>
 
-{#key version}
-	{#if currentCombat}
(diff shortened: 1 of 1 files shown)


trevlar (approved):
LGTM

PR #235, opened 2026-01-14
Show when primary die/modifier is set

### LINKED ISSUE #128: Set Primary Die
We need way to see when someone sets the Primary Die in the chat. This way the GM knows that the Cheat is using that ability or that a player is fudging their die rolls. Otherwise maybe it should be limited to the GM for GM Fiat/ Fudging for the story.

Image

Fronix:
### Added
- Adds primary die details to damage roll summary

### Fixed
- Fixes disabled field not being set correctly in `ItemActivationConfigDialog`
- Fixes lint command syntax in package.json

### Examples
image

Fixes #128

### DIFF
--- a/src/view/dialogs/ItemActivationConfigDialog.svelte
+++ b/src/view/dialogs/ItemActivationConfigDialog.svelte
@@ -8,7 +8,7 @@
 	let primaryDieValue = $state();
 	let primaryDieModifier = $state();
 
-	const { damageTypes } = CONFIG.NIMBLE;
+	const { damageTypes, hitDice } = CONFIG.NIMBLE;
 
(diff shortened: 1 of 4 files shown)


src/view/dialogs/ItemActivationConfigDialog.svelte
```diff
@@ -82,27 +82,27 @@
 	<div class="nimble-roll-modifiers-container">
 		<div class="nimble-roll-modifiers">
 			<label>
-				situational modifiers:
+				Situational modifiers:
```
trevlar: Would you add these to i18n too?

trevlar (approved):

PR #262, opened 2026-02-01
Spell Compendium Organization and Filtering

7D7D:
Updated the Spell Compendium view so spells are sorted into tier level and by school.

**Each spell has two tags in its data:**
- system.school (fire, ice, lightning, etc.)
- system.tier (0 = Utility, 1 = Cantrip, 2 = Tier 1, 3 = Tier 2, …)

**Filtering by a school button:**
- Only spells whose system.school matches that button are shown.

**Grouping into tiers:**
- Spells are grouped by system.tier.
- The tiers are ordered: Utility (0), Cantrip (1), then Tier 1 (2), Tier 2 (3), etc.

**Inside each tier:**
- Spells are grouped by school in a fixed order (Fire → Ice → Lightning → Necrotic → Radiant → Wind → Secret → Utility).
- Within each school group, spells are listed alphabetically by name.
- So the only thing that decides a spell’s school and tier is its system.school and system.tier fields in the spell’s data.

 **Added filter buttons for each school.**

**Added clear all button**

All filter
Default list
Fire filter
Ice filter
Lightning filter
Necrotic filter
Radiant filter

### DIFF
--- a/src/hooks/compendiumSpellsFilter.ts
+++ b/src/hooks/compendiumSpellsFilter.ts
@@ -4,10 +4,6 @@
  */
 
 const GLOBAL_NIMBLE_SPELLS_COLLECTION = 'nimble.nimble-spells';
-let activeSchoolFilter = '';
-let isApplyingFilter = false;
-let cachedSpellDataById: Map<string, { school: string; tier: number }> | null = null;
-let cachedSpellNameById: Map<string, string> | null = null;
 
 const SCHOOL_KEYS = [
 	'fire',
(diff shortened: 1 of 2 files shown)


trevlar (changes requested):
Will you adjust the PRs to keep the changes focused on the issue being resolved. Please don't make me scroll through 200 files of irrelevant changes. You can add those in a separate PR if you must.

trevlar (dismissed):

7D7D:
Here’s a summary of what I changed.

1. Removed debug logging from nimble.js and kept the side-effect import.
2. Added a clarifying comment in nimble.js explaining that nimble.ts'; is intentional for system bootstrapping.
3. Refactored spell filter state in compendiumSpellsFilter.ts from module-level globals to per-compendium-window state using `WeakMap`.
4. Updated filter/search/render flows to pass and use that per-window state everywhere (`applyFilter`, button setup, render hook, icon metadata cache).
5. Fixed `TIER_LABELS` typing mismatch by changing it to `Record<string, string>` with string keys.
6. Extracted repeated selectors into constants (spell item selector, header selector, search selector).
7. Replaced many inline styles with CSS classes in injected styles, leaving only dynamic color values as CSS custom properties.
8. Added debounced metadata application for `MutationObserver` updates so repeated DOM mutations are batched.
9. Updated tooltip handling:
   - plain text tooltips now use `data-tooltip-text`
   - localization key tooltip for “All” uses `data-tooltip`
   - added aria-label for clear/all/school buttons
10. Replaced collapse-button hijacking with a dedicated clear button control.
11. Added ordering logic so header controls are arranged as: `Search `-> `Clear All Filters` -> `Sort Alphabetically` -> `Collapse All Folders`.
12. Adjusted the clear button style to visually match sort/collapse icon buttons (no border/background/shadow across states).

Screenshot 2026-02-09 200345


NOTE: The "Collapse All Folders" button and functionality have been returned but do not affect the Spell Compendium in any way. I can not see a need for it at present. So let me know if we should keep it for the future or remove it.

src/hooks/compendiumSpellsFilter.ts
```diff
@@ -0,0 +1,1246 @@
...
+
+		// Default case
+		return {
+			_id: e._id,
+			system: {
+				school: e.system?.school || '',
+				tier,
+				propertiesSelected,
+			},
+		};
```
trevlar: These can all be consolidated.


```suggestion
		const school = e.system.school ?? e.school || e.system?.school || ''

		return {
			_id: e._id,
			system: {
				school,
				tier,
				propertiesSelected,
			},
		};
```

trevlar (approved):
LGTM, One small thing. Though I'm interested in getting this merged.

PR #264, opened 2026-02-01
Feature/minions not missing

7D7D:
### Changes Made

- Adjusted damage roll preprocessing so misses are detected even when crits are disabled, and enforced minion damage roll options to match the minion rules; added a minion-specific test.

- DamageRoll now creates a primary die when either `canCrit` or `canMiss` is true, and only adds the explosion modifier when `canCrit` is true. This lets minions (canCrit=false, canMiss=true) miss on a 1 without ever exploding. DamageRoll.ts

- Minion actors now force `canCrit=false` and `canMiss=true` at roll time to centralize the minion rule even if an item is misconfigured. ItemActivationManager.ts

### Previous Issue Reason

- The attack flow for monsters goes through ItemActivationManager, which flattens activation effects and creates a DamageRoll for the first damage node, passing canCrit/canMiss from that node. ItemActivationManager.ts
- DamageRoll only builds a PrimaryDie when canCrit is true; isMiss is derived from that PrimaryDie. When canCrit is false, PrimaryDie is never created, so isMiss stays undefined, and the chat message defaults to “hit.” DamageRoll.ts and chatMessage.ts
- Minion compendium entries already set canCrit: false and canMiss: true on their damage nodes, so they fall into this gap and never show a miss on a roll of 1. goblin-minion.json (similar in bandit-minion.json, kobold-minion.json)
- Criticals are implemented by adding an exploding modifier (x) to the primary die. That both flags crits and adds extra damage on max. For minions, we want the primary die for miss detection but no explosion. DamageRoll.ts
- The “single damage die” part of the minion rule is currently data-driven; minion entries in the packs already use 1dX formulas and mention “follows minion rules,” but the system doesn’t enforce formula shape in code.

### DIFF
--- a/src/managers/ItemActivationManager.ts
+++ b/src/managers/ItemActivationManager.ts
@@ -122,7 +122,7 @@ class ItemActivationManager {
 					const isMinion =
 						actorTags?.has('minion') ?? (this.actor?.type as string | undefined) === 'minion';
 					const resolvedCanCrit = isMinion ? false : (canCrit ?? true);
-					const resolvedCanMiss = isMinion || canMiss ?? true;
+					const resolvedCanMiss = isMinion || (canMiss ?? true);
 					node.rollMode = dialogData.rollMode ?? 0;
 
 					// Use modified formula if provided


7D7D:
Updated with your code suggestions and a small fix "Cannot use "??" with "||" without parentheses" error.

trevlar (approved):

PR #271, opened 2026-02-04
Hide NPC Rolls

leadanymind:
- **Adds ability to hide rolls for npc skill checks and damage rolls**
- **Adds system settings checkbox for default value for hiding rolls as gm rolls**


src/view/dialogs/CheckRollDialog.svelte
```diff
@@ -16,15 +17,25 @@
 
 <article class="nimble-sheet__body" style="--nimble-sheet-body-padding-block-start: 0.5rem">
 	<RollModeConfig bind:selectedRollMode />
-
+	<div class="nimble-roll-modifiers-container">
+		<label>
+			Hide Roll?
```
trevlar: Will you add this to en.json and connect it in config.ts? So that we have translation options for it.

trevlar (approved):

PR #278, opened 2026-02-08
fix: correct action refresh timing for PCs and NPCs

trevlar:
## Summary
- Fixed PC actions incorrectly refreshing at round change (should only refresh at end of their turn)
- Fixed NPC/monster action pips starting empty in round 1 (now initialized when combat starts)
- `_onEndRound` now only refreshes non-character combatants (NPCs, minions, solo monsters)

## Test plan
- [ ] Start combat with PCs and NPCs
- [ ] Verify NPC action pips are full at the start of round 1
- [ ] Use a heroic reaction on a PC to reduce their actions
- [ ] Advance to the next round and verify the PC's reduced actions carry over (not reset at round change)
- [ ] Verify PC actions refresh at the end of their turn
- [ ] Verify NPC actions refresh at the end of each round

### DIFF
--- a/src/documents/combat/combat.svelte.ts
+++ b/src/documents/combat/combat.svelte.ts
@@ -73,6 +73,24 @@ class NimbleCombat extends Combat {
 			);
 		}
 
+		// Initialize actions for non-character combatants (NPCs, minions, solo monsters)
+		// Characters get their actions set during initiative roll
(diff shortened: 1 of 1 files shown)


Fronix (approved):
Ah, that's what I missed!

LGTM!

PR #307, opened 2026-02-13
Spellblade Mana edge case during combat

7D7D:
Implemented behavior for the Commander Spellblade edge case across combat logic, cleanup, and sheet rendering.

### What we changed

- Added Spellblade detection helpers in combat.svelte.ts (line 17) and nimble.ts (line 13).
- Added initiative-time mana grant in combat.svelte.ts (line 209):
    - Trigger condition: combatant is a character, has initiative === null (first roll), and is Commander+Spellblade.
    - Effect: set system.resources.mana.baseMax and system.resources.mana.current to max(0, INT mod).
- Added combat-end cleanup only on actual combat deletion in nimble.ts (line 89):
    - On deleteCombat, clear Spellblade mana to 0 via nimble.ts (line 30).
    - This means canceling the end-combat modal does not clear mana.
- Removed pre-confirmation cleanup path (no early clear on endCombat() call).
- Fixed sheet visibility/race issues in PlayerCharacterSheet.svelte:
    - Added robust combat resolution (game.combat first, then active combat in contents, then viewed) in PlayerCharacterSheet.svelte (line 53).
    - Added hook-driven reactivity for open sheets in PlayerCharacterSheet.svelte (line 151).
    - Mana bar visibility for Spellblade now depends on active started combat + rolled initiative in that combat in PlayerCharacterSheet.svelte (line 230).
    - Final render gate is still hasMana at PlayerCharacterSheet.svelte (line 460).

### How the edge-case rule is now triggered and managed
1. Combat starts or initiative is rolled.
2. When the Spellblade character rolls initiative for the first time in that combat, mana is granted equal to current INT modifier.
3. Mana bar appears only after that roll, while combat is active.
4. If combat ends and is confirmed (“Yes”), combat is deleted, deleteCombat fires, and Spellblade mana is cleared.
5. If end-combat is canceled, no deletion hook runs, so mana is not cleared.

Screenshot 2026-02-13 150244

Screenshot 2026-02-13 150257

Screenshot 2026-02-13 150315

Screenshot 2026-02-13 150331

Screenshot 2026-02-13 150346


src/documents/combat/combat.svelte.ts
```diff
@@ -177,6 +206,21 @@ class NimbleCombat extends Combat {
...
+				combatant.initiative === null &&
+				this.#isCommanderSpellblade(combatant);
+
+			if (shouldGrantSpellbladeMana && combatant.actor?.isOwner) {
+				const spellbladeMana = this.#getSpellbladeManaFromInt(combatant);
+				spellbladeManaUpdates.push(
+					combatant.actor.update({
+						'system.resources.mana.baseMax': spellbladeMana,
+						'system.resources.mana.current': spellbladeMana,
+					} as Record<string, unknown>),
```
trevlar: Generally, game logic, such as character abilities, should be defined as rules in JSON feature packs, not hardcoded in the system.

trevlar (changes requested):
The subclass has an empty rules array, yet the description clearly states: "you now gain INT mana when you roll Initiative (this mana is lost if unspent when combat ends)"                                    
                                                                                                                                                                                                                 
  This is exactly the kind of behavior that should be defined as a rule on the subclass item, not hardcoded in the combat system and character sheet. The codebase already has a robust rules system with types  
  like:                                                                                                                                                                                                          
  - maxHpBonus                                                                                                                                                                                                   
  - speedBonus                                                                                                                                                                                                   
  - initiativeBonus                                                                                                                                                                                              
  - maxHitDice                                                                                                                                                                                                   
  - etc.                                                                                                                                                                                                         
                                                                                                                                                                                                                 
  Recommended Approach                                                                                                                                                                                           
                                                                                                                                                                                                                 
  Instead of the current implementation, create a new rule type like combatMana or initiativeMana:                                                                                                               
                                                                                                                                                                                                                 
  // packs/subclasses/core/commander/spellblade.json                                                                                                                                                             
  {                                                                                                                                                                                                              
    "rules": [                                                                                                                                                                                                   
      {                                                                                                                                                                                                          
        "type": "combatMana",                                                                                                                                                                                    
        "id": "spellblade-combat-mana",                                                                                                                                                                          
        "label": "Arcane Command - Combat Mana",                                                                                                                                                                 
        "formula": "max(0, @intelligence.mod)",                                                                                                                                                                  
        "trigger": "initiativeRoll",                                                                                                                                                                             
        "clearOn": "combatEnd"                                                                                                                                                                                   
      }                                                                                                                                                                                                          
    ]                                                                                                                                                                                                            
  }                                                                                                                                                                                                              
                                                                                                                                                                                                                 
  Then the combat system would check for items with combatMana rules instead of hardcoding identifier === 'spellblade' && parentClass === 'commander'.                                                           
                                                                                                                                                                                                                 
  Benefits:                                                                                                                                                                                                      
  - Data-driven (defined in compendium, not code)                                                                                                                                                                
  - Extensible (any future subclass/item can use it)                                                                                                                                                             
  - No duplicated detection logic across 3 files                                                                                                                                                                 
  - GMs/modders can see and modify the behavior                                                                                                                                                                  
  - Unit testable

7D7D:
Ready for review

###  Latest Changes

- Added a new rule type combatMana so combat-only mana is data-driven instead of hardcoded.
- Registered combatMana in rule config and added i18n labels (en, es, fr).

- Added Spellblade subclass rule in spellblade.json:
  - type: "combatMana"
  - formula: "max(@intelligence, 0)"
  - trigger: "initiativeRoll"
  - clearOn: "combatEnd"
 
- Implemented shared combat-mana rule utilities in combatManaRules.ts:
  - Collect actor combat-mana rules.
  - Resolve rules from embedded item data and compendium source.
  - Compute initiative mana grant total.
  - Read/write combat grant tracking map via flags.nimble.combatManaGrants.
 
- Updated combat flow in combat.svelte.ts:
  - On initiative roll, evaluates combatMana rules and grants mana.
  - Stores per-combat grant state so cleanup is scoped correctly.
  - Removed subclass/class hardcoding.
 
- Updated cleanup in nimble.ts:
  - On combat end/delete, clears only mana tied to that combat grant entry.
 
- Updated sheet visibility logic in PlayerCharacterSheet.svelte:
  - For initiative-based combat mana, mana bar shows only when:
    - active combat is started,
    - actor has rolled initiative,
    - and granted mana is present.
  - Live updates when combat state changes.
  - Temporary debug/fallback hardening was removed for final clean behavior.
  
- Added tests in combatManaRules.test.ts covering rule collection, source fallback/priming, and grant calculations.

Screenshot 2026-02-14 145824
Screenshot 2026-02-14 145839

Fronix (commented):
What is the reason for assuming all characters should be checked for `combatMana`?

src/documents/combat/combat.svelte.ts
```diff
@@ -177,6 +184,34 @@ class NimbleCombat extends Combat {
...
+						combatManaUpdates.push(
+							combatant.actor.update({
+								'system.resources.mana.baseMax': combatMana,
+								'system.resources.mana.current': combatMana,
+								'flags.nimble.combatManaGrants': grants,
+							} as Record<string, unknown>),
+						);
+					}
+				}
+			}
```
Fronix: There are probably more cases of "only in combat". I suggest moving this to a `handleInitiativeRules` which we can put in a separate file under `combat` as to not clutter the `combat.svelte.ts` file.

```
combat/
  combat.svelte.ts
  handleInitiativeRules.ts
  handlers/
```

handleInitiativeRules.ts
```typescript
import combatManaHandler from './handlers/combatManaHandler.ts'

export function handleInitiativeRules() {
 combatManaHandler();
 // other handlers
}
```

7D7D:
> Collaborator

That's a good question. How should this be handled? Do we have no checks and just hard code that this is the only subclass with that feature? Then we can add others to that list if needed in the future.

trevlar (changes requested):
Merge conflict on this one.

PR #332, opened 2026-02-22
Improve Active Effects and Conditions display on actor sheets and token HUD

### LINKED ISSUE #330: Show applied conditions/effects in sheet and canvas
Right now it's a bit tricky to see if an actor has a condition. We should have a tab in the character sheet where you can see the current conditions, effects and passive effects. When a actor is selected in the canvas they should show up in the top right corner.

Fronix:
This PR improves how Conditions and Active Effects are shown and managed in Foundry:

**Important:** Need feedback on the UI here

- Adds clearer Condition display on actor sheets and on-canvas HUD
- Adds rich hover tooltips with duration, source, and description
- Shows compact duration badges on HUD icons
- Keeps standard Conditions separate from non-condition Active Effects
- Lets users toggle Conditions from the sheet and remove temporary Active Effects
- Improves readability and theming in both light and dark modes

## Test Plan
- [ ] Select a token with active Conditions and verify HUD icons render
- [ ] Hover a HUD icon and verify tooltip shows name, duration, source, description
- [ ] Right-click a standard Condition icon on HUD and verify it is removed
- [ ] Right-click a non-condition Active Effect icon on HUD and verify the effect is deleted
- [ ] Verify no HUD content shows when selected token has no active statuses/effects
- [ ] Open actor sheet and verify Temporary Effects and Passive Effects are listed separately
- [ ] Verify standard Conditions are listed only in the Conditions catalog
- [ ] Toggle a Condition on/off in the catalog and verify actor state updates
- [ ] Remove a Temporary Effect from sheet and verify it is deleted
- [ ] Verify search and active-only filter work in Conditions catalog
- [ ] Verify active Conditions are visually distinct in the catalog
- [ ] Verify tooltip uses full words for Rounds/Turns (not `r`/`t`)
- [ ] Verify UI readability and contrast in both Foundry light and dark themes
- [ ] Verify catalog scrollbar does not overlap controls/content

## Examples
<div style="display: flex; flex-direction: column;">
image
image
</div>

<div style="display: flex; flex-direction: column;">
image
image
</div>

image

---

Use this to add test effects

```javascript
const actor = canvas.tokens.controlled[0]?.actor;
if (!actor) throw new Error("Select a token first.");

await actor.createEmbeddedDocuments("ActiveEffect", [
  {
    name: "Test Effect 45s",
    img: "icons/svg/aura.svg",
    statuses: ["test-seconds"],
    duration: { seconds: 45 }
  },
  {
    name: "Test Effect 10m",
    img: "icons/svg/aura.svg",
    statuses: ["test-minutes"],
    duration: { seconds: 600 }
  },
  {
    name: "Test Effect 2h",
    img: "icons/svg/aura.svg",
    statuses: ["test-hours"],
    duration: { seconds: 7200 }
  },
  {
    name: "Test Effect 3r",
    img: "icons/svg/aura.svg",
    statuses: ["test-rounds"],
    duration: { rounds: 3 }
  },
  {
    name: "Test Effect 2t",
    img: "icons/svg/aura.svg",
    statuses: ["test-turns"],
    duration: { turns: 2 }
  }
]);
```


public/lang/en.json
```diff
@@ -624,7 +624,25 @@
...
+			"unlimited": "Unlimited",
+			"source": "Source",
+			"sourceNone": "None",
+			"noConditions": "No conditions",
+			"noEffects": "No effects",
+			"temporaryEffects": "Temporary Effects",
+			"passiveEffects": "Passive Effects",
+			"searchConditions": "Search conditions",
+			"showActiveOnly": "Show Active Only",
+			"showAllConditions": "Show All Conditions",
```
trevlar: Show All

trevlar (changes requested):
I like where this is going. A few things that could be adjusted. 

1. the Search box and button are not the same height which doesn't look great.
2. Some of the conditions are outside their boundary. Maybe we have the text ellipses if the dialog is smaller than the text. Hovering over the boxes show the tooltip to describe it anyways.
3. Instead of Show Active Only and Show All Conditions Let's have an icon to represent active. Or just have a button with the label Active and All. Don't have the button change width keep it enough text to always fit the same width. Makes it a better UX if the button isn't moving when you click it.
4. PC Sheet has double scrollbars which doesn't look great and is a bit confusing.

Important Issues (should fix)

1. Redundant/unusual reactivity pattern in ActorConditionsList.svelte:20-28:
```typescript
let effectTick = $derived(effectVersion);
// ...
return effectTick >= 0 ? conditions : conditions;
```
1. This pattern is awkward and confusing. The ternary always returns conditions regardless of the check. This appears to be a workaround to
force reactivity, but there are cleaner approaches.
2. Error notification hardcoded in English in ActorConditionsList.svelte:104,113,126:
```typescript
ui.notifications.error('Failed to remove condition.');
ui.notifications.error('Failed to update condition.');
ui.notifications.error('Failed to remove effect.');
```
2. These should use the localize() function for consistency with the rest of the codebase.
3. Type assertion pattern in ActorConditionsList.svelte:43-47:
```typescript
const typedEffect = effect as { isTemporary?:
boolean };
const duration = (effect as { duration?: {
  remaining?: number; rounds?: number }
}).duration;
```
3. This loses type information. Consider creating a proper type for the effect interface to avoid repeated inline type assertions.

---
Suggestions (nice to have)

1. Consider extracting duration formatting logic
  - The compactSecondsLabel,
deriveDurationDetails, and tooltip expansion logic in prepareActorConditions.ts could be extracted to a shared utility if this pattern is used elsewhere.
2. Magic numbers in prepareActorConditions.ts:32-33:
```typescript
const ROUND_SECONDS = 6; const TURN_SECONDS = 6;
```
These might be better imported from a shared config if they're used elsewhere in the codebase.
3. Consider adding keyboard navigation for the condition catalog grid - currently only mouse/touch interactions are supported.
4. Scrollbar styling - The scrollbar-gutter: stable is good, but the scrollbar itself isn't styled to match the UI theme in _actor-conditions.scss.

trevlar (changes requested):
Merge Conflict

trevlar (changes requested):
the hanging buttons are cut off in the character sheet frame.

Image
Image

trevlar (changes requested):
Conflicts on translations

trevlar (approved):
Other than the MR Conflict looks great!

trevlar (changes requested):
MR Conflict

trevlar (approved):

PR #334, opened 2026-02-23
fix(items): fix vicious weapon/spell mechanics and roll dialog display

trevlar:
## Summary
- Exclude crit-only damage from the roll dialog display so vicious weapons don't show duplicate dice
- Implement proper vicious property for Dravok's Terrible Maw (4 separate damage effects, each can crit independently)
- Fix Razor Wind vicious damage structure

## Test plan
- [ ] Verify Sling and Sickle no longer show duplicate damage dice in roll dialog
- [ ] Test Dravok's Terrible Maw - each of the 4 d4s should be able to crit independently and add a vicious d4
- [ ] Test Razor Wind - should add vicious d4 on crit, vicious die should not crit itself

### DIFF
--- a/packs/legendaryMonsters/core/dravok-all-seeing-tyrant.json
+++ b/packs/legendaryMonsters/core/dravok-all-seeing-tyrant.json
@@ -791,7 +791,7 @@
 							"id": "SqqL7WHNreF474eJ",
 							"type": "damage",
 							"damageType": "piercing",
-							"formula": "1d4x+1d4x+1d4x+1d4x",
+							"formula": "1d4",
 							"parentContext": null,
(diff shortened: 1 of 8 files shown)


Fronix (approved):
@trevlar 

## Test plan
### ✅ Verify Sling and Sickle no longer show duplicate damage dice in roll dialog

<div style="display: flex; flex-direction: column;">
image
image
</div>

### ✅ Test Dravok's Terrible Maw - each of the 4 d4s should be able to crit independently and add a vicious d4
image

### ✅ Test Razor Wind - should add vicious d4 on crit, vicious die should not crit itself
image

PR #344, opened 2026-02-24
Spanish translation

carlosvgs:
Spanish translation


Fronix:
@carlosvgs have you run `npm run lang:reconcile` and then `npm run lang:detect-untranslated`?

Fronix:
> hello! Nope, I haven't, just translated the file to Spanish and then uploaded it here

Please check the `es.json` file that should be the spanish translation file, and after ensure you run those commands, it will check if any keys are missing.

trevlar (changes requested):
Should be in es.json not esp.json

PR #361, opened 2026-02-28
feat(hooks): add item activation hooks and chat message flags

### LINKED ISSUE #347: Add support for AutomatedAnimations
## Module support
Add support for the module AutomatedAnimations, maintained by theripper93 who keeps things nice and updated so a pretty safe module to have support for.

**Documentation: https://wiki.theripper93.com/free/autoanimations#adding-system-support**

### LINKED ISSUE #353: [Module Support] Expose a useItem hook for Modules
### Problem
As of right now there is no real possiblity to hook on a useItem event, which means our only way to react on those events is to read the chat and parse events from there.

### Solution
Expose a global useItem event broadcast so modules can listen in.

Possibly connected to #347 as that would probably need the same effort there.

Fronix:
## Hooks
Exposes two new system hooks on the item activation pipeline:

- `nimble.preUseItem`: Fires before an item is used (after rolls are resolved, before chat message creation and mana deduction). Listeners can return false to cancel the activation.
- `nimble.useItem`: Fires after an item has been used and the chat message has been created. Non-cancellable.
Both hooks provide the item, rolls, critical/miss state, and current targets. The spell override additionally includes upcast data.

## Chat message flags
Item activation chat messages now include flags.nimble with itemId, itemUuid, actorId, and tokenUuid. This enables modules like [AutomatedAnimations](https://github.com/theripper93/autoanimations) to discover the source item/token/actor from chat messages using their [generic system handler](https://github.com/theripper93/autoanimations/blob/main/src/system-support/aa-chatmessage.js).

Closes #353, #347 

## Note
This should give basic access for most modules, if there are more hooks that need to be added we will have to take those case by case, unless anyone knows of more hooks that we need to expose.

### DIFF
--- a/src/documents/item/base.svelte.ts
+++ b/src/documents/item/base.svelte.ts
@@ -199,6 +199,14 @@ class NimbleBaseItem<ItemType extends SystemItemTypes = SystemItemTypes> extends
 				style: CONST.CHAT_MESSAGE_STYLES.OTHER,
 				sound: CONFIG.sounds.dice,
 				rolls,
+				flags: {
+					nimble: {
+						itemId: this.id,
+						itemUuid: this.uuid,
+						actorId: this.actor?.id,
+						tokenUuid: this.actor?.token?.uuid,
+					},
+				},
 				system: {
 					actorName: this.actor?.name ?? '',
 					actorType: this.actor?.type ?? '',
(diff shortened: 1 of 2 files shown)


trevlar (approved):
LGTM

Fronix:
I'll merge this, if any new hooks are needed we will address those separately.

PR #373, opened 2026-02-28
fix(sheets): improve edit mode toggle

### LINKED ISSUE #203: Character sheet editing and viewing is finicky
I'm clicking around my character sheet and oops accidentally clicked on an edit box. Oops accidentally changed the name. It seems like it's always editable which is nice I am in the habit of clicking on things I read and it causes dialog box after dialog box to come up. A better design would include a toggle which will mark the sheet as read only or edit mode.

Fronix:
## Summary

Improves edit mode toggle for character sheets to address issue #203, where users could accidentally modify sheets by clicking on editable elements while reading.

- Change `editingEnabled` flag default from `true` to `false`, making sheets read-only by default
- Hide edit/delete/create buttons when not in edit mode via conditional rendering
- Improve visual distinction of the edit toggle button with enhanced styling (disabled state colors and transitions)
- Update components to derive `editingEnabled` state from actor flags instead of context store
- Refactor ActorConditionsList to properly use reactive actor reference

image
image

## Changes

- Character sheets are now **read-only by default**. Users must toggle edit mode to make changes.
- `editingEnabled` flag now defaults to `false` instead of `true`
- Edit/delete/create buttons are no longer visible when editing is disabled (previously shown but disabled)

---

Fixes #203

### DIFF
--- a/src/view/sheets/PlayerCharacterSheet.svelte
+++ b/src/view/sheets/PlayerCharacterSheet.svelte
@@ -780,11 +780,6 @@
 		}
 	}
 
-	.nimble-button--hidden {
-		visibility: hidden;
-		pointer-events: none;
-	}
-
 	.nimble-heading--hp {
 		--nimble-button-icon-y-nudge: 0;
 
--- a/src/view/sheets/components/MovementSpeed.svelte
+++ b/src/view/sheets/components/MovementSpeed.svelte
@@ -47,23 +47,17 @@
 	<section class="nimble-movement" style="grid-area: speed;">
 		<header class="nimble-section-header nimble-movement__primary">
 			<h3 class="nimble-heading" data-heading-variant="section">Speed</h3>
-			<span class="nimble-movement__value">{walkSpeed}</span>
-			{#if editingEnabled}
-				<button
-					class="nimble-button"
-					data-button-variant="icon"
-					type="button"
-					aria-label={configTooltip}
-					data-tooltip={configTooltip}
-					onclick={() => actor.configureMovement()}
-				>
-					<i class="fa-solid fa-edit"></i>
-				</button>
-			{/if}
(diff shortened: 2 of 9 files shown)


src/view/sheets/components/MovementSpeed.svelte
```diff
@@ -88,9 +87,7 @@
 	}
 
 	.nimble-movement__primary {
-		display: flex;
-		align-items: baseline;
-		gap: 0.5rem;
+		margin-block-end: 0;
 	}
```
trevlar: This breaks the movement view when there are multiple move speeds...

Before Changes:

Image

After changes:

Image

src/view/sheets/PlayerCharacterSheet.svelte
```diff
@@ -396,25 +396,37 @@
...
-                    --nimble-actor-image-y-offset: {actorImageYOffset}px;
-                    --nimble-actor-image-scale: {actorImageScale}%;
-                "
+					--nimble-actor-image-x-offset: {actorImageXOffset}px;
+					--nimble-actor-image-y-offset: {actorImageYOffset}px;
+					--nimble-actor-image-scale: {actorImageScale}%;
+				"
 			/>
-		</button>
+		{/if}
```
trevlar: I think it would be nice to just disable the button and use an empty tooltip/aria-label when editing is disabled. It uses specific settings on the config tab to set the image position.

trevlar (changes requested):
These tab button icons on the sheet are not centered when in editing mode

Image

src/view/sheets/components/MovementSpeed.svelte
Fronix: Oh right, it shows more than that, totally forgot, good catch!

PR #375, opened 2026-03-01
feat: add Apply Healing button and healingPotionBonus rule (#368)

### LINKED ISSUE #368: Add apply healing button
Needs a button to apply healing
Image

trevlar:
## Summary

- Add "Apply Healing" button to healing nodes in chat messages (healing potions, healing spells, healing abilities)
- Create `healingPotionBonus` rule type for Field Medic's "Roll 1 additional die for any health potion you administer"
- Configure healing effects for class features that were missing them
- Add migrations to update existing characters with new rules and effects

### Core Apply Healing Feature
- `NimbleChatMessage.applyHealing()` - New method to apply healing to targets
- `HealingNode.svelte` - Added "Apply Healing" button with styling
- Added tests for `applyHealing` method

### Healing Potion Bonus Rule
- `healingPotionBonus.ts` - New rule type that adds bonus dice to healing potions
- `ItemActivationManager.ts` - Applies bonus dice when activating healing potions
- Field Medic feature updated with the new rule

### Compendium Feature Updates
- **Field Medic** - Added `healingPotionBonus` rule
- **Healing Salve** - Added healing effect `(@will)d6`
- **Lay on Hands** - Added healing effect `5*@level`
- **Searing Light** - Added healing `(@will)d8` and damage `(@will)d8` effects

### Migrations
- `Migration008FieldMedicHealingPotionBonus` - Updates existing Field Medic features
- `Migration009HealingEffects` - Updates existing Healing Salve, Lay on Hands, and Searing Light features

## Test plan

- [x] Verify "Apply Healing" button appears on healing potions when used
- [x] Verify "Apply Healing" button appears on healing spells when cast
- [x] Verify Field Medic grants +1 die to healing potions
- [x] Verify healing is applied correctly to selected targets
- [ ] Verify migrations update existing characters correctly

## Related

Closes #368

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -44,6 +44,7 @@
 		},
 		"chat": {
 			"applyHealing": "Apply Healing",
+			"healing": "Healing",
 			"healingApplied": "Healing Applied",
 			"undoHealing": "Undo Healing",
 			"undo": "Undo",
--- a/src/documents/chatMessage.ts
+++ b/src/documents/chatMessage.ts
@@ -8,7 +8,7 @@ import { getRelevantNodes } from '#view/dataPreparationHelpers/effectTree/getRel
 type ActivationCardTypes = 'feature' | 'minionGroupAttack' | 'object' | 'spell';
 
 /** Record of applied healing for undo functionality */
-interface AppliedHealingRecord {
+export interface AppliedHealingRecord {
 	effectId: string;
 	healingType: string;
 	amount: number;
@@ -349,6 +349,11 @@ class NimbleChatMessage extends ChatMessage {
 		}
 	}
 
+	/**
+	 * Reverts previously applied healing by restoring HP to the snapshot taken at apply time.
+	 * Note: This is snapshot-based - if something else modified HP between apply and undo,
+	 * those changes will be silently overwritten when reverting to the previous values.
+	 */
 	async undoHealing(effectId: string): Promise<void> {
 		if (!this.isActivationCard()) return;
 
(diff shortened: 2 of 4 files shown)


Fronix (dismissed):
Overall great, just some minor stuff that might be an issue.

### Potential issues

**1. The "Healing" label isn't localized**
`HealingNode.svelte:73` has a hardcoded `<h3>Healing</h3>` while everything else in that component goes through `localize()`.

**2. Inline SVG instead of Font Awesome**
The checkmark icon for "Healing Applied" (`HealingNode.svelte:88-99`) is a raw SVG path. Any reason this isn't just `fa-solid fa-check`?

**3. `isHealingItem` is misleadingly named**
In `ItemActivationManager.ts:213`, `objectType === 'consumable'` flags ALL consumables, not just healing potions. The bonus only actually applies to healing nodes (gated at line 252), but the variable name suggests something that isn't true. Rename to `isConsumable` or drop the outer check.

**4. Duplicate type definition**
`AppliedHealingRecord` is defined identically in both `chatMessage.ts:11-24` and `HealingNode.svelte:6-19`. If one changes, the other silently drifts. Export it from `chatMessage.ts` and import it.

**5. `getAppliedHealingRecord` is unused**
`chatMessage.ts:398-402` — defined but never called anywhere. Remove it or add a comment explaining planned use.

**6. Type error in `HealingNode.svelte:22-25`**
```typescript
const { actorType, permissions } = messageDocument.system as {
    actorType: string;
    permissions: unknown;
};
```
---

### Worth knowing about

- **Bonus dice regex only hits the first dice group**: `#applyHealingBonus` uses a non-global regex, so `2d4+1d6` would become `3d4+1d6`. Probably fine for healing potions (single pool), but a comment explaining the intent would help.
- **Undo is snapshot-based**: if something else modifies HP between apply and undo, the undo will silently overwrite those changes. Inherent to the approach and acceptable, but worth a code comment.
- **No tests for the new rule or bonus logic**: `HealingPotionBonusRule.prePrepareData()` and `#applyHealingBonus()` have no dedicated tests. Edge cases like formulas without dice (`"5"`, `"@level*5"`) aren't covered.
- **The existing Apply Damage button has no permission check**: `RollSummary.svelte:55-65` lets any player click "Apply Damage." The new healing button correctly gates on GM/author. Pre-existing issue.
- **Some CSS in HealingNode is reduntant**: `HealingNode.svelte:253-258`

PR #389, opened 2026-03-03
Spell Compendium Sort fix 

7D7D:
Implemented with a low-complexity, single-file patch.

-   Added mapTierForCompendiumDisplay() to enforce:
    -   utilitySpell -> 0
    -   non-utility spells -> system.tier + 1
-   Updated both normalization branches to use that helper, so sorting headers and tier badges now align with 0=Utility, 1=Cantrip, 2=Tier 1, ....

Screenshot 2026-03-02 230742


trevlar (changes requested):
Can we get unit tests around this that verify it's working and prevent regression.

PR #396, opened 2026-03-05
Enhancements for Drag and Drop on Items in the character sheet.

### LINKED ISSUE #387: Dropping things in character sheet isn't clear
Some feedback is needed when you drop things on a characters sheet. Maybe it can jump to the correct tab after dropping to show that thing actually got added.

Insax:
Add a handling to achieve 3 things in total:

1.) If an Item (of any kind and type) is dropped onto a character sheet, the tab automatically switches to the correct tab.
2.) Every dropped item is highlighted by a short glow to ensure visual feedback.
3.) If the item is not in the viewport, scroll down or up until its visible inside the viewport.

Resolves #387


Fronix (changes requested):
## Things that need fixing before merge

### The flash mechanism is over-engineered, consider using `animationend` instead

The flash is currently applied through three systems at once:

1. **Svelte reactivity**, item IDs flow through `appState.droppedItemFlashIds` → store → context → `class:` directive in each tab component
2. **Direct DOM manipulation**, `_replaceHTML` calls `#flashDroppedItemCards`, which manually queries `[data-item-id]` elements, removes the class, forces a reflow with `void itemCard.offsetWidth`, and re-adds it
3. **A `setTimeout`**, fires after 1200ms to clear the store, manually strip DOM classes via `#clearDroppedItemFlashCards`, and trigger another `render()`

The CSS animation already uses `forwards`, so it holds at the final state (fully transparent), visually the flash is done regardless of whether the class gets removed. The timeout only exists to reset state so a subsequent drop can re-trigger the flash. But this creates several problems: the timeout can fire after the sheet is closed (accessing destroyed references), the duration has to stay in sync between TS and CSS, and there's an extra `render()` call just for cleanup.

A simpler approach would be to use the browser's `animationend` event:

1. On drop → add item IDs to reactive state
2. Svelte applies the class via `class:` directive
3. CSS animation plays
4. `animationend` fires → clear the ID from state → Svelte removes the class
5. Next drop works naturally because the class was removed

This would eliminate `#flashDroppedItemCards`, `#clearDroppedItemFlashCards`, `#dropItemFlashClearTimeout`, the reflow hack, the `DROP_ITEM_FLASH_DURATION_MS` constant, the cleanup timeout, and the leaked-timeout-on-close bug, all at once. The scroll-into-view logic could move into an `$effect` or a one-time callback after the tab switch renders.

---

### `data-item-id` doesn't use the same property everywhere

The `#flashDroppedItemCards` method finds cards by querying `[data-item-id="..."]`, and `shouldFlashDroppedItem` checks whether a card should have the flash class. But across the three tabs, the values fed into these come from different properties:

- **FeaturesTab**, `data-item-id={item.reactive._id}`, flash check uses `item.reactive._id` (consistent)
- **FeaturesTab subclass**, `data-item-id={subclass._id}`, flash check uses `subclass.reactive._id` (mismatch)
- **InventoryTab**, `data-item-id={item._id}`, flash check uses `item.reactive._id` (mismatch)
- **SpellsTab**, `data-item-id={spell._id}`, flash check uses `spell.reactive._id` (mismatch)

Today `._id` and `.reactive._id` probably resolve to the same string, but this is inconsistent and could silently break for synthetic token actors or edge cases. Pick one and use it everywhere, `reactive._id` seems like the right choice given how the rest of the codebase works.

---

### The tab-switching `$effect` writes back to the same state it reads

`PlayerCharacterSheet.svelte:238-248`, This effect reads `appState?.activePrimaryTab`, acts on it, then sets `appState.activePrimaryTab = null`. Because the write changes the same value the effect subscribed to, Svelte re-runs the effect a second time (harmlessly, since the guard catches it). It works, but it's relying on Svelte's loop protection rather than being correct by design.

Wrapping the cleanup in `untrack(() => { appState.activePrimaryTab = null; })` would prevent the unnecessary re-trigger.

---

### The golden flash color is hardcoded

`_document-card.scss:348-369`, The keyframe animation uses `rgba(255, 214, 84, ...)` nine times. The style guide says to never hardcode colors and always use `--nimble-*` custom properties so they work in both light and dark mode. A `--nimble-drop-flash-color` variable with opacity variations would follow convention.

---

### `getContext` calls are untyped

The style guide shows typed context usage: `getContext<NimbleCharacter>('actor')`. The PR adds untyped `getContext('droppedItemFlashIds')` in all three tab components (FeaturesTab:121, InventoryTab:50, SpellsTab:154). Since `lang="ts"` was added to these files, these should be typed:

```ts
const droppedItemFlashIdsStore = getContext<Readable<Set<string>>>('droppedItemFlashIds');
```

---

### Constants and types are inlined in the sheet controller

`PlayerCharacterSheet.svelte.ts` defines several constants and a type at the top of the file that don't belong there:

- **`PrimaryTabName`**, This type lists all seven tab names (`core`, `conditions`, `inventory`, etc.), but the same names already exist in the `navigation` array inside `PlayerCharacterSheet.svelte` (lines 189-232). The tab names are now defined in two places with no connection between them. If a tab is added or renamed, both need updating. This type should be derived from the navigation data, or defined once in a shared location that both the controller and the component import.

- **`ITEM_TYPE_TO_PRIMARY_TAB`**, This maps item types to tab names. That's configuration, which tab an item type belongs to is a game-level concept, not sheet controller logic. The style guide says policy values that affect behavior belong in `config.ts` or a dedicated constants file, not inlined in a class file.

- **`DROP_ITEM_FLASH_CLASS`**, A CSS class name embedded in the TS controller, coupling it to specific CSS implementation. If the `animationend` approach is adopted this goes away entirely, but even as-is it shouldn't live here.

- **`DROP_ITEM_FLASH_DURATION_MS`**, Also goes away with `animationend`, but same issue, a CSS timing value hardcoded in a controller.

The tab type and item-to-tab mapping should move to a shared location (a `types/` file for the type, `config.ts` or a sheet constants file for the mapping). If the flash mechanism is reworked with `animationend`, the two `DROP_ITEM_FLASH_*` constants can simply be removed.

---

## Things worth improving

### The explicit `render()` calls after sort/create may double-render

`PlayerCharacterSheet.svelte.ts:160, 172, 183`, After `_onSortItem` and `createEmbeddedDocuments`, Foundry's document update system already triggers a re-render through the reactivity pipeline. The additional `void this.render()` calls likely cause a redundant second render. Worth verifying whether these are actually needed or if the flash/tab state changes are picked up automatically.

---

### Dropping an external item onto an inventory row might get handled twice

`PlayerCharacterInventoryTab.svelte:92-100`, The new `handleItemDrop` function calls `sheet._onDropItem(event, dropData)` when an external item is dropped onto an existing inventory card. But the sheet itself likely has a drop handler from `ActorSheetV2` that could also fire for the same event. The `stopPropagation()` inside `_onDropItem` helps, but it's worth confirming this doesn't result in duplicate processing.

---

## Minor things

- **Dead override**, `_attachFrameListeners` at `PlayerCharacterSheet.svelte.ts:97-99` only calls `super._attachFrameListeners()` with nothing else. Looks like leftover from a previous approach and can be removed.

- **Redundant `overflow: hidden`**, The `--drop-flash` modifier in `_document-card.scss:24-27` adds `overflow: hidden`, but the base `.nimble-document-card` already declares it at line 20. The `position: relative` is needed, but the overflow is just noise.

- **`$state` on the navigation array**, `PlayerCharacterSheet.svelte:189` wraps the navigation array in `$state()` but it's never mutated. A plain `const` would be clearer and avoids unnecessary reactivity tracking.

src/view/sheets/PlayerCharacterSheet.svelte
```diff
@@ -279,6 +268,51 @@
...
+
+			const containerRect = scrollContainer.getBoundingClientRect();
+			const itemRect = droppedItemCard.getBoundingClientRect();
+			const isOutOfView =
+				itemRect.top < containerRect.top || itemRect.bottom > containerRect.bottom;
+			if (!isOutOfView) return;
+
+			droppedItemCard.scrollIntoView({ block: 'nearest', inline: 'nearest', behavior: 'smooth' });
+		});
+	});
```
Fronix: The race condition you mentioned. I think the flash part is actually fine since it's purely reactive (the CSS class binding applies naturally once the item exists in the list). The scroll is where the timing breaks down.

The `tick().then(querySelector)` approach assumes the DOM element exists after Svelte flushes, but Foundry's `createEmbeddedDocuments` is async and the item may not be in `actor.reactive.items` yet at that point. The `scrollSignature` + `actorItemCount` dependency is what saves it (the effect re-fires when the count changes), but it's a fragile indirect coupling. For example, if two items are dropped and the count only increments once, the signature may not change a second time.

**One approach that might solve this:**
use a `MutationObserver` on the scroll container (`.nimble-sheet__body`) watching for `childList` additions. When flash IDs are set, start observing. When a node with a matching `data-item-id` appears, scroll to it and disconnect. This completely decouples from both Svelte's and Foundry's render timing since you're reacting to the actual DOM insertion. The observer can be set up and torn down inside this `$effect`'s lifecycle.

trevlar (dismissed):
Fix the conflicts in this merge request. Let me know if you need any help with that.

Fronix (changes requested):
There are unrelated changes in the `/docs` folder, can you remove those please or put them up in a different PR.

src/view/sheets/playerCharacterPrimaryTabs.ts
Fronix, line 1: This file is not imported anywhere, if it's not used anymore it should be removed.

src/view/sheets/PlayerCharacterSheet.svelte
```diff
@@ -51,6 +81,80 @@
...
+			scrollDroppedItemCardIntoView(droppedItemCard);
+			return true;
+		};
+
+		if (tryScrollToDroppedItem()) return;
+
+		const observer = new MutationObserver(() => {
+			if (!tryScrollToDroppedItem()) return;
+			observer.disconnect();
+		});
```
Fronix, line 148: `MutationObserver` has no timeout/safety, it watches for the dropped item card to appear in the DOM. If the card never renders (e.g., item creation fails silently, or the item ends up filtered out by a search term), the observer stays active indefinitely until the next reactive cycle. Consider adding a `setTimeout` fallback to disconnect after a reasonable period (e.g., 2-3 seconds).

src/view/sheets/pages/PlayerCharacterInventoryTab.svelte
```diff
@@ -86,6 +96,21 @@
 		}
 	}
 
+	async function handleItemDrop(event, item) {
```
Fronix: This function should be typed

src/view/sheets/PlayerCharacterSheet.svelte
```diff
@@ -51,6 +81,80 @@
...
+						new Set([...currentDroppedItemIds, ...droppedItemIds]),
+					);
+				});
+			}
+		}
+
+		untrack(() => {
+			sheetState.activePrimaryTab = null;
+		});
+	});
```
Fronix, line 121: Complex `$effect` for tab switching could use a comment, the `untrack`/`tick` dance in the tab-switching effect is non-trivial: it clears flash IDs inside `untrack`, waits a tick, then re-merges them. This prevents the flash effect from triggering before the tab switch renders. A brief comment explaining *why* would help future maintainers understand this isn't accidental.

src/view/sheets/pages/PlayerCharacterFeaturesTab.svelte
```diff
@@ -107,6 +116,11 @@
...
 	}
 
+	function handleDropFlashAnimationEnd(event: AnimationEvent, itemId: string) {
+		if (event.animationName !== DROP_ITEM_FLASH_ANIMATION_NAME) return;
+		sheet.clearDroppedItemFlash(itemId);
+	}
+
 	function sortItemCategories([categoryA], [categoryB]) {
 		return validTypes.indexOf(categoryA) - validTypes.indexOf(categoryB);
 	}
```
Fronix: This function needs to be typed

Fronix (changes requested):

Fronix (approved):
I think this looks good for now

PR #412, opened 2026-03-06
Apply Damage button - affecting action count

7D7D:
For this feature update, the relevant change is small:

The Apply Damage flow still updates HP, but the combatant sync logic no longer resets a living player's remaining actions. Current actions are now only changed when a combatant is actually defeated or revived, so clicking Apply Damage will not refill actions to max.

Also added regression tests to confirm:

-   a living character taking damage keeps the same action count
-   a living NPC taking damage keeps the same action count
-   defeat/revive behavior still works as before

### DIFF
--- a/src/documents/chatMessage.test.ts
+++ b/src/documents/chatMessage.test.ts
@@ -3,6 +3,11 @@ import { NimbleChatMessage } from './chatMessage.js';
 
 type TestGlobals = {
 	fromUuidSync: ReturnType<typeof vi.fn>;
+	game: {
+		user: {
+			isGM: boolean;
+		};
+	};
 	ui: {
 		notifications: {
 			info: ReturnType<typeof vi.fn>;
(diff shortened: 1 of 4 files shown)


Fronix (changes requested):
A few things

src/utils/replaceRollFormulaData.ts
```diff
@@ -0,0 +1,25 @@
+type RollConstructorWithReplace = typeof Roll & {
+	replaceFormulaData?: (formula: string, data?: Record<string, unknown>) => string;
+};
+
+export default function replaceRollFormulaData(
```
Fronix: Add jsdoc to the function

7D7D:
> A few things

Which things do I need to look at again? I don't see anything in your comment.

src/view/handlers/updateDocumentImage.ts
trevlar: How is this file related to the action count?

trevlar (changes requested):
I'm not sure how this fixes the issue with applying damage affecting action count.

trevlar (approved):

PR #414, opened 2026-03-06
fix: enable editing saves on NPC monster sheets

### LINKED ISSUE #413: Editing non PC sheets limited
Probably a regression from #373 

Image

Image

trevlar:
## Summary
- Add `editingEnabled` prop to `SavingThrows` component, mirroring the pattern used in the movement speed fix (#397)
- Pass `editingEnabled` to `SavingThrows` in `NPCCoreTab` so the saves config button is visible and functional

## Test plan
- [ ] Open an NPC/monster sheet
- [ ] Verify the saves edit button (pencil icon) is visible next to the saves header
- [ ] Click the edit button and confirm the saving throw config dialog opens

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,17 +9,20 @@
 ### Development Setup
 
 1. **Clone the repository:**
+
    ```bash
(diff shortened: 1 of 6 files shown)


Fronix (approved):
Works for me

PR #424, opened 2026-03-10
feat(sheets): floating combat actions tracker on PC sheet

trevlar:
## Summary

Adds an action tracker bar to the PC sheet that displays during combat, showing action pips and an end turn button.

Screenshot 2026-03-11 at 8 52 00 PM

Screenshot 2026-03-11 at 8 52 13 PM

**Features:**
- Positioned flush with sheet edge, aligned with HP section
- Compact button sizing to match sheet proportions
- Clickable action pips to spend/restore actions
- End turn button with subtle ring animation when ready
- Theme-aware styling for light/dark modes

## Test plan
- [ ] Enter combat with a PC and verify the action tracker appears
- [ ] Verify action pips display correctly and can be clicked to spend/restore
- [ ] Verify end turn button shows green with animation when all actions spent
- [ ] Test in both light and dark modes


Fronix (dismissed):
Functionally it works well.

The width of the floating bar is wider than the action buttons on the sheet, it also in general doesn't look like it fits well. Not sure it's a huge thing but it does kind of look out of place. I'd suggest less padding for the dice boxes and add some clearer borders for lightmode.

image
image

PR #449, opened 2026-03-17
fix: ActionTracker shows for started combat even if not active

trevlar:
## Summary
- Fixed ActionTracker not appearing when combat is started but `combat.active` is false
- Now checks for either `combat.active` or `combat.started` to determine if combat is in progress

## Problem
The ActionTracker component only showed when `combat.active` was true, but in some cases combat can be started (round >= 1) while `active` is false. This caused the ActionTracker to not appear even when the player was in active combat.

## Test plan
- [x] Start combat with a character
- [x] Verify ActionTracker appears on the character sheet
- [x] Verify ActionTracker shows initiative roll button or action pips as appropriate

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -78,3 +78,26 @@ jobs:
       - run: pnpm install --frozen-lockfile
       - name: Check circular dependencies
         run: pnpm circular-deps
+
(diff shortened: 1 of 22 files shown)


Fronix (approved):
LGTM

PR #461, opened 2026-03-19
Enhancement/convert to spaces

### LINKED ISSUE #348: "Convert to Spaces" in Scenes context menu
### The Problem
As of right now most Maps/Map Packs usually designate their Grid Size in feet (ft).
When working with Illumination this can become a hassle sometimes as you have to move through all the tokens present on the scene and alter them in order to alter their vision ranges. then you also go through all the lightning sources on the scene and alter those too. Then you might also have to change elevation changes (from the Levels Module) and alter those too and so on and so on.

As those map packs are quite popular it happens quite frequently i believe

### Possible Solution
Add a new button to a scene context menu that detects if the grid size is [a multiple of 5] ft. If yes the button should go through all objects on the scene (except objects that already use the grid sizes and tokens) and change all values to the appropriate spaces value.

### Benefit
Moving from a "feet based" system to nimble becomes much easier, saves the DM a lot of time and makes the system more approachable when using other systems.

Insax:
Enables an automatic conversion to spaces for the following:
- Grid Size
- Ambient Light (Bright/Dim)
- Token Vision
- Token Light
- Ambient Sound

It only starts intervening if the given scene has [a multiple of 5] feet or ft defined as grid size.

This allows for automatic conversion for premade mappacks that are made especially for 5e to work for nimble without much work for the DM.

Resolves #348


docs/.vitepress/theme/index.ts
Fronix, line 1: Please remove these unrelated changes in `/docs`

Fronix (changes requested):

docs/.vitepress/theme/index.ts
Insax: Done, i might just wanna add that was added by the Pre-Push Hook checks, might need an exception there then.

src/documents/scene/scene.ts
```diff
@@ -0,0 +1,142 @@
...
+	sight?: {
+		range?: number | null;
+	};
+	updateSource: (changes: Record<string, unknown>) => void;
+};
+
+type AmbientSoundLike = {
+	radius?: number;
+	updateSource: (changes: Record<string, unknown>) => void;
+};
```
Fronix: Move this to a `sceneTypes.ts` file or into `types/scene.ts` so we keep things clean.

src/documents/scene/scene.ts
```diff
@@ -0,0 +1,142 @@
...
+			convertAmbientLightRange(light);
+		}
+
+		for (const token of this.tokens.contents as unknown as TokenLike[]) {
+			convertTokenLightRange(token);
+		}
+
+		for (const sound of this.sounds.contents as unknown as AmbientSoundLike[]) {
+			convertAmbientSoundRadius(sound);
+		}
```
Fronix: There are three `as unknown as` casts. These bypass type safety entirely. Consider using a type guard function or at least narrowing the cast with a more specific intermediate type. The `*Like` types were created for this purpose, a runtime check would be safer:

```typescript
function isAmbientLightLike(value: unknown): value is AmbientLightLike {
  return typeof value === 'object' && value !== null && 'updateSource' in value;
}
```

src/documents/scene/scene.test.ts
Fronix, line 1: The tests cover the main conversion and skip-on-ineligible cases, but miss:
- Ambient **light** conversion (lights array is empty in the test fixture)
- Token with **both** light and vision ranges
- Token with light but **no** vision
- Grid units = `"ft"` (alternate spelling)
- Grid distance = 0 (edge case — is 0 a multiple of 5?)
- Scene with no tokens/sounds/lights (empty collections)

Fronix (changes requested):
Looks good just a few small things

Fronix (approved):

PR #463, opened 2026-03-20
fix: Condition and hit point tracking unique per token

trevlar:
## Summary
- Fix recovery from dead state not working when HP goes above 0
- Fix dead state not being reapplied when HP returns to 0 after manual status removal  
- Add fallback for when object identity check fails with synthetic actors (unlinked tokens)
- Sync defeated status on tokens outside of combat (consistent with how bloodied works)

## Root Cause
The code relied on object identity (`combatant.actor === actor`) to match the triggering actor with its combatant. In Foundry VTT with synthetic actors (unlinked tokens), `combatant.actor` can return a different object reference than the actor passed to the `updateActor` hook, causing the status effect toggle to be skipped.

Additionally, defeated status was only synced for actors in combat, while bloodied status synced regardless of combat state - an inconsistency.

## Test plan
- [x] All 585 tests pass (14 new tests added)
- [x] Test NPC token: reduce HP to 0 → should show dead
- [x] Test NPC token: increase HP above 0 → should remove dead
- [x] Test NPC token outside combat: reduce HP to 0 → should show dead
- [x] Test manually removing dead status then setting HP to 0 → should reapply dead

### DIFF
--- a/src/hooks/combatantDefeatSync.ts
+++ b/src/hooks/combatantDefeatSync.ts
@@ -38,24 +38,77 @@ function getShouldBeDefeatedFromActorState(combatant: Combatant.Implementation):
 	return hpValue <= 0;
 }
 
+function getShouldBeDefeatedFromActor(actor: Actor.Implementation): boolean | null {
+	// For characters, use wounds logic
+	if (actor.type === 'character') {
+		const wounds = getActorWoundsValueAndMax(actor);
+		if (!wounds) return null;
+
+		return wounds.value >= wounds.max;
+	}
+
+	// For all other actor types (NPCs, monsters, etc.), use HP logic
+	const hpValue = getActorHpValue(actor);
+	if (hpValue === null) return null;
+
+	return hpValue <= 0;
+}
+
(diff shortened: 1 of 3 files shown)


src/hooks/combatantDefeatSync.ts
```diff
@@ -38,24 +38,77 @@ function getShouldBeDefeatedFromActorState(combatant: Combatant.Implementation):
...
+			// For linked tokens, use || to ensure defeated if ANY combatant should be defeated
+			if (triggeringActorShouldBeDefeated === null) {
+				triggeringActorShouldBeDefeated = shouldBeDefeated;
+			} else {
+				triggeringActorShouldBeDefeated = triggeringActorShouldBeDefeated || shouldBeDefeated;
+			}
+		} else {
+			// Different actor instance (other unlinked tokens of the same base type)
+			const currentState = otherActorDefeatStates.get(combatantActor) ?? false;
+			otherActorDefeatStates.set(combatantActor, currentState || shouldBeDefeated);
```
Fronix, line 110: Good thinking here

Fronix (approved):
LGTM

PR #464, opened 2026-03-20
fix: improve migration reliability

Fronix:
## Summary

After some investigation into why migrations can be flaky, found and fixed several issues in the migration infrastructure:

- **Source ID lookup hardening** — Migrations checked `flags.core.sourceId` to identify compendium-sourced items, but items granted via the GrantItem rule set `_stats.compendiumSource` instead. Added a shared `getSourceId()` helper to `MigrationBase` that checks both fields (matching `NimbleBaseItem.sourceId` logic), and updated Migration006-012 to use it. Current migrations mostly target drag-and-dropped features so this is mainly defensive, but it prevents silent skips if future migrations target granted equipment.

- **GM-only migration guard** — Migrations ran for every user including players, who can't write to world settings or update documents. Now scoped to GM only.

- **Per-phase version tracking** — `worldSchemaVersion` was only updated after *all* migration phases completed. If anything crashed mid-run, the next reload would re-run everything from scratch. Now updates after each phase.

- **Missing `await` on table update** — `table.update()` was fire-and-forget, so errors weren't caught and completion wasn't guaranteed.

- **Error visibility** — Failed documents were silently swallowed. Now tracks errors and shows a `ui.notifications.warn()` with the count, with full details logged to console.

- **Missing localization keys** — Migration notification strings (`NIMBLE.migration.world.starting`, etc.) weren't in `en.json`, so users saw raw key paths.

## Test plan

- [ ] Log in as a non-GM player with pending migrations — confirm no migration runs and no errors
- [ ] Reset `worldSchemaVersion` to 0 on a test world `game.settings.set('nimble', 'worldSchemaVersion', 0)`, reload, confirm all migrations run and complete with readable notifications
- [x] Confirm `pnpm check` passes

### DIFF
--- a/src/hooks/ready.ts
+++ b/src/hooks/ready.ts
@@ -13,24 +13,26 @@ import registerMinionGroupTokenActions from './minionGroupTokenActions.js';
 let canvasConditionsPanelComponent: object | null = null;
 
 export default async function ready() {
-	// Run migrations if needed
-	const worldSchemaVersion = game.settings.get(
-		'nimble' as 'core',
-		'worldSchemaVersion' as 'rollMode',
-	) as unknown as number;
-	const latestSchemaVersion = MigrationRunnerBase.LATEST_SCHEMA_VERSION;
+	// Only the GM should run migrations (requires world-level write permissions)
+	if (game.user?.isGM) {
+		const worldSchemaVersion = game.settings.get(
+			'nimble' as 'core',
+			'worldSchemaVersion' as 'rollMode',
+		) as unknown as number;
+		const latestSchemaVersion = MigrationRunnerBase.LATEST_SCHEMA_VERSION;
 
-	if (worldSchemaVersion < latestSchemaVersion) {
-		console.log(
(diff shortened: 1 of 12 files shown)


src/hooks/ready.ts
```diff
@@ -13,24 +13,26 @@ import registerMinionGroupTokenActions from './minionGroupTokenActions.js';
 let canvasConditionsPanelComponent: object | null = null;
 
 export default async function ready() {
-	// Run migrations if needed
-	const worldSchemaVersion = game.settings.get(
-		'nimble' as 'core',
-		'worldSchemaVersion' as 'rollMode',
-	) as unknown as number;
-	const latestSchemaVersion = MigrationRunnerBase.LATEST_SCHEMA_VERSION;
+	// Only the GM should run migrations (requires world-level write permissions)
```
trevlar, line 16: Good catch!

trevlar (approved):
Thank you for fixing this

PR #494, opened 2026-03-22
fix: make action dice discrete and show tooltips to the right

### LINKED ISSUE #482: Revise action dice to remove only one at a time and not require linear removal.

trevlar:
## Summary
- Each action pip click now spends/restores exactly **one** action instead of setting actions to the clicked pip's index
- Fixes unintuitive behavior where clicking pip 1 would consume all actions at once
- Tooltips now appear to the right of the action tracker buttons instead of below

Fixes #482

## Test plan
- [x] Enter combat with a character and roll initiative
- [x] Verify clicking any available (green) pip decrements actions by 1
- [x] Verify clicking any spent (gray) pip increments actions by 1
- [x] Verify tooltips appear to the right of the action tracker buttons
- [ ] Test with boons/items that grant more than 3 actions (system already supports this)

Closes #482

### DIFF
--- a/src/view/sheets/components/ActionTracker.svelte
+++ b/src/view/sheets/components/ActionTracker.svelte
@@ -27,6 +27,7 @@
 					type="button"
 					aria-label={localize('NIMBLE.ui.heroicActions.rollInitiative')}
 					data-tooltip={localize('NIMBLE.ui.heroicActions.rollInitiative')}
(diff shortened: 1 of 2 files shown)


Fronix (approved):
LGTM

**Note:**
- Agressive boon doesn't work yet
- Epic Agility boon doesn't work yet

PR #495, opened 2026-03-22
fix(inventory): make equipped the source of truth for object rule state

### LINKED ISSUE #462: Items get auto-equipped and are not equippable if editmode is off
## Two issues

1. When you drop an item in the inventory, the item gets auto-equipped. It should not be.
2. If the edit-mode is toggled off, you can't equip/unequip items. This should be possible without having to have editmode enabled.

https://github.com/user-attachments/assets/2a3f004b-df66-4dbf-aaec-328486255939

Insax:
- add `system.equipped` to object item data model and ambient item typings
- add `RulesManager.enableAllRules()` / `disableAllRules()` helpers
- rework inventory equip toggle to:
  - read UI state from `system.equipped`
  - toggle all rules via RulesManager
  - persist `system.equipped`
  - rollback rule state if item update fails
- sync embedded object rule `disabled` flags to `!system.equipped` on item preCreate (drop-time sync)
- rename object helper to `toggleEquipment()`
- localize equip toggle prompt with new `NIMBLE.prompts.toggleEquipment`
  - add key to `en`, `es`, and `fr` using English base text
- add focused tests for RulesManager all-rules toggles
- keep inventory toggle available regardless of edit mode

This fix adresses #462


src/view/sheets/pages/PlayerCharacterInventoryTab.svelte
```diff
@@ -255,33 +210,41 @@
...
+										});
+
+										if (updatedItem) return;
+
+										if (newEquippedState) {
+											await rules.disableAllRules();
+										} else {
+											await rules.enableAllRules();
+										}
+									}}
```
trevlar: Will you extract this onclick handler into a named function and passed into onclick handler?

trevlar (approved):

PR #496, opened 2026-03-22
feat: drag heroic actions and reactions to macro bar

### LINKED ISSUE #465: Ability to drag actions and reactions to macro bar.
The ability to drag the heroic actions and heroic reactions to the macro bar. This allows easy use of the actions. They would bring up the appropriate dialog and roll. 

This would include Attack like Unarmed Strike, Cast Spell, Move, Assess, Interpose, Defend, Interpose & Defend, Opportunity Attack, and Help.

trevlar:
Screenshot 2026-03-21 at 8 53 52 PM
Screenshot 2026-03-21 at 8 53 44 PM

## Summary

- Enable dragging heroic actions and reactions from the character sheet's Actions tab to the macro bar
- Draggable action/reaction buttons:
  - **Action tab buttons**: Attack, Spell, Move, Assess
  - **Reaction tab buttons**: Defend, Interpose, Opportunity Attack, Help, Interpose & Defend
  - **Individual buttons**: Move (Confirm Move), Unarmed Strike, Help, Defend, Interpose
- Actions that execute directly from macro without opening a panel:
  - **Move**: deducts action pip and posts chat message
  - **Unarmed Strike**: opens roll dialog, executes attack, deducts action pip
  - **Defend**: uses defend reaction and posts chat message
  - **Interpose**: uses interpose reaction and posts chat message
  - **Interpose & Defend**: uses both reactions and posts both messages
  - **Help**: uses help reaction and posts chat message
- Confirmation dialogs when using reactions without available actions or when reaction is already spent
- Other actions open the character sheet to the appropriate panel for further user interaction

Fixes #465

## Test plan

- [x] Open a character sheet and go to the Actions tab
- [x] Drag an action tab (Attack, Spell, Move, Assess) to the macro bar - verify icon appears
- [x] Select a character token and click the macro - verify behavior:
  - Attack/Spell/Assess: opens sheet to correct panel
  - Move: posts move message and deducts action (in combat)
- [x] Drag the "Confirm Move" button to macro bar and test it works
- [x] Drag the "Unarmed Strike" card to macro bar and verify it rolls attack
- [x] Drag reaction tabs (Defend, Interpose, Opportunity Attack, Help) to macro bar
- [x] Drag individual reaction buttons (Defend, Interpose, Help) to macro bar
- [ ] Test reactions execute directly when clicked (in combat)
- [x] Test confirmation dialog appears when no actions remaining or reaction already spent
- [x] Test the Interpose & Defend combo button in the Defend/Interpose panels
- [ ] Run `pnpm check` to verify no lint/type/test errors

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -804,6 +804,26 @@
 				"enterCombat": "Enter combat to use actions",
 				"outsideCombat": "You are outside of combat",
 				"noActions": "You don't have any actions",
+				"confirmReaction": {
+					"title": "Use Reaction?",
+					"noActionsMessage": "You don't have any actions remaining.",
+					"spentMessage": "You've already used {reaction} this round.",
+					"bothMessage": "You don't have any actions remaining and you've already used {reaction} this round.",
+					"confirmQuestion": "Do you still want to use the {reaction} reaction?",
+					"confirm": "Use Reaction",
+					"cancel": "Cancel"
+				},
+				"confirmMove": {
+					"title": "Move?",
+					"noActionsMessage": "You've used all your actions this round.",
+					"confirmQuestion": "Do you still want to move?",
+					"confirm": "Move",
+					"cancel": "Cancel"
+				},
+				"confirmAttack": {
+					"confirmQuestion": "Do you still want to attack?",
+					"confirm": "Attack"
+				},
 				"noSpells": "You don't have any spells",
 				"endTurn": "End Turn",
 				"rollInitiative": "Roll Initiative",
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -824,6 +824,10 @@
 					"confirmQuestion": "Do you still want to attack?",
 					"confirm": "Attack"
 				},
+				"macroWarnings": {
+					"selectCharacterToken": "Select a character token to use this action",
+					"mustBeInCombat": "You must be in combat to use {action}"
+				},
 				"noSpells": "You don't have any spells",
 				"endTurn": "End Turn",
 				"rollInitiative": "Roll Initiative",
--- a/src/documents/combat/combat.svelte.test.ts
+++ b/src/documents/combat/combat.svelte.test.ts
@@ -2031,6 +2031,257 @@ describe('NimbleCombat', () => {
 		expect(combat.updateEmbeddedDocuments).not.toHaveBeenCalled();
 	});
 
+	it('allows force to override the no-actions heroic reaction blocker', async () => {
+		globals().game.user.isGM = false;
+		globals().game.user.role = 1;
+		const combatId = 'combat-sheet-force-no-actions';
+		const activeCharacter = createMockCombatant({
+			id: 'active-character',
+			type: 'character',
+			sort: 1,
+			isOwner: false,
+			initiative: 15,
+			actionsCurrent: 3,
+			actionsMax: 3,
+			actor: createCombatActorFixture({ hp: 8, woundsValue: 0, woundsMax: 6 }),
+			combatId,
+		});
+		const reactingCharacter = createMockCombatant({
+			id: 'reacting-character',
+			type: 'character',
+			sort: 2,
+			isOwner: true,
+			initiative: 12,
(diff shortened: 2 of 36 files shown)


Fronix:
@trevlar overall, i think this looks good, there is one thing I feel is missing, we shouldn't block the clicking of heroic reactions, instead if no ally is targeted we should trigger a "Select target"-mode where we tell the user to select a token. That way you don't have to go back and forth to do what you initially wanted to do.

image
image

Fronix (dismissed):
## General
Tested all the different ways to use the reactions and they work correctly, it's a bit confusing that when some are on the bar and you click, the character sheet pops up but I think we can live with that for now.

## Blocking

- `src/macros/activateHeroicActionMacro.ts:1022`, `inCombat` check is inverted when there is no combat; `undefined !== null` evaluates to `true`, so the function skips the warning and later crashes on `combatant!.update(...)`.
  ```ts
  const combatant = combat?.combatants.find((c) => c.actorId === actor.id);
  const inCombat = combatant?.initiative !== null;
  if (!inCombat) { ui.notifications.warn(...); return; }
  ```
  Fix: `const inCombat = !!combatant && combatant.initiative !== null;` (and guard `combat` too).

- `src/macros/createHeroicActionMacro.ts:1832`, magic number `3` for ownership.
  ```ts
  const perms = m.ownership?.default === 3 || m.ownership?.[game.user!.id] === 3;
  ```
  Fix: use `CONST.DOCUMENT_OWNERSHIP_LEVELS.OWNER`.

## Suggestions

- `src/macros/activateHeroicActionMacro.ts:1072-1258`, 5 `execute*Reaction` functions are near-identical boilerplate (resolve combat/combatant → confirm → useHeroicReactions → ChatMessage.create). Extract a single helper that takes `{ reactionKeys, reactionType, buildSystem }`; would eliminate ~150 lines of repetition.
- `src/macros/activateHeroicActionMacro.ts:879`, macro imports from `../view/sheets/components/attackUtils.js`; crosses the view boundary from a non-view layer. Move `getUnarmedDamageFormula`/`hasUnarmedProficiency` under `utils/` or `documents/`.
- `src/macros/activateHeroicActionMacro.ts:1052`, move path decrements to `actionsRemaining - 1` (can go negative), whereas `activateItemMacro.ts:1673` clamps at 0. Decide on one policy; negative pips are likely unintended even when forced.
- `src/view/sheets/components/DefendReactionPanel.svelte.ts:14-18` & `InterposeReactionPanel.svelte.ts:7-12`, factory signatures are now 8 positional getter args; switch to a single options object to avoid arg-order bugs.
- `src/view/sheets/components/DefendReactionPanel.svelte.ts` / `InterposeReactionPanel.svelte.ts`, `handleInterposeAndDefend` bodies are duplicated verbatim in both files; lift into a shared helper.
- `src/macros/activateHeroicActionMacro.ts` + all three reaction panel `.svelte.ts` files, same `showReactionConfirmation` wrapping + `force ? { force: true } : undefined` pattern repeats ~6 times; consolidate into one helper.
- `src/hooks/hotBarDrop.ts:3`, untyped params (`_, data, slot`); while preexisting, this file is now touched and could gain a proper signature.
- `src/view/sheets/pages/PlayerCharacterHeroicActionsTab.svelte:3019`, `$effect` mutates `sheetState.heroicActionTarget = null` inside `untrack`; acceptable but consider a dedicated action/ack pattern instead of shared-state null reset.
- `src/macros/activateHeroicActionMacro.ts:1408`, `Array.from(game.user?.targets?.map(...))`, `Set#map` doesn't exist; this relies on Foundry's extended Set. Use `getTargetedTokens(actor.id)` like the other branches for consistency.
- Test `src/macros/activateHeroicActionMacro.test.ts:835` asserts `useHeroicReactions` called with `undefined` third arg for the `interposeAndDefend` combo. Code passes `force ? { force: true } : undefined`; fine today but brittle if the helper evolves, prefer `expect.objectContaining` or omit the third arg assertion.

Fronix (approved):
LGTM

PR #503, opened 2026-03-23
Fix: Remove combatant when token deleted, fix death state

### LINKED ISSUE #470: Scene/combat cleanup is not staying in sync.
Deleted tokens are not removed from the combat tracker.

Deleting a token from the scene should remove it from combat.

trevlar:
## Summary
- Adds automatic removal of combatants from the combat tracker when their corresponding token is deleted from the scene
- Fixes death state sync to properly handle unlinked tokens that share the same base actor ID but have independent HP values
- Refactors combatant creation hooks to use targeted single-combatant sync functions to avoid cross-contamination between unlinked tokens

## Changes
- **New file `tokenCombatantSync.ts`**: Registers a `deleteToken` hook that removes orphaned combatants when tokens are deleted
- **`combatantDefeatSync.ts`**: Added `syncSingleCombatantDeathState()` for targeted sync on combatant creation; refactored actor defeat state tracking to use per-instance maps
- **`combatantHealthStateSync.ts`**: Added `syncCombatantHealthState()` wrapper for targeted sync on combatant creation
- **`nimble.ts`**: Registers the new token-combatant sync hook

## Test coverage
- **New `tokenCombatantSync.test.ts`** (8 tests): Token deletion removes combatants, scene/token ID validation, multi-combat handling, GM permission checks
- **`combatantDefeatSync.test.ts`** (+7 tests): createCombatant hook tests for death state sync, unlinked token isolation, turn advancement
- **`combatantHealthStateSync.test.ts`** (+5 tests): createCombatant hook tests for bloodied/lastStand sync, unlinked token isolation

## Test plan
- [x] Create multiple unlinked tokens of the same actor on a scene
- [x] Add them to combat and verify each maintains independent HP/death state
- [x] Delete a token from the scene and verify its combatant is removed from the tracker
- [x] Verify death state correctly applies to only the affected token when HP reaches 0
- [x] Verify health state (bloodied/lastStand) applies correctly per-token


src/hooks/combatantHooks/tokenCombatantSync.ts
Fronix, line 1: We have quite a lot of `combatant...` hooks in the hooks folder, i think we need a folder for `combatantHooks` just to make it easier to see all related hooks.

src/hooks/tokenCombatantSync.test.ts
```diff
@@ -0,0 +1,309 @@
...
+	return getTestGlobals<CombatDefeatSyncTestGlobals>();
+}
+
+function getTestGlobals<T>() {
+	return globalThis as unknown as T;
+}
+
+async function flushAsync() {
+	await new Promise((resolve) => setTimeout(resolve, 0));
+}
```
Fronix: Very picky but better to tackle it now than later.

`globals()` and `flushAsync()` are copy-pasted across test files. `globals()` now appears in 6 test files and `flushAsync()` in 3. `getTestGlobals()` is unique to this new file, the other 5 files define `globals()` directly without the extra indirection, so that's also inconsistent.

| File | `globals()` | `getTestGlobals()` | `flushAsync()` |
|------|:-----------:|:------------------:|:--------------:|
| `src/hooks/tokenCombatantSync.test.ts` | x | x | x |
| `src/hooks/combatantDefeatSync.test.ts` | x | | x |
| `src/hooks/combatantHealthStateSync.test.ts` | x | | x |
| `src/documents/chatMessage.test.ts` | x | | |
| `src/utils/getHeroicReactionUsageState.test.ts` | x | | |
| `src/documents/combat/combat.svelte.test.ts` | x | | |

Could be cleaned up

`tests/helpers.ts`
```ts
  export function getTestGlobals<T>(): T {                                                                                                                                                                                                  
      return globalThis as unknown as T;
  }

  export async function flushAsync(): Promise<void> {
      await new Promise((resolve) => setTimeout(resolve, 0));
  }
```

Then each test file becomes:

```ts
  import { getTestGlobals, flushAsync } from '../../tests/helpers.js';

  function globals() {
      return getTestGlobals<CombatDefeatSyncTestGlobals>();
  }
```

src/hooks/combatantHooks/combatantDefeatSync.ts
Fronix, line 1: There are two functions here with very similar names but different parameter types:
- `getShouldBeDefeatedFromActorState(combatant)` > takes a `Combatant.Implementation`
- `getShouldBeDefeatedFromActor(actor)` > takes an `Actor.Implementation`

The "FromActorState" vs "FromActor" distinction is easy to miss. Renaming the first to something like `getShouldBeDefeatedFromCombatant` would make the difference obvious at a glance.

Fronix (approved):
Small picky stuff but overall LGTM

PR #506, opened 2026-03-23
fix(ui): replace `__img` border with `::after` overlay divider

underscoreRobin:
### **Description**
The right-side divider on document card images was implemented as `border-right` directly on the `` element. This caused the image to render at 31×32px instead of 32×32px, which made `object-fit: cover` resample the image onto an uneven dimensions, resulting in a visibly pixelated image.

This PR wraps the `` in a `__img-wrapper` div and moves the divider to a `::after` pseudo-element positioned outside the image box, so the image always renders at its intended square dimensions.

### **Changes:**
- Added `__img-wrapper` with `::after` overlay divider
- Removed `border-right` from `__img`
- Updated all card usages across inventory, features, spells, NPC, and conditions list tabs

### DIFF
--- a/src/scss/components/_document-card.scss
+++ b/src/scss/components/_document-card.scss
@@ -195,13 +195,28 @@
     display: contents;
   }
 
-  &__img {
-    grid-area: img;
-    flex-shrink: 0;
-    height: var(--nimble-card-image-height, 2rem);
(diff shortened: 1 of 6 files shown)


trevlar (approved):
LGTM

PR #507, opened 2026-03-24
Fix: Apply NPC Armor rules in combat damage application

### LINKED ISSUE #467: Apply Damage Button Doesn't take Monster Armor into Account.
I've noticed that the Apply Damage button only ever applies the total damage of the attack. If a monster has Medium or Heavy armor it is completely ignored.

Insax:
- pass serialized damage roll data and critical flag into apply-damage
- apply armor logic per target in chatMessage damage handling
- medium armor: use dice-only damage unless the hit is a crit
- heavy armor: halve damage (round up) and ignore flat modifiers unless the hit is a crit
- respect global half-damage outcome in the armor-adjusted calculation
- add/adjust tests for armor + crit + half-damage scenarios
- add missing localization for 0 damage.

### DIFF
--- a/src/documents/chatMessage.test.ts
+++ b/src/documents/chatMessage.test.ts
@@ -338,6 +338,31 @@ describe('NimbleChatMessage.applyDamage', () => {
 		expect(globals().ui.notifications.info).toHaveBeenCalledWith('No damage to apply.');
 	});
 
+	it('shows no-damage feedback when total damage is 0 or negative', async () => {
+		const actor = {
+			system: {
+				attributes: {
+					hp: {
+						value: 10,
+						temp: 2,
+						max: 10,
+					},
+				},
+			},
+			update: vi.fn().mockResolvedValue(undefined),
+		};
+
+		globals().fromUuidSync.mockReturnValue({ actor });
+
+		const message = createActivationMessage();
+		await message.applyDamage(0, { outcome: 'fullDamage' });
+		await message.applyDamage(-4, { outcome: 'fullDamage' });
+
+		expect(globals().fromUuidSync).not.toHaveBeenCalled();
+		expect(actor.update).not.toHaveBeenCalled();
+		expect(globals().ui.notifications.info).toHaveBeenCalledWith('No damage to apply.');
+	});
+
 	it('warns when there are no targets', async () => {
 		const message = createActivationMessage([]);
 		await message.applyDamage(4, { outcome: 'fullDamage' });
--- a/src/documents/chatMessage.test.ts
+++ b/src/documents/chatMessage.test.ts
@@ -34,6 +34,43 @@ function createActivationMessage(targets: string[] = ['Scene.scene.Token.token']
 	} as unknown as ChatMessage.CreateData);
 }
 
+function createSerializedDamageRoll(params: {
+	diceResults: number[];
+	flatBonus?: number;
+	isCritical?: boolean;
+	excludedPrimaryDieValue?: number;
+}) {
+	const flatBonus = params.flatBonus ?? 0;
+	const diceTotal = params.diceResults.reduce((sum, result) => sum + result, 0);
+	const total = diceTotal + flatBonus;
+
+	const terms: Record<string, unknown>[] = [
+		{
+			number: params.diceResults.length || 1,
+			faces: 6,
+			results: params.diceResults.map((result) => ({
+				result,
+				active: true,
+				discarded: false,
+			})),
+		},
+	];
+
+	if (flatBonus !== 0) {
+		terms.push({ operator: '+' });
+		terms.push({ number: flatBonus });
+	}
+
+	return {
+		class: 'DamageRoll',
+		formula: `${params.diceResults.length || 1}d6${flatBonus ? ` + ${flatBonus}` : ''}`,
+		total,
+		isCritical: params.isCritical ?? false,
+		excludedPrimaryDieValue: params.excludedPrimaryDieValue ?? 0,
+		terms,
+	};
+}
+
 describe('NimbleChatMessage.applyHealing', () => {
 	beforeEach(() => {
 		globals().fromUuidSync = vi.fn();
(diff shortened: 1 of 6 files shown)


Insax:
This change will adress #467

src/documents/chatMessage.ts
```diff
@@ -36,6 +37,131 @@ interface ActivationCardSystemData {
...
+
+		const diceDamage = getDiceDamageTotal(serializedRoll) ?? rollTotal;
+		let adjustedDamage = diceDamage;
+		if (applyOutcomeHalfDamage) adjustedDamage = Math.ceil(adjustedDamage * 0.5);
+		if (applyHeavyArmor) adjustedDamage = Math.ceil(adjustedDamage * 0.5);
+		totalAdjustedDamage += adjustedDamage;
+	}
+
+	return Math.max(0, Math.floor(totalAdjustedDamage));
+}
```
Fronix: `resolveArmorAdjustedDamage` accepts `damage` but recalculates from `serializedRoll.total` / dice terms when rolls exist (line 146). The `damage` param only matters in the no-rolls fallback (lines 132–142). This is correct behavior (you need the dice/flat breakdown), but the function signature implies it always adjusts the passed-in value.

Consider renaming to `calculateArmorAdjustedDamage` or adding a short comment at the top of the function explaining the two modes: roll-based recalculation vs. fallback adjustment.

src/documents/chatMessage.ts
```diff
@@ -36,6 +37,131 @@ interface ActivationCardSystemData {
...
+				active?: unknown;
+				discarded?: unknown;
+			};
+			if (serializedResult.active === false || serializedResult.discarded === true) continue;
+
+			const resultValue = Number(serializedResult.result);
+			if (!Number.isFinite(resultValue)) continue;
+			diceDamage += resultValue;
+		}
+	}
```
Fronix, line 111: Lines 78-85 check `term instanceof foundry.dice.terms.Die`, then lines 87-105 handle plain serialized objects. After checking Foundry's serialization flow and `DamageRoll.toJSON()` (`src/dice/DamageRoll.ts:506`), the `instanceof` branch is confirmed dead code:

- `DamageRoll.toJSON()` calls `super.toJSON()`, which serializes terms as **plain objects** (not live `Die` instances).
- The `roll` passed through `options` into `applyDamage` is always the raw serialized data from `toJSON()`, never a reconstructed roll. Reconstruction via `DamageRoll.fromData()` only happens for tooltip rendering in `DamageRoll.svelte:52`, and that result is not passed into `options`.

**Recommendation:** Remove the `instanceof` branch (lines 78-85) entirely. The serialized-object path (lines 87-105) handles all real inputs. Keeping the dead branch adds maintenance cost and creates a false impression that live `Die` instances are a supported input.

src/documents/chatMessage.test.ts
Fronix, line 1: There's a test for heavy armor + halfDamage (line 427), but no equivalent for medium armor + halfDamage. The expected behavior (dice-only, then halved for outcome) should be verified:

```ts
// Medium armor + halfDamage: diceResults [6], flatBonus 2 → dice = 6, half -> 3
it('st

File: src/documents/chatMessage.ts [hunk 3]
```diff
@@ -36,6 +37,131 @@ interface ActivationCardSystemData {
...
+		if (damageOptions?.isCritical === true) {
+			if (applyOutcomeHalfDamage) return Math.ceil(params.damage * 0.5);
+			return params.damage;
+		}
+
+		// Without roll metadata, the incoming value already includes outcome scaling.
+		// In this fallback, only apply armor reduction.
+		if (applyHeavyArmor) return Math.ceil(params.damage * 0.5);
+		return params.damage;
+	}
```
Fronix, line 182: The fallback path where `serializedRolls.length < 1` has distinct armor behavior (heavy halves the incoming value at line 140). This path has no test coverage in `src/documents/chatMessage.test.ts`.

src/documents/chatMessage.ts
```diff
@@ -258,10 +384,13 @@ class NimbleChatMessage extends ChatMessage {
 			const currentHp = hpData?.value;
 			if (typeof currentHp !== 'number' || Number.isNaN(currentHp)) continue;
 
+			const adjustedDamage = resolveArmorAdjustedDamage({ actor, damage, options });
+			if (!Number.isFinite(adjustedDamage) || adjustedDamage <= 0) continue;
```
Fronix: `if (!Number.isFinite(adjustedDamage) || adjustedDamage <= 0) continue;`. When armor reduces damage to 0, the target is silently skipped with no feedback. If there is no damage to apply the button should be disabled with a tooltip "No damage to apply".

Fronix (changes requested):
Solid stuff, there are just some small things that could become issues later.

---

I would however want to make a change in the chat cards after this gets merged, it came up now once we actually apply damage based on the armour.

Take this example:
Image

It is very unclear how much damage will be applied to each of these targets, first target will get -9 the other -5 but the damage says 11. Even thought the code will do the right thing, we don't communicate that a different value will be taken.

I've added a new issue for this [510](https://github.com/Nimble-Co/FoundryVTT-Nimble/issues/510), i could add that after this PR gets merged.

src/documents/chatMessage.ts
Insax: I checked some permutations and if all roll results are 0 we report back with "No Damage to apply"

if only partial damage values are 0 we skip over the 0 values and send a notification to inform the user that Token XXX has 0 damag to apply. This is how we can support multiple targets and user information at the same time.

src/documents/chatMessage.ts
```diff
@@ -224,12 +443,12 @@ class NimbleChatMessage extends ChatMessage {
...
 	}
 
-	async applyDamage(value: number, options?: Record<string, unknown>): Promise<void> {
+	async applyDamage(value: number, options?: DamageApplyOptions): Promise<void> {
 		if (!this.isActivationCard()) return;
 		if (!game.user?.isGM) return;
 
 		if (options?.outcome === 'noDamage') {
-			ui.notifications?.info('No damage to apply.');
+			ui.notifications?.info(localize('NIMBLE.chat.noDamageToApply'));
```
Fronix, line 451: Not sure it's an issue but when i tested to apply damage with a card that would have applied 0 damage, i got no warning.

src/documents/chatMessage.ts
```diff
@@ -238,40 +457,52 @@ class NimbleChatMessage extends ChatMessage {
...
 			}
 		}
+
+		for (const tokenName of damageApplicationPlan.zeroDamageTargetNames) {
+			ui.notifications?.info(
+				localize('NIMBLE.chat.targetSkippedZeroDamage', {
+					tokenName,
+				}),
+			);
+		}
```
Fronix, line 495: Same here, not sure it's an issue but might be a problem if there is code that's not calculating targets correctly

Fronix (approved):
I'm willing to approve this but might be worth taking one last look at my comments

PR #513, opened 2026-03-25
Monsters marked as finished their turn on their turn

### LINKED ISSUE #468: Monsters marked as finished their turn on their turn
When a monster's turn comes up they are instantly marked as if they have taken their turn. This prevents the usage of the Combat System popup.

**Expected**
Monsters should only be marked as taken their turn when the GM goes to the next in turn order. If the GM goes back to the previous monster the turn marker should reset.

Image

7D7D:
### PR Summary

Issue
-   Non-player combatants were being treated as if they had already finished their turn when their `current actions` hit `0`.
-   That caused two user-facing problems:
    -   the Combat Tracker could mark monsters as done too early
    -   the NCSW popup could become unusable because the monster looked like it had no valid turn left
-   This broke the intended behavior:
    -   a non-player should only be marked as having taken its turn after the GM advances past it
    -   if the GM goes back to that monster, its turn state should reset and its actions should be restored

Root Cause
-   The system was using the same value for two different meanings:
    -   remaining actions
    -   whether a non-player had already taken its turn
-   In several places, `current actions == 0` was being treated like "turn complete," which is not always true

What This PR Changes
-   Non-players are no longer auto-skipped just because they have `0` actions when their turn comes up
-   Combat Tracker turn-complete logic for non-players now follows initiative progression, not action count
-   Rewinding to a monster turn restores that monster's actions to max
-   Rewinding to a minion-group turn restores all alive members of that group to max actions
-   Newly created non-player combatants are initialized with usable current actions instead of falling into a zero-action state

Behavior After This Fix
-   A monster can become active even if its stored actions are `0`
-   The GM can still use the monster's turn/NCSW flow instead of the system skipping past it
-   A non-player is only marked "done" after initiative actually moves past it
-   If the GM rewinds to that non-player, the completed state clears and its actions are restored

Tests Added/Updated
-   Updated the combat turn test so exhausted non-players are not skipped on `nextTurn()`
-   Added tracker regression coverage to verify:
    -   an active zero-action monster is still considered current
    -   advancing past it marks it as done
    -   rewinding makes it active/current again

### DIFF
--- a/src/documents/combat/combat.svelte.test.ts
+++ b/src/documents/combat/combat.svelte.test.ts
@@ -68,6 +68,13 @@ describe('NimbleCombat', () => {
 		combatPrototype.previousRound = vi.fn(async function (this: Combat) {
 			return this;
 		});
+		combatPrototype.createEmbeddedDocuments = vi.fn(async function (
+			this: Combat,
+			_embeddedName: string,
+			data: unknown,
+		) {
+			return data as foundry.abstract.Document.StoredForName<Combat.Embedded.Name>[] | undefined;
+		});
 
 		const textEditorImpl =
 			globals().foundry.applications.ux.TextEditor.implementation.getDragEventData;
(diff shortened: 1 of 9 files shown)


Fronix (approved):
LGTM i did a small structure change just to fix some minor de duplication

PR #514, opened 2026-03-25
Combat Tracker window width

### LINKED ISSUE #479: Combat tracker does not resize correctly when the window width changes
Responsive layout issue.

7D7D:
### PR Summary

-   Combat tracker width now scales from the actual visible browser viewport instead of the gap between Foundry sidebars.
-   Max width no longer leaves large unused space on browser-based play setups with different resolutions or zoom levels.
-   The width calculation now reserves space for the tracker shell controls, so the right-side action buttons stay visible instead of clipping off-screen.
-   The tracker now responds to both normal window resize events and visual viewport resize events, which improves behavior when browser zoom changes the visible area.
-   The shell width reserve is now defined in one shared constant instead of being duplicated between layout math and CSS.
-   The default Combat Tracker width setting was reduced from `10` to `5`, so new/default setups start at a more moderate width.
-   Added focused regression tests for the width calculation and updated settings tests for the new default.
-   Validation passed with the full project check suite.


Fronix (approved):
LGTM! Just a cleanup in the tests

PR #515, opened 2026-03-26
Initiative Duplicate fix

### LINKED ISSUE #474: Initiative can be rolled multiple times simultaneously
Duplicate initiative rolling is allowed when it should probably be prevented or throttled/debounced.

7D7D:
Checked using AGENTS.md and STYLE_GUIDE.md

Initiative Fix
-   Prevented duplicate initiative rolls by adding a per-combatant initiative lock at the combat document layer.
-   `rollInitiative `now dedupes repeated ids, serializes same-client in-flight requests, and skips rolling if a fresh lock already exists.
-   The lock is acquired before dice are rolled and cleared when the final initiative update is written.
-   Added a stale-lock timeout so abandoned locks do not block initiative forever.

Safety Guard
-   Added a combatant update guard so one request cannot overwrite another active initiative lock.
-   The guard still allows clearing an active lock and replacing a stale lock.

UI Updates
-   Updated the combat tracker to hide or block initiative prompts while a roll is already pending.
-   Updated the action tracker to disable the initiative button and show a waiting state during a pending roll.
-   Updated the core character sheet initiative button to subscribe to combat state and disable itself while the lock is active.

Test Coverage
-   Added a regression test proving two concurrent initiative requests against the same combatant only produce one roll.
-   Added a regression test proving a foreign active lock prevents another client from rolling.
-   Added combatant-level tests for lock overwrite behavior, stale-lock replacement, and lock clearing.

Docs
-   Added the new shared `initiativeRollLock `utility to the style guide inventory.

### DIFF
--- a/src/documents/combat/combat.svelte.test.ts
+++ b/src/documents/combat/combat.svelte.test.ts
@@ -9,6 +9,7 @@ import {
 	getTestGlobals,
 	type NimbleCombatDocumentTestGlobals,
 } from '../../../tests/mocks/combat.js';
+import { initiativeRollLock } from '../../utils/initiativeRollLock.js';
 import { NimbleCombat } from './combat.svelte.js';
 import { clearExpandedTurnIdentityHint } from './expandedTurnIdentityStore.js';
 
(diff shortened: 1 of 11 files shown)


Fronix (approved):
LGTM, small things adjusted and merge conflict.

PR #516, opened 2026-03-26
feature: add HP scrolling text and centralize HP manipulation flow

### LINKED ISSUE #444: Reactivate foundry's 'Scrolling Status Text' functionality
Original discord request: https://discord.com/channels/1389098475374903437/1482664074671030493

> A few versions ago, FoundryVTT included support for "Scrolling Status Text", those are the little numbers that pop up when token values change showing things like "-5" or "+9" 
I'd like to request the inclusion of this feature into the system. 

> You can accomplish this by making use of the following Foundry method:
https://foundryvtt.com/api/classes/foundry.canvas.groups.InterfaceCanvasGroup.html#createscrollingtext

> Here's an example of how it is included in 5e: 
https://github.com/foundryvtt/dnd5e/blob/65ffaa948c3e338d2cab7d8e73a0f17decff6f6c/module/documents/actor/actor.mjs#L3470

> I managed get it working on my end by including it in chatMessage.ts line 273. 
But my implementation was really hacky because I'm not familiar with the project. 
Also this only shows the text when people press the "Apply Damage" button from chat, but you'd probably want some kind of event handler on the health value or include it in same base actor function like they did with 5e.

> Note: There's a toggle to enable/disable these in the base settings under:
Settings > Core > Scrolling Status Text (Checkbox) 

Image

Insax:
- add actor-level HP APIs (`applyDamage`, `applyHealing`, `setCurrentHP`, `setMaxHP`, `setTempHP`)
- emit Foundry scrolling status text from actor HP/temp deltas in update lifecycle
- route chat damage/healing + sheet HP edits through actor methods
- add chat message tests for actor-method delegation and healing undo behavior
- add HP scrolling text colors into a dedicated constants module

This intends to fix #444


Insax:
Possible Issue: It changes a bit how we apply damage and healing. Directly modifying the HP values still can bypass this change, so we would need a standard to only use the functions on the actor to manipulate HP values.

Fronix (changes requested):
## Suggestions

- `src/documents/actor/base.svelte.ts:68`, `CanvasScrollingTextInterface` is only used for a single cast inside `#emitHpChangeScrollingText`; consider inlining the cast or moving the type to a shared Foundry types file if reused later
- `src/documents/chatMessage.ts:39-42`, `HpMutableActor` wraps methods in `Partial` + runtime `typeof` checks, but all actors extend `NimbleBaseActor` which now always has these methods; the defensive checks are YAGNI
- `src/view/sheets/NPCSheet.svelte:11-24`, Functions `updateCurrentHP`/`updateMaxHP`/`updateTempHP` reference `actor` but are defined before `let { actor } = $props()` on line 26; style guide orders props (6) before functions (10)
- Merge conflicts

src/documents/actor/base.svelte.ts
```diff
@@ -329,11 +314,26 @@ class NimbleBaseActor<ActorType extends SystemActorTypes = SystemActorTypes> ext
...
+							anchor?: number;
+							jitter?: number;
+							fontSize?: number;
+							fill?: number;
+							stroke?: number;
+							strokeThickness?: number;
+						},
+					) => Promise<void>;
+			  }
+			| undefined;
```
Fronix, line 336: This could be cleaner but not blocking, keep in mind for next time to avoid large inline types.

Fronix (approved):
LGTM

PR #518, opened 2026-03-27
fix(ncsw): Preserve user targets when group attack panel refreshes

### LINKED ISSUE #498: Selecting a Token or changes in the combat tracker resets the targeted token
With the newest update there is a new "bug" in the system.

When you target a token (either via T, double rightclick as a player or use the target token option in foundry) the target resets once any other token is selected.

The same happens if the turnorder is manipulated in the combat tracker after a player has targeted something. It also resets once a player hits "End Turn"

If this behaviour is intended it should be reconsidered as it is, because frankly its annoying.

Insax:
- stop clearing `game.user.targets` in `hideGroupAttackPanel()` by default
- only clear targets when `clearTargets: true` is explicitly requested
- prevents target loss during combat tracker updates, canvas/token left-clicks, and other NCSW refresh/hide paths


Fronix (approved):
LGTM

PR #519, opened 2026-03-27
fix(combat): End turn action refill race

### LINKED ISSUE #472: End Turn sometimes sometimes does not restore hero actions
Sometimes the End Turn button does not grant actions back to heroes

Actions not being refilled for the hero after the end of the turn?

Insax:
This PR fixes an intermittent race condition where ending a turn could sometimes fail to refill hero actions (especially heroic reactions), because stale async combatant updates could land after _onEndTurn and overwrite refreshed values.
What changed
 - Added a per-combatant action mutation queue via queueCombatantActionMutation.
 - Updated requestAdvanceCombatTurn to wait for pending mutations on the active combatant before advancing the turn.
 - Routed relevant action/reaction updates through that queue:
   - sheet action tracker pip updates
   - heroic actions tab pip deductions
   - combat tracker action delta updates
   - useHeroicReactions and toggleHeroicReactionAvailability in combat document logic
   - Ensured attack/spell activation callbacks in the heroic actions tab properly await action deduction.
 - Follow-up policy cleanup:
   - added lang="ts" to PlayerCharacterHeroicActionsTab.svelte
   - added queueCombatantActionMutation() to the Shared Code Inventory in docs/STYLE_GUIDE.md

## Why this fixes it
Turn advancement is now ordered after all pending action mutations for the active combatant, so older writes can no longer overwrite post-end-turn refill values.

## Testing
 - Local manual race-condition testing was not performed, since this issue is intermittent and difficult to reproduce reliably in a local environment.


Fronix (changes requested):
## Suggestions

- `src/documents/combat/combat.svelte.ts:2`: Relative import path; prefer path alias `#utils/combatTurnActions.js` (pre-existing pattern in file, so low priority)
- `src/utils/combatTurnActions.ts:15`: Module-level mutable `Map` (`pendingCombatantActionMutations`) has no cleanup/reset mechanism; if a combat is deleted mid-mutation the map entry lingers until GC. Consider a `clearPendingMutations(combatId)` hook or WeakRef-based cleanup for long-running sessions.
- `src/utils/combatTurnActions.ts:35`: `queueCombatantActionMutation` is the 7th public export from this file. One-export-per-file rule is already violated by prior exports, but this deepens it.
- `src/view/sheets/components/ActionTracker.svelte.ts:132`: Inside `updateActionPips`, the mutation callback re-fetches the combatant via `combat.combatants.get(combatantId) ?? combatant`. The fallback to the outer `combatant` ref defeats the purpose of re-fetching fresh state (it may be stale). Consider returning early if `get()` returns `undefined`.
- `src/view/sheets/pages/PlayerCharacterHeroicActionsTab.svelte.ts:176`: Same stale-fallback pattern: `combat.combatants.get(combatantId) ?? combatant`.
- `src/view/ui/CtTopTracker.state.svelte.ts:306`: Same stale-fallback pattern: `combat.combatants.get(combatantId) ?? combatant`.

Fronix (approved):
Looks good now, `util` folder is getting a bit big with a lot of combatant/combat/minion things but that will be addressed in a separate PR.

Fronix:
I was unable to reproduce this locally, tried with remote foundry instance as well but didn't get it. I assume it's very dependent on connection or something but I can't see a reason not to add these safeguards.

PR #525, opened 2026-03-28
fix: Turn Tracking

### LINKED ISSUE #493: Expanding the monster list in the combat tracker breaks turn tracking
When expanding the monster list in the combat tracker, turn tracking becomes broken.

7D7D:
CODEX report:

Issue Summary
-   Reported issue: expanding the monster stack in the Combat Tracker makes turn tracking appear broken.
-   Scope: the custom top Combat Tracker UI, specifically the button that expands the monster stack into individual cards.

How I Investigated
-   Read the project guidance first to stay inside the repo's Svelte, Foundry, and testing conventions.
-   Traced the Combat Tracker flow through `CtTopTracker.state.svelte.ts`, `CtTopTrackerStore`, and `ctTopTracker/combat.utils.ts`.
-   Traced the real combat turn model through `documents/combat/combat.svelte.ts` and `combatTurns.ts`.
-   Compared the UI's visible entry construction against the actual normalized turn order used by the combat document.
-   Ran targeted tests during the investigation, then validated with the full repo check.

Key Findings
-   The underlying combat document already had a clear turn model.
-   `NimbleCombat.setupTurns()` removes dead combatants, normalizes grouped minions down to their effective leader, and expands solo monsters into repeated turns where appropriate.
-   The Combat Tracker UI rebuild path did not fully respect that normalized turn model.
-   `getCombatantsForScene()` in `ctTopTracker/combat.utils.ts` started with the normalized `combat.turns`, but then appended any other alive scene combatants that were not present in 'turns'.
-   For grouped minions, this was the critical mistake.
-   Grouped minion members are intentionally removed from the real turn order and represented by the leader's turn.
-   The CT rebuild path added those non-leader minions back into the visible list anyway.
-   When monster cards were collapsed, that mismatch was mostly hidden inside the single monster stack.
-   When monster cards were expanded, those extra minion members appeared as separate visible entries and made the tracker look like it had extra turns or a broken active position.

Root Cause
-   The expanded monster view mixed two different concepts:
-   combatants that are alive and visible in the scene
-   combatants that actually own initiative turns
-   Grouped minion members belong in the first category, but not the second.
-   The CT used both categories to build the expanded track, so the expanded UI could diverge from the real turn order.

What I Fixed
-   Updated `src/view/ui/ctTopTracker/combat.utils.ts`.
-   Added a grouped-minion exclusion step when rebuilding CT entries.
-   The CT now checks minion group summaries and excludes minion members that are not the effective alive leader.
-   This keeps the expanded monster view aligned with the real normalized turn order.
-   Non-grouped alive combatants still behave as before.
-   I did not change `NimbleCombat.setupTurns()` because the core combat turn model was already correct.

Follow-up Problem I Hit
-   During the first pass, I also tried to force `combat.combatant` to stay in sync by assigning to it directly.
-   That was a bad assumption for live Foundry.
-   In runtime, `Combat.combatant` is getter-only.
-   Foundry threw `TypeError: Cannot set property combatant of #<Combat> which has only a getter`.
-   I removed that mutation from both sync helpers.
-   I then adjusted the tests to model Foundry more accurately by using getter-based 'combatant' behavior instead of plain mutable object behavior.

Test Coverage Added
-   Added a helper-level regression in `src/view/ui/ctTopTracker/combatTracker.utils.test.ts`.
-   This test verifies that grouped minion members are excluded from CT turn-track reconstruction while other valid non-turn combatants are still preserved.
-   Added a store-level regression in `src/view/ui/ctTopTracker/topTrackerStore.svelte.test.ts`.
-   This test verifies that expanded monster mode does not surface grouped minion members as separate turn entries.
-   Added `src/utils/combatTurnSync.test.ts`.
-   This test locks in getter-based 'combatant' semantics so future changes do not repeat the same invalid assignment mistake.
-   Updated the CT sync expectations to match real Foundry behavior.

Files Changed
-   `src/view/ui/ctTopTracker/combat.utils.ts`
-   `src/view/ui/ctTopTracker/combatTracker.utils.test.ts
-   `src/view/ui/ctTopTracker/topTrackerStore.svelte.test.ts`
-   `src/utils/combatTurnSync.ts`
-   `src/utils/combatTurnSync.test.ts`

Verification
-   Ran targeted Combat Tracker and combat-document test suites while iterating.
-   Ran full `pnpm check` after the final fix.
-   Full validation passed:
-   formatting
-   linting
-   circular dependency check
-   type-checking
-   full test suite
-   Final full suite result: 648 passing tests.

Final Outcome
-   Expanding the monster stack no longer creates fake visible turn entries for grouped minion members.
-   The expanded Combat Tracker now reflects the real normalized initiative order instead of mixing turn owners with non-turn group members.
-   The temporary runtime error from the first patch attempt was removed, and the final implementation is compatible with Foundry's getter-only `combat.combatant` API.

### DIFF
--- a/src/utils/combatTurnSync.test.ts
+++ b/src/utils/combatTurnSync.test.ts
@@ -1,0 +1,66 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { createCombatActorFixture, createCombatantFixture } from '../../tests/fixtures/combat.js';
+import { clearExpandedTurnIdentityHint } from '../documents/combat/expandedTurnIdentityStore.js';
+import { syncCombatTurns } from './combatTurnSync.js';
+
+describe('syncCombatTurns', () => {
+	beforeEach(() => {
+		clearExpandedTurnIdentityHint('combat-turn-sync-active-combatant');
+	});
+
+	it('syncs combat.combatant to the normalized active turn identity', () => {
+		const playerOne = createCombatantFixture({
+			id: 'player-one',
+			type: 'character',
+			actor: createCombatActorFixture({ type: 'character' }),
+		});
+		const playerTwo = createCombatantFixture({
+			id: 'player-two',
+			type: 'character',
+			actor: createCombatActorFixture({ type: 'character' }),
+		});
+		const solo = createCombatantFixture({
+			id: 'legendary-one',
+			type: 'soloMonster',
+			actor: createCombatActorFixture({ type: 'soloMonster' }),
+		});
+
+		const expandedTurns = [playerOne, solo, playerTwo, solo] as Combatant.Implementation[];
+		const rawTurns = [playerOne, playerTwo, solo] as Combatant.Implementation[];
+		const combat = {
+			id: 'combat-turn-sync-active-combatant',
+			flags: {
+				nimble: {
+					expandedTurnIdentity: {
+						combatantId: 'player-two',
+						occurrence: 0,
+					},
+				},
+			},
+			turns: rawTurns,
+			turn: 0,
+			setupTurns: vi.fn(() => expandedTurns),
+		} as unknown as Combat & {
+			_nimbleExpandedTurnIdentity?: { combatantId: string; occurrence: number | null } | null;
+		};
+		Object.defineProperty(combat, 'combatant', {
+			configurable: true,
+			get() {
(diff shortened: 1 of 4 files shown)


trevlar (approved):
LGTM

PR #562, opened 2026-04-05
feat(rules): adjacency rule sync and Lionhearted boon automation

joechaotic:
## Summary
- Auto-track which tokens have the most adjacent enemies during combat, storing counts as actor flags
- Expose `enemiesAdjacent:N` and `enemiesAdjacent:most` derived tags on actors for use in rule evaluation
- Automate the Lionhearted boon: bearer gains +1 AC when they have the most adjacent enemies
- Add two world settings: **Auto-Track Token Adjacency** and **Include Diagonals in Adjacency**

## Changes
- **`adjacencySync.ts`** (new): Registers combat/movement hooks to sync adjacency counts across all tokens. Schedules a debounced sync on token movement and combat updates. Handles Foundry v13 timing (canvasReady fires before ready) by syncing immediately if canvas is already ready at hook-registration time.
- **`tokenAdjacency.ts`** (new): Pure utility - `countAdjacentEnemies` computes per-token enemy-adjacency counts and determines who has the most.
- **`adjacencySettings.ts`** (new): Registers the two world settings. The tracking setting has an `onChange` handler that clears all actor adjacency flags when tracking is disabled.
- **`base.svelte.ts`**: `_populateDerivedTags()` reads adjacency flags and adds `enemiesAdjacent:N` / `enemiesAdjacent:most` tags when tracking is enabled.
- **`armorClass.ts` / `lionhearted.json`**: Lionhearted boon now uses the `enemiesAdjacent:most` tag condition to grant +1 AC automatically.
- **`config.ts`**: Registers the new `adjacencyCondition` rule type.

## Test plan
- [x] `pnpm check` passes (format, lint, circular deps, types, tests)
- [x] With tracking enabled, token AC updates correctly as combatants move
- [x] Disabling tracking clears all adjacency flags on all actors
- [x] Re-enabling tracking and reloading restores correct adjacency counts
- [x] Lionhearted bearer gains +1 AC when they have the most adjacent enemies
- [x] Lionhearted bearer loses the bonus when they no longer have the most
- [x] `adjacencyIncludesDiagonals` setting correctly affects adjacency calculation
- [x] Tying for most adjacent enemies -- see note below

> **Note on tie behavior:** The test plan states tying for most adjacent enemies does not grant the bonus, but the current implementation grants it to all tied combatants (`hasMost = count > 0 && count === maxCount`). The correct behavior is pending Evan's answer in the [Nimble Core Rules doc](https://docs.google.com/document/d/1cZVpnkOlZYqBbdkv5HZniF6JFbI9wHGieKtmWnRHemc/edit?tab=t.azist1krv8hs#heading=h.kmsgc7zezq58) and will be addressed once he has had a chance to respond.

Co-authored with [Claude Code](https://claude.ai/claude-code)


vite.config.local.mts
trevlar, line 1: This file needs to be removed.

public/assets/icons/LICENSE.md
trevlar, line 1: This file needs to be reverted to the original state.

src/settings/adjacencySettings.ts
```diff
@@ -0,0 +1,44 @@
...
+	game.settings.register(
+		'nimble' as 'core',
+		ADJACENCY_SYNC_SETTING_KEY as 'rollMode',
+		{
+			name: 'NIMBLE.settings.autoTrackTokenAdjacency.name',
+			hint: 'NIMBLE.settings.autoTrackTokenAdjacency.hint',
+			scope: 'world',
+			config: true,
+			type: Boolean,
+			default: true,
```
trevlar: Since this is a new feature, let's default it to be turned off. Giving people the option to opt in if they want to try it. It would also give us time to test the feature for quality before having everyone run into issues.

`default: false,`

trevlar (changes requested):
A few small things to clean this up.

joechaotic:
@trevlar OK, thanks for all the feedback with these. It should be good to go.

vite.config.local.mts
trevlar: You'll need to `git rm vite.config.local.mts` to remove it from the staged/committed files and commit and push the change.

trevlar (dismissed):
One file still present that needs removed.

Fronix (changes requested):
## Suggestions

- `src/hooks/combatantHooks/adjacencySync.ts:39`, Tie behavior: `hasMost` is true for ALL combatants sharing the max count. The test plan states "Tying for most adjacent enemies does not grant the bonus", which implies a unique-maximum check.
  ```ts
  const hasMost = count > 0 && count === maxCount;
  ```
  If unique-max is intended, add a count of how many combatants share the max:
  ```ts
  const maxHolders = counts.filter((c) => c === maxCount).length;
  // then inside the map:
  const hasMost = count > 0 && count === maxCount && maxHolders === 1;
  ```
We might need to confirm this but I don't see it as a blocker.

- `src/hooks/combatantHooks/adjacencySync.ts:30`, `countAdjacentEnemies` is called without an explicit `includeDiagonals` argument, so it falls through to `getAdjacencyIncludesDiagonals()` at the function default. This works but hides the setting dependency inside a default parameter. Consider reading the setting once in `syncAdjacency` and passing it explicitly for clarity.

- `src/utils/tokenAdjacency.ts:20-31`, `tokenCenter` uses `token.document.width/height` to compute center, but only handles 1×1 tokens correctly for the distance thresholds. If larger tokens (2×2+) are ever used, the center-to-center distance will exceed the threshold even when edges touch. This makes the `tokenAdjacency` unusable for larger monsters if needed.

src/settings/adjacencySettings.ts
```diff
@@ -0,0 +1,44 @@
...
+	game.settings.register(
+		'nimble' as 'core',
+		ADJACENCY_SYNC_SETTING_KEY as 'rollMode',
+		{
+			name: 'NIMBLE.settings.autoTrackTokenAdjacency.name',
+			hint: 'NIMBLE.settings.autoTrackTokenAdjacency.hint',
+			scope: 'world',
+			config: true,
+			type: Boolean,
+			default: false,
```
Fronix, line 14: Adjacency sync hooks are registered once during `ready` based on the setting value at startup. Changing the setting at runtime doesn't register/unregister them.

Add `requiresReload: true` here to tell the GM that a reload is needed.

src/settings/adjacencySettings.ts
```diff
@@ -0,0 +1,44 @@
...
+	game.settings.register(
+		'nimble' as 'core',
+		ADJACENCY_INCLUDES_DIAGONALS_SETTING_KEY as 'rollMode',
+		{
+			name: 'NIMBLE.settings.adjacencyIncludesDiagonals.name',
+			hint: 'NIMBLE.settings.adjacencyIncludesDiagonals.hint',
+			scope: 'world',
+			config: true,
+			type: Boolean,
+			default: true,
```
Fronix, line 49: Same as above `requiresReload: true` is needed here

src/documents/actor/base.svelte.ts
```diff
@@ -265,7 +266,23 @@ class NimbleBaseActor<ActorType extends SystemActorTypes = SystemActorTypes> ext
...
+		const { enemiesAdjacentCount: count, hasMostAdjacentEnemies: hasMost } = adjacency;
+
+		if (typeof count === 'number' && count > 0) {
+			this.tags.add(`enemiesAdjacent:${count}`);
+		}
+
+		if (hasMost) {
+			this.tags.add(`enemiesAdjacent:${ADJACENCY_QUALIFIER.MOST}`);
+		}
+	}
```
Fronix, line 288: `_populateDerivedTags` reads `flags.nimble.adjacency` unconditionally, but these flags can persist from a prior session where `autoTrackTokenAdjacency` was enabled. When the setting is off, stale flags leak `enemiesAdjacent:most` into the domain, causing adjacency-predicated rules like Lionhearted to apply their +2 armor permanently.

You need to wrap the tag population behind the sync enabled check

```suggestion
	_populateDerivedTags(): void {
		if (getAdjacencySyncEnabled()) {
			const adjacency = this.getFlag('nimble', 'adjacency') as
				| { enemiesAdjacentCount?: number; hasMostAdjacentEnemies?: boolean }
				| undefined;

			if (adjacency) {
				const { enemiesAdjacentCount: count, hasMostAdjacentEnemies: hasMost } = adjacency;

				if (typeof count === 'number' && count > 0) {
					this.tags.add(`enemiesAdjacent:${count}`);
				}

				if (hasMost) {
					this.tags.add(`enemiesAdjacent:${ADJACENCY_QUALIFIER.MOST}`);
				}
			}
		}
	}
```

Fronix (changes requested):
After actually testing the feature i noticed two things that need to be fixed.

src/documents/actor/base.svelte.ts
```diff
@@ -267,6 +268,8 @@ class NimbleBaseActor<ActorType extends SystemActorTypes = SystemActorTypes> ext
 	}
 
 	_populateDerivedTags(): void {
+		if (!getAdjacencySyncEnabled()) return;
```
Fronix: I'd prefer if you wrapped it around the code, so next thing that goes in here isn't accidentally blocked by adjacency being disabled

Fronix (changes requested):

joechaotic:
@Fronix All your review feedback has been addressed — ready for re-review when you get a chance!

Fronix (approved):
At this stage I believe this is ready to merge.

@trevlar would like your 2 cents here as well

PR #571, opened 2026-04-06
feat(level-up): grant class features on level up and remove on level down

trevlar:
## Summary
- Grant class features automatically when leveling up
- Remove granted class features when leveling down
- Add epic boon selection when leveling up to level 19
- Improve level down dialog to show features that will be removed

## Changes
- **Class Feature Grants**: Class features are now automatically granted as embedded documents when leveling up. Features are tracked in level up history and removed when leveling down.
- **Epic Boon Selection**: At level 19, players can select an epic boon from the available options. The selected boon is granted and tracked for removal on level down.
- **Level Down Dialog**: Now displays features that were granted at that level and will be removed. Reduced red styling to only the warning message and confirm button.
- **New Components**: `EpicBoonSelection.svelte`, `LevelUpClassFeatureSelection.svelte`, `getEpicBoons.ts` utility

## Test plan
- [x] Level up a character and verify class features are granted
- [x] Level down and verify the granted features are removed
- [x] Level up a character to level 19 and verify epic boon selection appears
- [x] Select an epic boon and complete level up - verify boon is added to character
- [x] Level down from 19 and verify the epic boon is removed
- [x] Level down dialog shows features being removed without excessive red styling
- [x] Multiple accordions can be expanded simultaneously in epic boon selection

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -42,7 +42,7 @@ These are mandatory implementation constraints, not slogans.
 **KISS**: Use explicit control flow (`if`/`switch`). Avoid dynamic component lookups, runtime string-to-function dispatch, or Svelte action factories that obscure what runs and when.
 **YAGNI**: Every abstraction (utility, store, component) must have a current caller. If a code path isn't supported yet, throw an error, don't add a stub or no-op fallback.
 **DRY + Rule of Three**: Two similar blocks in the same file are fine. Extract to a shared utility or component only after the same pattern appears three times across different files, and only if the extracted code respects [Code Promotion Rules](#code-promotion-rules).
-**One Export Per File**: Each TypeScript module exports one public function, class, constant, or store. Internal helpers used only by that export stay in the file but are not exported. Svelte components are inherently one-per-file.
+**One Export Per File**: Each TypeScript module exports one public function (over 30 lines), class, or store. Related constants and types may share a file. Internal helpers used only by that export stay in the file but are not exported. Svelte components are inherently one-per-file.
 
 ### Foundational Principles
 
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -471,7 +471,8 @@
 			"hitDiceLabel": "Hit Dice:",
 			"rollHitDice": "Roll Hit Dice",
 			"rollHitDiceTooltip": "Roll your Hit Die with advantage and increase your max HP by that much.",
-			"takeAverage": "Take Average ({average})"
+			"takeAverage": "Take Average ({average})",
+			"featuresHint": "The following features are available at this level for your class."
 		},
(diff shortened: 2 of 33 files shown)


Fronix (changes requested):
## General

Clean PR, tested multiple classes, they all give stuff at level up. Confirmed some of them to be correct choices, I will not be testing all classes with all levels though, that will take too long to verify.

Below are some minor suggestions that should be quick to fix.

## Suggestions

- `src/view/dialogs/CharacterLevelUpDialogState.svelte.ts:236`, wrong setter type for `skillPointChanges` (stores numbers, typed `Record<string, null>`)
  ```ts
  set skillPointChanges(v: Record<string, null>) {
  ```
  Fix: use the return type of `generateBlankSkillSet()` (e.g. `Record<string, number>`).
- `src/view/dialogs/CharacterLevelUpDialogState.svelte.ts:96`, `featuresLoading` initialized `true` but the loading `$effect` only runs when `characterClass` is defined; characters without a class leave the dialog stuck in loading and `isComplete` permanently false.
  ```ts
  $effect(() => {
    if (!characterClass) return;
    ...
  ```
  Fix: set `featuresLoading = false` in the early-return branch.
- `src/documents/actor/character.ts:1243-1278`, duplicated "create embedded item + capture compendiumSource + collect ids" pattern for features and the epic boon.
  Fix: build a single `featuresToGrant` array including the boon and do one `createEmbeddedDocuments` call.
- `src/utils/getEpicBoons.ts:21,33`, `item as object as NimbleBoonItem` double cast is non-idiomatic.
  Fix: prefer `as unknown as NimbleBoonItem`.
- `src/view/dialogs/CharacterLevelUpDialog.test.ts:9,16`, relative imports `../../utils/getClassFeatures.ts` instead of path aliases.
  Fix: use `#utils/getClassFeatures.ts`.
- `src/view/components/ExpandableDocumentList.svelte.ts:38`, `displayedItems` collapses to only the selected item once chosen, hiding the deselect affordance for the others, intentional? If users want to switch selection they must deselect first; consider keeping the full list visible.
- `src/view/dialogs/CharacterLevelUpDialogState.svelte.ts:51`, subclass-load `$effect` re-runs only on `hasSubclassSelection`/`characterClass` identity; if `identifier` changes without the object identity changing, list won't refresh. Minor edge case.
- `src/migration/migrations/Migration014GrantedFeatureIds.ts:38`, skips items where `system.subclass` is truthy, silently dropping subclass-granted features from history. Confirm intended.
- `src/view/dialogs/CharacterLevelDownDialogState.svelte.ts:38`, `@ts-expect-error` for `CONFIG.NIMBLE.abilities`, consider adding the field to NIMBLE config types instead.

PR #572, opened 2026-04-06
fix(ui): use fixed portrait position on PC/NPC sheets

### LINKED ISSUE #489: Use fixed portrait image instead of cover image
The portrait image adjusts position as the PC sheet is resized. We should fix this so it doesn't change size/view.

To give perspective of the issue:

Image
Image

joechaotic:
## Summary

Fixes the PC/NPC portrait image shifting when the sheet is resized by switching from `object-fit: cover` to `object-fit: none`, anchoring the view to the top center.

## Changes

- Set `object-fit: none` on `.nimble-icon__image--actor` so the image renders at its natural size rather than scaling to fill the container
- Added `transform-origin: top center` so user-configured scaling (`--nimble-actor-image-scale`) anchors from the top
- Added `vite.config.local.mts` to `.gitignore` to prevent local config from being committed

## Test Plan

1. Open a PC or NPC sheet with a portrait image set
2. Resize the sheet wider and narrower
3. Confirm the portrait stays locked at the top and does not shift or reposition
4. Open the Settings tab and adjust the image scale/offset sliders — confirm they still work as expected

## Checklist

## Related Issues

Closes #489

### DIFF
--- a/src/scss/components/_image.scss
+++ b/src/scss/components/_image.scss
@@ -76,6 +76,8 @@
       top: var(--nimble-actor-image-y-offset, 0);
       left: var(--nimble-actor-image-x-offset, 0);
       scale: var(--nimble-actor-image-scale, 100%);
+      transform-origin: top center;
+      object-fit: none;
       object-position: top center;
       height: 100%;
     }
--- a/src/scss/components/_image.scss
+++ b/src/scss/components/_image.scss
@@ -14,8 +14,8 @@
 
   &--actor {
     position: relative;
-    height: 15rem;
     width: 100%;
+    max-height: 15rem;
     border-top-left-radius: 6px;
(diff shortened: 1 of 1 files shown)


Fronix (commented):
Note sure i'm a huge fan on this
Image

joechaotic:
@Fronix I would agree. My testing looked good. Probably because my image was smaller. I'll check to see if we can restrain it, etc.

joechaotic:
Hey @Fronix, pushed a small update — would you mind giving it another test when you get a chance? Thanks!

Fronix (approved):
Yeah much better, I can see this being the standard for now. Doesn't require users to special-crop images to fit the character sheet and token.

trevlar (approved):

PR #704, opened 2026-04-13
feat(level-up): grant subclass features on level up

### LINKED ISSUE #700: Support granting subclass features on level up
### Problem or Motivation

Characters can receive class features during level up, but subclass features are not currently granted through the same flow.

This makes subclass progression manual. Players and GMs have to remember which subclass features unlock at each level, find the correct items, and add them to the
actor themselves. That increases the chance of missed features and makes character advancement less consistent.

The main use case is a character with an existing subclass gaining a level in that class and receiving the subclass features unlocked at the new class level.

### Proposed Solution

When a character levels up in a class, the level-up flow should also check whether that class has a selected subclass.

If the subclass has features that unlock at the new class level, those features should be included alongside the normal class feature grants. They should use the
same review and grant behavior as existing class features wherever possible.

The flow should avoid adding duplicate subclass features if the actor already has them. If no subclass is selected, level up should still work normally and should
not throw an error.

### Alternatives Considered

Subclass features could remain fully manual, but that makes level-up automation incomplete and leaves important character progression easy to miss.

Subclass features could be copied directly into class progression data, but that would duplicate data and make subclasses harder to maintain.

This could also be deferred until a larger advancement overhaul, but subclass features are central to level progression and would provide immediate value in the
current level-up flow.

### Additional Context

Useful cases to account for:

- A character levels up in a class that already has a selected subclass.
- A character levels up in a class with no selected subclass.
- A subclass grants multiple features at the same level.
- The actor already has one or more of the subclass features being granted.
- A multiclass character levels up in one class while having subclasses on another class.

trevlar:
## Summary

- When leveling up a class, the level-up dialog now also grants any subclass features unlocked at the new level for the subclass the character has already selected.
- If no subclass is owned for the class being leveled, flow is unchanged (no-op). Handles multiclass: only subclasses whose `parentClass` matches the class being leveled are considered.
- Features flow through the existing `grantedFeatureIds` path, so level-down cleanly reverts them via the current machinery without further changes.

### Implementation

- New `src/utils/getSubclassFeatures.ts` — `buildSubclassFeatureIndex()` scans world items + compendium packs for features where `system.subclass === true`, indexing by `classIdentifier → subclassIdentifier (= slugified subclass name, matches feature \`group\`) → level`. `getSubclassFeaturesFromIndex()` resolves feature docs for a given class/subclass/level.
- `CharacterLevelUpDialogState.svelte.ts` — the existing class-features effect now also builds the subclass index, looks up the actor's embedded subclass for the class being leveled, and merges its level-N features into `autoGrant` (de-duplicated against already-owned features).
- No changes needed in `character.ts` or level-down logic — subclass features ride the same `grantedFeatureIds` path.

### Data notes

- Audited all 144 subclass feature JSONs across the 26 core subclasses: every file already has `subclass: true`, matching `class` + `group`, and a valid `gainedAtLevel(s)`. No pack JSON changes were needed.

Closes #700

## Test plan

- [ ] Unit tests (`src/utils/getSubclassFeatures.test.ts`, 8 cases) — index building, dedup, lookup, filtering, missing-UUID handling
- [ ] Manual: level up a Hunter past level 3 with Beastmaster selected → verify subclass features for the new level appear in the dialog and are granted on submit
- [ ] Manual: level up a class where no subclass is selected → verify no errors and no extra features
- [ ] Manual: multiclass character — level up class A while subclass belongs to class B → verify only A's (none, in this case) features are considered
- [ ] Manual: level down after granting subclass features → verify features are removed

### DIFF
--- a/packs/classFeatures/core/berserker/savage-arsenal/death-blow.json
+++ b/packs/classFeatures/core/berserker/savage-arsenal/death-blow.json
@@ -36,7 +36,7 @@
 		},
 		"description": "<p>After you deal damage from a crit, you may expend any number of Fury Dice. Sum the dice and deal double that amount of damage</p>",
 		"featureType": "class",
-		"class": "",
+		"class": "berserker",
 		"gainedAtLevel": 4,
 		"gainedAtLevels": [
 			4,
--- a/packs/classFeatures/core/berserker/savage-arsenal/deathless-rage.json
+++ b/packs/classFeatures/core/berserker/savage-arsenal/deathless-rage.json
@@ -36,7 +36,7 @@
 		},
 		"description": "<p>(1/turn) While Dying, you may suffer 1 Wound to gain 1 action.</p>",
 		"featureType": "class",
-		"class": "",
+		"class": "berserker",
 		"gainedAtLevel": 4,
 		"gainedAtLevels": [
 			4,
--- a/packs/classFeatures/core/berserker/savage-arsenal/eager-for-battle.json
+++ b/packs/classFeatures/core/berserker/savage-arsenal/eager-for-battle.json
@@ -49,7 +49,7 @@
 		},
 		"description": "<p>Gain advantage on Initiative. Move 2×DEX spaces for free on your first turn each encounter.</p>",
 		"featureType": "class",
-		"class": "",
+		"class": "berserker",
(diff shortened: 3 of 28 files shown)


Fronix (approved):
LGTM!

PR #709, opened 2026-04-16
feat(classFeatures): support multiple selections per group

### LINKED ISSUE #701: Support multiple selections per class feature group
### Problem or Motivation

Some class feature groups need to grant more than one selected feature from the same group at the same level.

Example: Hunter’s Thrill of the Hunt feature says the character chooses 2 Thrill of the Hunt abilities at level 2, then chooses one additional ability at levels 4,
6, 8, 12, and 14. The current class feature selection flow treats each non-progression group as “choose one,” so the user can only select one feature from thrill-of-
the-hunt during character creation or level-up.
Commanders combat tactics also has a similar multi-select requirement.

This leaves the actor missing valid class features unless the player manually adds the extra feature after the guided flow.

### Proposed Solution

The selection flow should:

- Allow a group to require more than one selected feature.
- Default existing selection groups to 1 selected feature so current content keeps working.
- Display clear copy such as Choose 2 and 1 of 2 selected.
- Let users toggle multiple feature cards in the same group until the required count is met.
- Prevent completing character creation or level-up until the required number of selections has been made.
- Grant every selected feature item from the group, not just one.
- Continue filtering out already-owned features during level-up.
- Leave *-progression auto-grant groups unchanged.

For content, Hunter thrill-of-the-hunt should require 2 selections at level 2, then 1 additional selection at levels 4, 6, 8, 12, and 14.

### Alternatives Considered

One workaround would be to split a multi-pick group into duplicate fake groups like thrill-of-the-hunt-1 and thrill-of-the-hunt-2. That would force the UI to show
two “choose one” groups, but it duplicates content structure, creates awkward headings, and does not scale cleanly for later levels.

Another option is to leave the second choice as a manual actor edit after creation or level-up. That avoids implementation work, but it undermines the guided flow
and makes it easy for characters to miss required class features.

Parsing the required count from feature descriptions was also considered, but that would be fragile and content-dependent. The required selection count should be
structured data.

### Additional Context

The current implementation appears to model selections as one feature per group: selectedFeatures is a Map<string, NimbleFeatureItem>, and the UI copy assumes
chooseOne. Supporting this will likely require the selected feature state and result payload to allow multiple selected items per group.

Relevant flows to check:

- Character creation class feature selection
- Level-up class feature selection
- Actor item grants from selected class features
- Level-up history / level-down cleanup
- Owned-feature filtering when choosing additional features later in a progression

trevlar:
## Summary

Closes #701

- Add `selectionCountByLevel` field to feature data model so groups can require more than one pick at a given level (defaults to 1 for backward compat)
- Change `selectedFeatures` from single-item to array-per-group across the character creation and level-up flows
- UI shows "Choose N" / "M of N selected" with card disable at cap
- Level-up dialog now includes class feature selection, filters already-owned features, and gates Submit on selection completeness
- Track `grantedFeatureIds` in levelUpHistory for clean level-down reversal
- Feature sheet exposes a "Selection Count By Level" input for world-level/homebrew features

### Content updated (6 groups)
| Group | Level | Picks |
|-------|-------|-------|
| Hunter thrill-of-the-hunt | 2 | 2 |
| Commander commanders-order | 2 | 2 |
| Mage spellshaper | 4 | 2 |
| Shepherd sacred-graces | 5 | 2 |
| Songweaver a-people-person | 5 | 2 |
| Stormshifter chimeric-boon | 6 | 2 |

### Out of scope
Stormshifter subclass "Circle of Fang & Claw" Master of Forms (2) at L15 grants 2 additional chimeric-boon picks conditionally — needs subclass-predicate infrastructure, tracked for a follow-up.

### Localization (flagged for review)
Added English strings: `chooseN`, `nOfMSelected`, `levelUpHint`

## Test plan
- [x] Character creation: Hunter at L1 — verify thrill-of-the-hunt shows "Choose 2" with progress counter, cards disable after 2 picks, Submit gates on completeness
- [x] Character creation: classes without multi-select (e.g. Berserker) still work as before (choose 1)
- [x] Level-up: Hunter L1→L2 — thrill-of-the-hunt appears with "Choose 2", grants both selected features
- [x] Level-up: Hunter L3→L4 — thrill-of-the-hunt shows remaining options (already-owned filtered), "Choose 1"
- [x] Level-down: reverting a level-up removes the granted feature items
- [x] Commander L1→L2 — commanders-order "Choose 2" works
- [x] Mage L3→L4, Shepherd L4→L5, Songweaver L4→L5, Stormshifter L5→L6 — multi-select verified
- [x] Feature sheet: "Selection Count By Level" input saves/loads correctly
- [ ] `pnpm check` passes (format, lint, circular-deps, type-check, 652 tests)


Fronix (changes requested):
Found some inconsistencies in leveling up, I think a general pass of all classes that are affected is warranted to avoid shipping simple bugs here.

## Mage
Invoker of Chaos showing up twice
Image

## Stormshifter
When leveling from 2->3 i get to choose spells from two schools but this should be from level 3->4
Image

Image

trevlar:
> ## Stormshifter
> When leveling from 2->3 i get to choose spells from two schools but this should be from level 3->4 Image
> 
> Image

That's not what I see. Deepening study at level 3 Utility spell at level 4

circle-of-sky-and-storm

Fronix:
> > ## Stormshifter
> > When leveling from 2->3 i get to choose spells from two schools but this should be from level 3->4 Image
> > Image
> 
> That's not what I see. Deepening study at level 3 Utility spell at level 4
> circle-of-sky-and-storm

True, forgot the subclass

Fronix (approved):
## Suggestions

- `src/view/sheets/FeatureSheet.svelte:20`, props lack a type annotation despite `lang="ts"` and the new `FeatureSheetProps` type living in `types/sheets/FeatureSheet.d.ts`. AGENTS.md mandates props types in `types/components/*.d.ts`-style files, not inline-untyped.
  ```svelte
  let { item, sheet } = $props();
  ```
  Fix: `import type { FeatureSheetProps } from '#types/sheets/FeatureSheet.d.ts';` and `let { item, sheet }: FeatureSheetProps = $props();`. (Also: align directory to `types/components/` per the documented convention, or update the convention.)


- `src/view/dialogs/CharacterLevelUpDialogState.svelte.ts:113`, `ownedFeatureUuids` mixes `item.uuid` (e.g. `Actor.X.Item.Y`) with `compendiumSource` (compendium UUID). The index lookup only ever matches against compendium UUIDs, so including `item.uuid` is dead weight.
  ```ts
  return [itemUuid, compendiumSource].filter(...)
  ```
  Fix: drop `itemUuid`; only `compendiumSource` participates in matching.
- `src/view/dialogs/components/characterCreator/FeatureGroupSelection.svelte.ts:2`, uses relative `'../../../../utils/sortDocumentsByName.js'` instead of `#utils/sortDocumentsByName.js` alias.
- `src/utils/getClassFeatures.ts:1246`, `resolveSelectionCount` silently picks the max across feature entries. If two pack files in the same group disagree on `selectionCountByLevel.{level}`, the lower count is dropped without warning. Consider `console.warn` on conflict to surface bad data, or document that the count must be set on every entry.
- `src/view/dialogs/components/characterCreator/ClassFeatureSelection.svelte.ts:42` and `src/view/dialogs/components/levelUpHelper/LevelUpClassFeatureSelection.svelte.ts:38`, `handleFeatureSelect` and the auto-select `$effect` are nearly byte-identical. Two callers today; if a third surfaces, extract to `featureSelectionToggle.ts`.
- `src/view/sheets/FeatureSheet.svelte:24`, `snippetsByTab` references snippets declared later in the template. Hoisting works in Svelte 5, but the order is jarring; consider declaring the snippets above the `<script>` epilogue or inlining the mapping where snippets are available.
- `src/view/sheets/FeatureSheet.svelte:124`, placeholder `"e.g. 2: 2  (overrides default of 1)"` is user-facing and not localized; existing strings on this sheet (e.g. headings) are also raw, pre-existing, but a good moment to wrap in `localize()`.
- `src/utils/getClassFeatures.test.ts` & `FeatureGroupSelection.svelte.test.ts`, only happy paths covered. Consider: feature count exceeds available options (cap behaviour), `selectionCountByLevel` missing for the current level (defaults to 1), and the auto-grant case where `features.length === selectionCount`.
- `src/view/sheets/FeatureSheetUtils.ts`, exports four functions in one module. Below the 30-line threshold so technically allowed, but worth considering whether `parseLevels` belongs with the existing utilities.

Fronix (approved):
LGTM

PR #728, opened 2026-04-21
fix(dialogs): apply savingThrowRollMode rules during character creation

### LINKED ISSUE #719: Celestial ancestry's "remove disadvantage on saves" benefit doesn't propagate from character creation to the PC sheet
### Foundry VTT Version

13.*

### Nimble System Version

0.8.3

### Description

The Celestial ancestry removes disadvantage on a saving throw as part of its ancestry benefits. When a character is created with the Celestial ancestry on the character creation screen, this benefit does not propagate to the finished character sheet — the affected save still shows as disadvantaged.

### Steps to Reproduce

1. Start creating a new character.
2. Select Celestial as the ancestry.
3. Finish character creation and open the resulting character sheet.
4. Inspect the modifiers listed for saves on the PC sheet.

### Expected Behavior

The save affected by the Celestial ancestry should no longer be listed as disadvantaged on the character sheet.

### Actual Behavior

The save is still listed as disadvantaged on the character sheet, as if the Celestial ancestry benefit were never applied.

### Additional Context

_No response_

trevlar:
## Summary

- Fixes `ReferenceError: resolveSavingThrowRollModes is not defined` thrown on character creation submit
- Replaces the hardcoded advantage/disadvantage assignment with a proper rule evaluation pass that mirrors the reset logic in `ActorSavingThrowConfigDialog`
- Reads `savingThrowRollMode` rules from all origin documents (class, ancestry, background), resolves targets against the current roll mode state, and respects choice-based rules (`requiresChoice`) using the player's `selectedAncestrySave`

Fixes #719 

## Test plan

- [ ] Complete character creation with a class that has advantage and disadvantage saves — verify both appear on the PC sheet
- [ ] Select a Flameborn (or any `requiresChoice: true`) ancestry — verify the chosen save gains advantage
- [ ] Select a Highborn (or any `target: "disadvantaged"`) ancestry — verify the class's disadvantage save is neutralised
- [ ] Complete character creation without an ancestry save choice — verify no unintended roll modes are applied

### DIFF
--- a/src/documents/dialogs/CharacterCreationDialog.svelte.ts
+++ b/src/documents/dialogs/CharacterCreationDialog.svelte.ts
@@ -10,6 +10,94 @@ import CharacterCreationDialogComponent from '../../view/dialogs/CharacterCreati
 
 const { ApplicationV2 } = foundry.applications.api;
 
+type SavingThrowRollModeRuleData = {
+	type: string;
+	disabled?: boolean;
+	priority?: number;
+	requiresChoice?: boolean;
+	selectedSave?: string | null;
+	target?: string;
+	mode?: string;
+	value?: number;
+};
+
+function resolveTargetSaves(
+	target: string,
+	selectedSave: string | null,
+	rollModes: Record<string, number>,
+	savingThrowKeys: string[],
+): string[] {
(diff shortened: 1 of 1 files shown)


Fronix (approved):
LGTM

PR #730, opened 2026-04-22
feat(sheets): search features, inventory, and spells by description text

### LINKED ISSUE #725: Features tab search should match against feature descriptions, not just titles
### Problem or Motivation

The search on the Features tab only matches against feature titles, which misses the common case of looking for features by what they *do* rather than what they're called. Feature titles are often evocative or class-flavored ("Second Wind," "Steady Aim") and don't always contain the mechanical keyword a player is actually searching for.

Concrete example: searching for "Field Rest" returns no results even when the character has features that trigger on a field rest, because "field rest" only appears in those features' descriptions, not their titles. The player ends up having to scroll through the full list and read every description manually — which is exactly what search is supposed to avoid.

This comes up whenever a player asks "which of my features care about X?" where X is a game-system keyword (rest types, damage types, conditions, action types, trigger phrases) that lives in description text rather than titles.

### Proposed Solution

Expand the Features tab search to match against both the title and the description of each feature. Title matches should probably rank higher than description-only matches in the result ordering so that a search for a feature by name still surfaces the exact feature first.

The match itself can remain simple (a case-insensitive substring match against the combined title + description text). If there's already a search index or filter function in use, this is ideally a one-field addition rather than a rewrite.

### Alternatives Considered

_No response_

### Additional Context

Worth checking while implementing:
- Whether the Inventory and Spell List tabs have the same title-only search limitation. If so, they likely want the same treatment, and it's probably the same shared search component.
- How the search handles rich-text / HTML in descriptions — matching should be against the rendered text, not against markup (a search for "rest" shouldn't false-positive on `<p class="rest-block">` or similar).
- Performance on characters with large feature lists is probably fine given typical counts (tens, not thousands), but worth a quick sanity check if the current implementation re-runs on every keystroke.

tristin-albers:
## Summary

Expand item search filters to match against item descriptions in addition to names, with HTML tag stripping so searches match rendered text.

## Changes

- Add description matching to `filterItems` for features, boons, inventory items, and spells
- Extend `AttackActionPanel` inline search to also match weapon/attack feature descriptions
- Extract shared `stripHtml` utility to `src/utils/stripHtml.ts`
- Add 11 unit tests covering name matching, description matching for all item types, HTML stripping, and edge cases

## Test Plan

- [x] Search Features tab for keyword in description (e.g. "field rest")
- [x] Search Inventory tab for keyword in item description
- [x] Search Spells tab for keyword in spell effect text
- [x] Search Actions tab for keyword in weapon/attack description

## Checklist

## Related Issues

Closes #725

### DIFF
--- a/src/view/dataPreparationHelpers/filterItems.ts
+++ b/src/view/dataPreparationHelpers/filterItems.ts
@@ -1,15 +1,41 @@
 import type { NimbleCharacter } from '../../documents/actor/character.js';
 import type { NimbleBaseItem } from '../../documents/item/base.svelte.js';
 
+const HTML_TAG_PATTERN = /<[^>]*>/g;
+
+function stripHtml(html: string): string {
+	return html.replace(HTML_TAG_PATTERN, '');
+}
+
+/** Description can be a plain HTML string, an object with string fields, or absent. */
+type ItemDescription = string | Record<string, string> | null | undefined;
+
+function getDescriptionText(description: ItemDescription): string {
+	if (typeof description === 'string') return stripHtml(description);
+	if (typeof description === 'object' && description !== null) {
+		return Object.values(description)
+			.filter((v): v is string => typeof v === 'string')
+			.map(stripHtml)
+			.join(' ');
+	}
+	return '';
+}
+
 export default function filterItems(
 	actor: NimbleCharacter,
-	requiredItemTypes: string[],
+	requiredItemTypes: string[] | string,
 	searchTerm: string,
 ): NimbleBaseItem[] {
+	const types = Array.isArray(requiredItemTypes) ? requiredItemTypes : [requiredItemTypes];
 	return actor.items.filter((item) => {
-		if (!requiredItemTypes.includes(item.type)) return false;
+		if (!types.includes(item.type)) return false;
 		if (!searchTerm) return true;
 
-		return item.name.toLocaleLowerCase().includes(searchTerm.toLocaleLowerCase());
+		const term = searchTerm.toLocaleLowerCase();
+		if (item.name.toLocaleLowerCase().includes(term)) return true;
+
+		const system = item.system as { description?: ItemDescription };
+		const descriptionText = getDescriptionText(system.description);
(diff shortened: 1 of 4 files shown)


src/view/dataPreparationHelpers/filterItems.ts
```diff
@@ -1,15 +1,41 @@
 import type { NimbleCharacter } from '../../documents/actor/character.js';
 import type { NimbleBaseItem } from '../../documents/item/base.svelte.js';
 
+const HTML_TAG_PATTERN = /<[^>]*>/g;
```
Fronix: This code is repeated in [AttackActionPanel.svelte.ts](https://github.com/Nimble-Co/FoundryVTT-Nimble/pull/730/changes#diff-8eccd43531310437f636e52169ea5be720bea1643afdb621466123ebea68c8ceR10-R14) move it to a utility file so we don't repeat code unnecessary

Fronix (changes requested):

src/view/dataPreparationHelpers/filterItems.ts
```diff
@@ -1,15 +1,41 @@
...
+
+function stripHtml(html: string): string {
+	return html.replace(HTML_TAG_PATTERN, '');
+}
```
Fronix: This code is repeated in AttackActionPanel.svelte.ts move it to a utility file so we don't repeat code unnecessary

Fronix (changes requested):

Fronix:
Also for future PRs please use the PR template

tristin-albers:
will do, sorry I missed that.

tristin-albers:
Committed abstraction for Html strip util. Also updated PR description to match the PR Template

Fronix (approved):
Looks good, I assume it's tested with successful results as I don't have the time to checkout and test this right now.

tristin-albers:
> Looks good, I assume it's tested with successful results as I don't have the time to checkout and test this right now.

yup i always manually test that checklist in the test plan :)

PR #732, opened 2026-04-22
feat(combat): decrement action tracker from all tabs on item activation

### LINKED ISSUE #723: Action tracker should decrement when actions are taken from the Inventory or Spell List tabs
### Problem or Motivation

The action tracker only ticks down when an action is used from the Actions tab. If a player casts a spell from the Spell List tab or uses an item from the Inventory tab, the action is resolved (attack rolls, damage, and effects all fire correctly), but the tracker doesn't register that an action was spent. The player ends up with a tracker that says they still have actions remaining when they don't, and has to either manually decrement it or mentally track around the UI.

This is especially easy to miss because the rest of the action flow works — the spell or item does its thing, so nothing *looks* broken until someone notices the tracker is out of sync a few rounds later. In practice, it means the Actions tab becomes the "correct" place to do anything that incurs action cost, which defeats the purpose of having the other tabs serve as functional entry points.

### Proposed Solution

The tracker-decrement logic should be tied to the action being taken, not to the tab from which it was initiated. Whatever hook currently fires when an Actions-tab entry is used should also fire when the equivalent action is triggered from Inventory (for usable items) or Spell List (for cast spells).

Concretely: find the shared path that resolves "an action was just spent" and make sure all three tabs route through it, rather than the Actions tab having its own decrement call that the other tabs skip.

### Alternatives Considered

- **Decrement in each tab's handler independently** — works but duplicates logic across three places and is the likely reason this drifted in the first place. Centralizing is cleaner.
- **Decrement only on dice roll/chat card post** — would catch most cases but miss actions that don't roll dice (utility spells, non-attack item uses).
- **Leave it and document "use the Actions tab for accurate tracking"** — rejected; the other tabs shouldn't be second-class.

### Additional Context

Repro steps:
1. Start a combat encounter so the action tracker is active with a full action budget.
2. From the **Actions** tab, use an action → tracker decrements as expected.
3. Reset/start a new turn.
4. From the **Spell List** tab, cast an action-cost spell → observe tracker does not decrement.
5. Repeat from the **Inventory** tab with a usable item that costs an action → observe tracker does not decrement.

tristin-albers:
## Summary

Centralize action cost deduction in `NimbleCharacter.activateItem()` so the action tracker decrements regardless of which tab initiates the activation. Previously only the Actions tab had decrement logic via a callback prop — Inventory and Spell List tabs skipped it entirely.

## Changes

- Override `activateItem` in `character.ts` to call `consumeCombatantAction` after successful activation with action cost
- Remove duplicate action cost logic from `AttackActionPanel` and `CastSpellActionPanel`
- Unarmed strike retains its own callback since it bypasses `activateItem`

## Test Plan

- [x] `pnpm check` passes
- [x] Actions tab — weapon attack decrements tracker
- [x] Spell List tab — casting a spell now decrements tracker
- [x] Inventory tab — using an action-cost item now decrements tracker
- [x] Unarmed strike from Actions tab still decrements
- [x] Outside combat, activating items from any tab causes no errors

## Checklist

## Related Issues

Closes #723

### DIFF
--- a/src/documents/actor/character.ts
+++ b/src/documents/actor/character.ts
@@ -13,6 +13,7 @@ import { HitDiceManager, incrementDieSize } from '../../managers/HitDiceManager.
 import { RestManager } from '../../managers/RestManager.js';
 import type { NimbleCharacterData } from '../../models/actor/CharacterDataModel.js';
 import calculateRollMode from '../../utils/calculateRollMode.js';
+import { consumeCombatantAction } from '../../utils/combatTurnActions.js';
 import getRollFormula from '../../utils/getRollFormula.js';
 import CharacterArmorProficienciesConfigDialog from '../../view/dialogs/CharacterArmorProficienciesConfigDialog.svelte';
 import CharacterLanguageProficienciesConfigDialog from '../../view/dialogs/CharacterLanguageProficienciesConfigDialog.svelte';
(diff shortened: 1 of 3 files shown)


tristin-albers:
Updated PR description to match PR Tempalte.

trevlar (changes requested):
All of the basics are working for me to deduct an action. Using a weapon from inventory and spells from the spells list, roll and deduct an action properly. As well as the heroic actions, everything seems to be working properly.

However, the reactions aren't working for me, I didn't see it create a roll, or deduct any action.

I confirmed that the reactions are broken in the latest dev branch.

trevlar (approved):
LGTM

PR #733, opened 2026-04-22
feat(combat): add temporary bonus action slots to the action tracker

### LINKED ISSUE #724: Allow adding temporary action dice to the action tracker beyond the default three
### Problem or Motivation

The action tracker on the right side of the character sheet shows the standard action count of 3, but there's no way to represent additional actions granted by features, items, or GM effects. When a character gets an extra action from some source, the player has to track it outside the sheet — on paper, in their head, or by asking the GM to remind them — which is exactly the kind of thing the tracker exists to prevent.

Right now, the tracker is a source of truth for the *base* action economy, but goes silent the moment anything exceeds it, which is a common enough situation in play that it's worth handling natively.

### Proposed Solution

Add a "+" affordance to the action tracker that adds a temporary action slot. Spending an action consumes slots in the normal way, so the temporary slot naturally disappears when used. No separate tracking for "which slot is temporary vs. base" is needed — if the player ends the turn with unused actions, the tracker resets to the base count on the next turn, and any unspent temporary slots go with it.

Beyond the manual "+" for ad hoc cases (GM grants, one-off effects), items and features that grant additional actions should use rules to automatically add slots. When a rule-bearing item or feature becomes active, it adds its granted slot(s) to the tracker; when it's removed, deactivated, or expires, the slots go with it. This keeps the manual "+" as the escape hatch for things that aren't (or can't be) modeled as rules, while making the common case — a known feature or magic item that grants an additional action — just work without the player having to remember to press a button.

Interaction model for the manual "+":
- **"+" button to add, consume normally to remove.** Simplest; matches the mental model of "granting" an action.
- **Shift+click or right-click on an unspent temporary slot to remove it.** Secondary affordance for the "added it by mistake" case.

Make the temporary slots visually distinguished from base slots (different color, subtle outline) so the player can see at a glance that they have an additional action this turn. Rule-granted slots can additionally show their source on hover.

### Alternatives Considered

- **Leave it to players to track externally** — current state; rejected because it's the motivating problem.
- **Use the combat tracker** — current state; not intuitive off of the PC Sheet. Good as a fallback/alternative available.

### Additional Context

_No response_

tristin-albers:
## Summary

Add a "+" button to the action tracker that grants temporary bonus action slots, capped at 10 total. Bonus pips render in gold to distinguish from base actions, and reset to zero when the character's turn ends.

## Changes

- Add `system.actions.base.bonus` field to `CharacterCombatantDataModel`
- Add `getCombatantBonusActions` and `getCombatantEffectiveMax` helpers
- Reset bonus to 0 on turn end in `combat.svelte.ts`
- Render bonus pips with gold styling and "+" button in `ActionTracker.svelte`
- Show effective max with golden border in `CtTopTracker.svelte`
- Localize bonus action button strings via `en.json`

## Test Plan

- [x] `pnpm check` passes
- [x] "+" adds a gold bonus pip, current increments
- [x] Spending actions grays out pips but bonus slots remain visible
- [x] Combat tracker shows golden border and golden max number
- [x] Ending turn resets bonus to 0 and pips return to base 3
- [x] Bonus persists across round boundaries
- [x] "+" button hides at 10 total pips

## Checklist

## Related Issues

Closes #724

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -735,6 +735,7 @@
 			},
 			"heroicActions": {
 				"title": "Heroic Actions",
+				"addBonusAction": "Add bonus action",
 				"reactionsTitle": "Heroic Reactions",
 				"reactions": {
 					"actionCost": "This will use 1 action (reaction)",
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -735,7 +735,7 @@
 			},
 			"heroicActions": {
 				"title": "Heroic Actions",
-				"addBonusAction": "Add bonus action",
+				"addAdditionalAction": "Add 1 action",
 				"reactionsTitle": "Heroic Reactions",
 				"reactions": {
 					"actionCost": "This will use 1 action (reaction)",
--- a/src/documents/combat/combat.svelte.test.ts
+++ b/src/documents/combat/combat.svelte.test.ts
@@ -2714,7 +2714,7 @@ describe('NimbleCombat', () => {
 
 		expect(combatant.update).toHaveBeenCalledWith({
 			'system.actions.base.current': 3,
-			'system.actions.base.bonus': 0,
+			'system.actions.base.additional': 0,
 			'system.actions.heroic.defendAvailable': true,
 			'system.actions.heroic.interposeAvailable': true,
 			'system.actions.heroic.opportunityAttackAvailable': true,
--- a/src/documents/combat/combat.svelte.test.ts
+++ b/src/documents/combat/combat.svelte.test.ts
@@ -2714,6 +2714,7 @@ describe('NimbleCombat', () => {
 
 		expect(combatant.update).toHaveBeenCalledWith({
 			'system.actions.base.current': 3,
+			'system.actions.base.bonus': 0,
 			'system.actions.heroic.defendAvailable': true,
 			'system.actions.heroic.interposeAvailable': true,
 			'system.actions.heroic.opportunityAttackAvailable': true,
--- a/src/documents/combat/combat.svelte.ts
+++ b/src/documents/combat/combat.svelte.ts
@@ -693,7 +693,7 @@ class NimbleCombat extends Combat {
 		if (combatant.type === 'character') {
 			await combatant.update({
 				'system.actions.base.current': getCombatantBaseActionMax(combatant),
-				'system.actions.base.bonus': 0,
+				'system.actions.base.additional': 0,
 				...this.#buildHeroicReactionAvailabilityUpdate(true),
 			} as Record<string, unknown>);
 		}
--- a/src/documents/combat/combat.svelte.ts
+++ b/src/documents/combat/combat.svelte.ts
@@ -693,6 +693,7 @@ class NimbleCombat extends Combat {
 		if (combatant.type === 'character') {
 			await combatant.update({
 				'system.actions.base.current': getCombatantBaseActionMax(combatant),
+				'system.actions.base.bonus': 0,
 				...this.#buildHeroicReactionAvailabilityUpdate(true),
 			} as Record<string, unknown>);
 		}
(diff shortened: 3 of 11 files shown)


tristin-albers:
Updated PR Description to match PR Template. Also committed a fix to change to use of the localize feature for the "add bonus action" language.

public/lang/en.json
```diff
@@ -735,6 +735,7 @@
 			},
 			"heroicActions": {
 				"title": "Heroic Actions",
+				"addBonusAction": "Add bonus action",
```
trevlar: It's not a bonus action. In all the books it says gain 1 action or additional action.

Let's change this to something similar like "Gain 1 action" or "Add 1 action"

src/documents/combat/combat.svelte.test.ts
```diff
@@ -2714,6 +2714,7 @@ describe('NimbleCombat', () => {
 
 		expect(combatant.update).toHaveBeenCalledWith({
 			'system.actions.base.current': 3,
+			'system.actions.base.bonus': 0,
```
trevlar: Let's change this to `system.actions.base.additional`

src/documents/combat/combat.svelte.ts
```diff
@@ -693,6 +693,7 @@ class NimbleCombat extends Combat {
 		if (combatant.type === 'character') {
 			await combatant.update({
 				'system.actions.base.current': getCombatantBaseActionMax(combatant),
+				'system.actions.base.bonus': 0,
```
trevlar: same

src/documents/combat/combatantSystem.ts
```diff
@@ -20,6 +20,15 @@ export function getCombatantBaseActions(combatant: Combatant.Implementation): Co
 	};
 }
 
+export function getCombatantBonusActions(combatant: Combatant.Implementation): number {
```
trevlar: getCombatantAdditionalActions

src/models/combatant/CharacterCombatantDataModel.ts
```diff
@@ -17,6 +17,13 @@ const nimbleCharacterCombatantSchema = () => ({
 				integer: true,
 				min: 0,
 			}),
+			bonus: new fields.NumberField({
```
trevlar: additional

trevlar (dismissed):
The action economy in Nimble is very different than what we read in D&D 5e. I want to eliminate any confusion between those. The Nimble books in particular never make mention of bonus action. So let's rethink what we name the new/additional actions granted to the character.

tristin-albers:
@trevlar that makes a lot of sense. I renamed those variables and functions above to be "additional" actions. Tested. Ready for your re-review when you get the chance. 

Thanks for the feedback!

src/view/sheets/components/ActionTracker.svelte
```diff
@@ -282,6 +353,16 @@
 		}
 	}
 
+	:global(.theme-dark) .action-tracker__add-additional {
+		background: hsl(220, 15%, 18%);
```
trevlar, line 357: additional action blue instead of yellow will look much better in light mode.

trevlar (changes requested):
A few items of feedback for the additional actions: can we make the dice pip blue instead of yellow? Similar to how we have a green HP bar and a blue temp HP bar? 

This will also make it look much better in light mode.

Also, these should be reproducible via CSS variables rather than hardcoded values. 


Set the yellow dice actions as blue.

Image

actionslightmode

src/view/sheets/components/ActionTracker.svelte
tristin-albers: makes sense, it's been updated to use blue and a css variable

trevlar (approved):
LGTM

PR #734, opened 2026-04-25
fix(reactions): unblock & confirm reactions on pc turn and macro bar

trevlar:
## Summary
- Reactions (Defend, Interpose, Help, Interpose & Defend, Opportunity Attack) no longer fail silently on the player's own turn or when dragged from the macro bar — `activeTurn` is now a soft block that the existing confirmation dialog can bypass alongside `noActions` and `spent`.
- The confirmation dialog stacks every applicable reason instead of picking one, with rulebook-matched wording: *"Reactions are performed when it is not your turn."* / *"Reactions cost 1 action and you have none remaining."* / *"Each reaction can be performed no more than 1/round, and you've already used X this round."*
- Fixes a pre-existing OA double-deduction (regression from #732): the panel now passes `skipActionDeduction: true` so the action is consumed once via `activateItem` rather than twice. The OA macro path no longer pre-spends — the panel's weapon click owns the deduction and reaction-spend.

## Test plan
- [ ] During your own turn, click Defend / Interpose / Help / Interpose & Defend — the modal explains "It is not your turn" and confirming rolls and posts to chat with one action deducted.
- [ ] On someone else's turn, all four panel reactions still work without a modal when available.
- [ ] Drag each reaction to the macro bar; trigger from the hotbar with each soft-block scenario (no actions, already spent, your turn, combination thereof).
- [ ] Drag Opportunity Attack to the macro bar; trigger from the hotbar — the OA panel opens and the chosen weapon deducts exactly one action (not two) and marks the reaction spent.
- [ ] Use OA from the panel directly during another's turn — same single-deduction outcome.
- [ ] \`pnpm check\` (lint + 998 tests + typecheck) is clean.


Fronix (approved):
LGTM

Issue #579, opened 2026-04-08
[Foundation] Predicate vocabulary expansion


Fronix:
Original scope was a `conditionalBonus` wrapper rule. We don't need one. Every rule already has a `PredicateField` and `Actor.getDomain()` already exposes tags like `armor:unarmored`, `enemiesAdjacent:most`, `class:berserker`. The gap is vocabulary, not engine.

Two pieces of work here: more `self:*` tags, and a `targetCondition` field on `damageBonus` for the cases where the predicate needs to look at the attack's target.

## Self-state tags

Populate in `Character.prepareDerivedData()` (and the matching paths on NPC / minion / solo actors where the tag makes sense):

| Tag | When |
| --- | --- |
| `self:bloodied` | HP ≤ 50% max |
| `self:dying` | HP at 0 with wounds remaining (mutually exclusive with `bloodied`) |
| `self:fullHp` | HP equals max |
| `self:concentrating` | Concentration flag set |
| `self:raging` | Pushed by the Rage activation (#577) |
| `self:beastshifted` | Pushed by Beastshift |
| `self:interposing` | Pushed by Interpose |
| `self:shield` | Shield equipped |
| `self:noShield` | No shield equipped |
| `self:noJudgmentDice` | Judgment Dice pool empty |
| `self:noToTH` | Thrill of the Hunt charges = 0 |
| `self:furyDiceMax` | Fury Dice pool at cap |
| `self:hasShadowMinion` | ≥1 active shadow minion |

The pool-state tags (`noJudgmentDice`, `noToTH`, `furyDiceMax`) actually belong to #576 / #578, since the counter rule is where the state lives. Listed here so the vocabulary is in one place; the publishing happens there.

## `targetCondition` on `damageBonus`

Some features need to gate on the *target*, not on self. Hunter's Mark (+damage to marked target), Hexbinder vs Afflicted / Smoldering, "vs Bloodied target" bonuses. That can't go through the existing predicate field because predicate is evaluated against the rule-owning actor's domain, not the target's.

Add a `targetCondition` field on `damageBonus`:

```jsonc
{
  "type": "damageBonus",
  "value": "@level",
  "delivery": "any",
  "source": "any",
  "targetCondition": { "target": "marked" }
}
```

Evaluated against the target's `getDomain()` at attack time using the existing `Predicate` class.

What to touch:

- `schema()` in `src/models/rules/damageBonus.ts`, add the field
[...]

Issue #358, opened 2026-02-27
Chat Card Core Requirements


Fronix:
To complete phase 1 we need to add the desired amount of functionality to the chat cards when in combat. We already have a lot of functionality but it needs to be cleaned up and possibly redesigned.

## What are we trying to solve?
- Unified damage and healing handling
Chat cards must support damage and healing using the same roll and resolution flow, with clear intent labels like apply damage or apply healing.

- Clear separation of Player UX and GM UX
Players only see common, relevant actions such as defend, while GMs have full control through toggles and overrides for edge cases.

- Per target control for the GM
GMs must be able to adjust outcomes per target, change the damage being applied, change targets

- Sensible defaults with manual override
The system assumes normal resolution by default, full damage, no modifiers, and allows manual changes rather than trying to auto solve every special case.

- Low clutter, expandable complexity
The chat card should stay visually simple, using accordions or expandable sections to hide advanced details while keeping them accessible.

---

### Below is stuff taken from previous discussions:

@DoubleKing-Prime:
**Very similar to the chat card we have now.** 
- I switched the positions of the Damage section and the Targets section. It seems like that is the logical process: the chat card doesn't appear until after the roll, so it's "damage" --> "who it's being applied to as a result".
- Top left is the GM view, with a number of small buttons, indicating full damage, no damage (immune or GM narrative reason), healing instead, etc. for each target to let the GM tailor damage if needed, before applying
- Top right is the player Dave's view. He just has the ability to Defend if he wants.
Bottom Right is a different player that ISN'T getting attacked. As we discussed, it is probably not worth tracking range for Interpose, so they have the option to Interpose for any of the targets. Clicking this would remove that target, and add you in it's place. Then it could change to reflect the Top Right card, with Defend as an option.

<img width="1068" height="1189" alt="Image" src="https://github.com/user-attachments/assets/237f9ced-db9c-4c08-b3d8-099191de3e37" />

---

### Inspiration
<div style="display: flex; flex-direction: column;">
<img width="301" height="595" alt="Image" src="https://github.com/user-attachments/assets/32ec02b4-b68d-4c48-8a79-6b6844ddba82" />
<img width="296" height="576" alt="Image" src="https://github.com/user-attachments/assets/c2930401-ce23-4bfd-b43d-cfe5bc07517e" />
</div>

PR #742, opened 2026-04-28
feat(rules): generalize meleeDamageBonus into damageBonus

### LINKED ISSUE #574: [Foundation] Rule type: damageBonus
## Purpose

Foundational rule type for flat or conditional bonus damage applied outside the melee-weapon path. `meleeDamageBonus` already exists for melee weapon attacks — this covers spell damage, ranged damage, and conditional damage bonuses.

## Proposed shape

```json
{
  "type": "damageBonus",
  "label": "Keen Eye, Steady Hand",
  "value": "@abilities.wis.mod",
  "appliesTo": "ranged",
  "condition": null
}
```

`appliesTo`: `ranged` | `spell` | `any` | specific damage type  
`condition`: optional predicate (later: ties into `conditionalBonus`)

## Blocks

Feature/spell issues that depend on this rule type will link here as `blocked by`. Known dependents:
- Hunter: Keen Eye, Steady Hand (+WIL to ranged damage)
- Hunter: Hunter's Mark (bonus damage to marked target)
- The Cheat: Sneak Attack scaling
- Various spell upcasts with +KEY damage riders

## Acceptance criteria

- [ ] Rule type registered in rule schema
- [ ] Data preparation pipeline applies bonus in correct damage calculation step
- [ ] Handles both static values and formulas (e.g. `@abilities.wis.mod`, `@details.level`)
- [ ] `appliesTo` filter correctly scopes bonus
- [ ] Unit tests cover all `appliesTo` branches
- [ ] Documented in rule-type reference

tristin-albers:
## Summary

Replaces the melee-only `MeleeDamageBonusRule` with a generalized `DamageBonusRule` that scopes bonuses along two independent axes — `delivery` (melee | ranged | any) and `source` (weapon | spell | any) — plus an optional `damageType` filter. Bonuses stack via an array accumulator and are consumed per damage node during item activation.

## Changes

- New `DamageBonusRule` with `value`, `damageType`, `delivery`, and `source` schema fields
- `getDamageBonusTotal()` utility in `attackUtils` centralizes consumption, filtering on delivery + source + damageType
- `ItemActivationManager` detects source from `item.type` and passes `node.damageType` per damage node
- Unarmed strike paths pass `delivery: 'melee', source: 'weapon', damageType: 'bludgeoning'`
- Migration016 converts `meleeDamageBonus` → `damageBonus` with `delivery: 'melee', source: 'weapon'`
- `LATEST_SCHEMA_VERSION` bumped to 16
- `MeleeDamageBonusRule` deleted
- Compendium: Reverberating Strikes, Keen Eye Steady Hand, and Fiery boon wired with `damageBonus` rules
- 26 new tests (rule lifecycle, utility filtering, real-world scenarios, integration tests)

## Design Validation

Every known dependent feature was validated against the schema:

| Feature | delivery | source | damageType | Correct? |
|---------|----------|--------|------------|----------|
| Reverberating Strikes (+LVL melee) | `melee` | `weapon` | `bludgeoning` | ✅ |
| Keen Eye (+WIL ranged weapon) | `ranged` | `weapon` | `""` | ✅ |
| Spell upcast +KEY | `any` | `spell` | `""` | ✅ |
| Storm Caster (+INT lightning spells) | `any` | `spell` | `lightning` | ✅ |
| Hunter's Mark (+LVL to marked target) | `any` | `any` | `""` | Deferred — needs conditional predicate (#579) |

**Known limitations (blocked by other foundation rules):**
- Hunter's Mark, Unending Judgment, Brave boon — need conditional predicates (#579)
- Sneak Attack, Combat Die bonuses — dice-based, different rule type
- Kinetic Barrage — cumulative event-triggered stacking

## Test Plan

- [ ] `pnpm check` passes (format, lint, circular deps, type-check, tests)
- [ ] Level 5+ Zephyr with Reverberating Strikes: unarmed strike includes `+ LVL`, melee spell (e.g. Snowblind) does NOT
- [ ] Hunter with Keen Eye: ranged weapon includes `+ WIL`, melee weapon does NOT, ranged spell does NOT
- [ ] Character with Fiery boon: `+1` applies to all attacks
- [ ] Multiple damageBonus rules stack correctly
- [ ] damageType filtering: lightning-only bonus does not apply to fire damage nodes

## Checklist

## Related Issues

Closes #574


Fronix (changes requested):
So I'm not sure the design solves the issue #574 exists to solve. Might been a miss in my initial audit of the needed automation rules but here's what I'm thinking.

`damageBonus` is meant to be **the** foundational rule for "this thing adds damage to an attack or a spell." That covers a lot of Nimble: **Hunter's Keen Eye Steady Hand** (+WIS to ranged weapon damage), **Hunter's Mark**, **Cheat-style Sneak Attack** scaling, **caster features that boost spell damage** (Empowered Evocation analogues, spell upcast +KEY riders, Mage school bonuses), **Zephyr's Reverberating Strikes**, and a long tail of ancestry and feat damage bumps. The point of building this as a foundation rule is that all of those should sit on top of it without bespoke code. So the question is: **can I write each of those features against this rule and get the right behavior?** For most of them, no.

In Nimble, every damaging attack has two independent properties. **Delivery** is melee or ranged (`attackType: 'reach' | 'range'`). **Source** is weapon or spell (`item.type`). Four real cells:

|  | Melee | Ranged |
|---|---|---|
| **Weapon** | Sword, unarmed | Bow, crossbow |
| **Spell** | Inflict Wounds, Shocking Grasp | Fire Bolt, Eldritch Blast |

The PR's `appliesTo: 'melee' | 'ranged' | 'spell' | 'any'` flattens both into one field, but `'melee'`/`'ranged'` describe delivery and `'spell'` describes source. **They're different axes.** The result is that you can't say "ranged weapon damage" without also buffing ranged spells, and you can't say "spell damage" at all because the activation code never checks `item.type`.

Concretely:

- **Keen Eye Steady Hand** written as `appliesTo: 'ranged'` buffs the Hunter's bow *and* their Fire Bolt the moment they pick up a cantrip.
- **Spell upcast +KEY riders** need `appliesTo: 'spell'`, but `getDamageBonusTotal` is only ever called with `'melee'` or `'ranged'`, so the `'spell'` value is **dead code**.
- **Empowered-Evocation-style features** ("+INT to any spell damage") have no expression: `'spell'` is dead and `'any'` would catch the caster's weapon swings too.
- **Hunter's Mark and Sneak Attack** need a `condition` predicate that's legitimately deferred.
- **Reverberating Strikes** works, and it's the only feature this PR exercises.

The fix is **splitting the two axes into separate fields**:

```ts
appliesTo: {
  delivery: 'melee' | 'ranged' | 'any',
  source:   'weapon' | 'spell' | 'any',
}
```

Every feature writes naturally. Reverberating Strikes is `{ delivery: 'melee', source: 'weapon' }`. Keen Eye Steady Hand is `{ delivery: 'ranged', source: 'weapon' }`. Spell upcast +KEY is `{ delivery: 'any', source: 'spell' }`. Hunter's Mark (once `condition` lands) is `{ delivery: 'any', source: 'any' }` plus the predicate. Existing `meleeDamageBonus` rules migrate to `{ delivery: 'melee', source: 'weapon' }`. The rule editor gets two dropdowns instead of one.

The other gap is **damage-type scoping** ("+1d6 fire to fire damage"). The `damageType` field is on the schema but the damage roll never reads it. Smaller fix, should be fixed here while we're in the same area.

---

### A worst-case worked example

Imagine a caster subclass feature: **Storm Caster**, *"Add your INT modifier to lightning damage you deal with spells."* That single feature exercises every dimension of the rule at once: a formula value, a damage-type filter, a source filter, and any delivery. Under the proposed shape it just works:

```json
{
  "id": "storm-caster-lightning-bonus",
  "type": "damageBonus",
  "label": "Storm Caster",
  "value": "@abilities.intelligence.mod",
  "damageType": "lightning",
  "appliesTo": {
    "delivery": "any",
    "source": "spell"
  }
}
```

What fires, what doesn't:

| Attack | Delivery | Source | Damage type | Bonus applies? |
|---|---|---|---|---|
| Lightning Bolt | ranged | spell | lightning | ✅ yes |
| Shocking Grasp (touch) | melee | spell | lightning | ✅ yes |
| Fire Bolt | ranged | spell | fire | ❌ wrong damage type |
| A lightning-enchanted longsword | melee | weapon | lightning | ❌ wrong source |
| Bow shot | ranged | weapon | piercing | ❌ wrong on both |

Under the current PR's enum there is no way to write this feature. `appliesTo: 'spell'` is dead code, `appliesTo: 'any'` would buff the caster's bow shots and weapon swings, and there's no damage-type filter at all so even if `'spell'` worked it would buff Fire Bolt too.

I think we can finish off this issue in this PR, all the automation foundation issues should probably not be "half done" to avoid more issues having to be created to finish them later, if it can be avoided.

The work that's left should be fine to finish off here:

- Reshape the schema into `delivery` + `source`
- Update the migration to match
- Fix the activation site to read both `attackType` and `item.type`
- Wire `damageType` through the damage roll
- Add integration tests covering ranged weapon, ranged spell, and melee spell

None of these are large on their own. Keeping them together lets us test the foundation end-to-end against real content: Reverberating Strikes is already in the pack, and we can stand up a Hunter ranged-weapon case alongside it.

tristin-albers:
I think I see what you did here, going forward I'll take the known dependents  and build out a rules table like that. That should help me identify if the design before the implementation more thoroughly for how it interacts with the rules. Your examples make a lot sense to me. Thank you for your extensive feedback, let me give another shot at this then. :)

tristin-albers:
Okay lets see if this better fits what we need here. Here's what I've done:

**Schema reshape**: Replaced appliesTo with two independent fields: delivery (melee | ranged | any) and source (weapon | spell | any). Every known dependent now expresses naturally:

- Reverberating Strikes: delivery: 'melee', source: 'weapon'
- Keen Eye, Steady Hand: delivery: 'ranged', source: 'weapon' — correctly excludes ranged spells
- Spell upcast +KEY: delivery: 'any', source: 'spell'
- Storm Caster example: delivery: 'any', source: 'spell', damageType: 'lightning'

**damageType wired end-to-end:** Moved the bonus calculation inside the damage node loop so each node's damageType is passed through getDamageBonusTotal. A lightning-only bonus won't apply to fire damage nodes.

**Source detection**: ItemActivationManager checks this.#item.type === 'spell' to determine source. Everything else (weapons, monster features) is classified as 'weapon'.

**Compendium updates**: Wired damageBonus rules on Keen Eye, Steady Hand and the Fiery minor boon as real-content validation alongside Reverberating Strikes.

**Integration tests**: Added 5 tests in ItemActivationManager.test.ts covering melee weapon, ranged weapon, ranged spell, melee spell, and damageType filtering.

**added combat dice bonuses**: Added combat dice bonuses to cover rolled dice damage bonuses for instance 

Known limitations (still blocked by other foundation rules):

- Hunter's Mark, Unending Judgment, Brave boon, Kinetic Barrage — all need conditional predicates (#579)
- Sneak Attack, just Combat Die bonuses (not wired) — While the formula for adding the rolled bonus damage is setup now, I still didn't want to wire it up as it has a condition of on crit. This is a good base, but will need 579 to complete it.

Fronix:
Better, this makes more sense. One thing I'm thinking of now is also that we should probably have some sort of way to send that these bonuses happen, in the chatcard or something.

I've had this issue in other systems where, stuff gets applied automatically but you can't really tell what applied it.

But maybe that is a different PR to work into the system overall instead so we can bench that for now.

tristin-albers:
yes I agree, I was looking to place the bonuses into parentheses in the roll of the chat card dropdown, but I couldn't pin down if it will only show roll math without text in there. Though I didn't find a good path with the time I had with this PR, I definitely agree, if you have time to set up an issue for it. I can take a swing at it with a deeper look.

Fronix (changes requested):
Looks good so far, there are just some few things I'd like cleaned up in the code to avoid small bugs.

## Weapons
**This can be done in a separate PR but needs to be done before this goes live:** We need to update and migrate weapons so they also work, I tested Reveberating Strike and it worked fine with unarmed strike but it did not work with weapons because they are missing `attackType`.

For every item with `system.objectType === 'weapon'`, set `system.activation.targets.attackType` to 'range' if `system.properties.selected includes 'range'`, otherwise 'reach'. Skip if attackType is already non-empty.

## Suggestions

- `src/models/rules/damageBonus.ts:114`: `if (resolvedValue === null || resolvedValue === 0) return;` lets negative resolved values through, but `getDamageBonusTotal` consumers gate on `> 0`, so a single negative bonus is silently dropped. Decide one direction (allow negatives end-to-end, or filter at rule-prep).
- `src/models/rules/damageBonus.ts:100-150`: `damageBonuses` array init + push is duplicated across the dice and numeric branches. Extract a small helper.
- `src/models/rules/damageBonus.ts:18`: `DICE_PATTERN = /\d*d\d+/i` matches any substring containing `dN`; an identifier or path containing e.g. `id6` would route to the dice branch. Consider stricter anchoring (`\bd\d+\b`).
- `src/managers/ItemActivationManager.ts:246`: when `attackType` is neither `'reach'` nor `'range'` the entire `if (delivery)` block is skipped, so a `delivery: 'any', source: 'any'` bonus (e.g. Fiery boon) won't apply to damage that isn't tagged as an attack. Confirm that's intentional or fall back to `'any'`.
- `src/managers/ItemActivationManager.ts:247`: `source = item.type === 'spell' ? 'spell' : 'weapon'` silently classifies `object`/`feature`/etc. as `'weapon'`. Tests rely on this (`mockItem.type = 'object'`). Worth a comment so the implicit mapping is documented.
- `src/config/registerRulesConfig.ts:47,75`: `damageBonus` was inserted at the old `meleeDamageBonus` slot (between `maxWounds` and `note`), breaking alphabetical order. Move both entries up to between `conditionImmunity` and `grantItem`.
- `public/lang/en.json:555`: same alphabetical-ordering nit; `damageBonus` should sit near the top of the `ruleTypes` block.
- `src/utils/attackUtils.ts:85-87,114-116`: JSDoc `@param delivery` says "'melee' or 'ranged'" but the type accepts `'any'` too. Update doc.
- `src/migration/migrations/Migration016DamageBonus.ts:30`: `delete rule.appliesTo` removes a field that never existed on `meleeDamageBonus`. Harmless but dead.

tristin-albers:
Thanks again for the well explained feedback Fronix! I went ahead and applied the feedback and suggestions from above.
                                                                                                                           
 Code cleanup:                                                                                                            
  - Negative values now filtered at rule-prep (<= 0)                                                                       
  - Extracted pushBonus() helper to deduplicate array init + push                                                          
  - Stricter DICE_PATTERN regex with negative lookbehind — won't match identifiers like id6                                
  - Documented the source classification (spell vs weapon) and empty attackType skip behavior in ItemActivationManager     
  - Fixed alphabetical ordering in registerRulesConfig and en.json                                                         
  - Updated JSDoc params to include 'any'                                                                                  
  - Removed dead delete rule.appliesTo from migration                                                                      
                                                                                                                       
Weapon migration:                                                                                                        
Added Migration017 to set attackType on all weapon items, 'range' if the weapon has the range property, otherwise       
'reach'. Updated weapon pack source files. Tested in Foundry, Staff now correctly receives the Reverberating     
Strikes bonus, and Shortbow on the Hunter gets the Keen Eye bonus.                                                       
                                                                                                                                                        
Confirmed empty attackType skipping damage bonuses is intentional when attackType is empty, the item isn't an attack (consumable, utility, etc.) and so it shouldn't receive damage bonuses. With the weapon migration in place, all weapons now have attackType set so this should only affect non-attack items now.                                                                                                
                                    
Side note:                                                                                                                           
I started exploring labeled roll terms (+ 5[Reverberating Strikes]) using Foundry's built-in [flavor] syntax in formulas, it works in the tooltip. But it opens up questions about damage type display vs filtering that need more thought, so I am holding off on that for now until we can talk more about it and spec it out in the future to keep this PR focused.
Mainly about if we want it to just be features, or damage bonuses in general and how that looks (i.e. +5[Bludgeoning, Reverberating Strikes] vs +5[Reverberating Strikes]). I'd be willing to take it on as a follow up though.

Fronix:
> Thanks again for the well explained feedback Fronix! I went ahead and applied the feedback and suggestions from above.
> 
> Code cleanup:
> 
>     * Negative values now filtered at rule-prep (<= 0)
> 
>     * Extracted pushBonus() helper to deduplicate array init + push
> 
>     * Stricter DICE_PATTERN regex with negative lookbehind — won't match identifiers like id6
> 
>     * Documented the source classification (spell vs weapon) and empty attackType skip behavior in ItemActivationManager
> 
>     * Fixed alphabetical ordering in registerRulesConfig and en.json
> 
>     * Updated JSDoc params to include 'any'
> 
>     * Removed dead delete rule.appliesTo from migration
> 
> 
> Weapon migration: Added Migration017 to set attackType on all weapon items, 'range' if the weapon has the range property, otherwise 'reach'. Updated weapon pack source files. Tested in Foundry, Staff now correctly receives the Reverberating Strikes bonus, and Shortbow on the Hunter gets the Keen Eye bonus.
> 
> Confirmed empty attackType skipping damage bonuses is intentional when attackType is empty, the item isn't an attack (consumable, utility, etc.) and so it shouldn't receive damage bonuses. With the weapon migration in place, all weapons now have attackType set so this should only affect non-attack items now.
> 
> Side note: I started exploring labeled roll terms (+ 5[Reverberating Strikes]) using Foundry's built-in [flavor] syntax in formulas, it works in the tooltip. But it opens up questions about damage type display vs filtering that need more thought, so I am holding off on that for now until we can talk more about it and spec it out in the future to keep this PR focused. Mainly about if we want it to just be features, or damage bonuses in general and how that looks (i.e. +5[Bludgeoning, Reverberating Strikes] vs +5[Reverberating Strikes]). I'd be willing to take it on as a follow up though.



> Thanks again for the well explained feedback Fronix! I went ahead and applied the feedback and suggestions from above.
> 
> Code cleanup:
> 
>     * Negative values now filtered at rule-prep (<= 0)
> 
>     * Extracted pushBonus() helper to deduplicate array init + push
> 
>     * Stricter DICE_PATTERN regex with negative lookbehind — won't match identifiers like id6
> 
>     * Documented the source classification (spell vs weapon) and empty attackType skip behavior in ItemActivationManager
> 
>     * Fixed alphabetical ordering in registerRulesConfig and en.json
> 
>     * Updated JSDoc params to include 'any'
> 
>     * Removed dead delete rule.appliesTo from migration
> 
> 
> Weapon migration: Added Migration017 to set attackType on all weapon items, 'range' if the weapon has the range property, otherwise 'reach'. Updated weapon pack source files. Tested in Foundry, Staff now correctly receives the Reverberating Strikes bonus, and Shortbow on the Hunter gets the Keen Eye bonus.
> 
> Confirmed empty attackType skipping damage bonuses is intentional when attackType is empty, the item isn't an attack (consumable, utility, etc.) and so it shouldn't receive damage bonuses. With the weapon migration in place, all weapons now have attackType set so this should only affect non-attack items now.
> 
> Side note: I started exploring labeled roll terms (+ 5[Reverberating Strikes]) using Foundry's built-in [flavor] syntax in formulas, it works in the tooltip. But it opens up questions about damage type display vs filtering that need more thought, so I am holding off on that for now until we can talk more about it and spec it out in the future to keep this PR focused. Mainly about if we want it to just be features, or damage bonuses in general and how that looks (i.e. +5[Bludgeoning, Reverberating Strikes] vs +5[Reverberating Strikes]). I'd be willing to take it on as a follow up though.

Alright, this looks good to me now. Regarding the labels we can deal with that later or part of #358

Fronix (approved):

PR #767, opened 2026-05-14
feat(rules): add dicePool and modifyPool rules

Fronix:
> **📚 Review order — this PR is the foundation of a 3-PR stack:**
>
> 1. **#767 (this PR)** — `dicePool` / `modifyPool` rule types + helpers + chargePool dieSize hint. Foundation only.
> 2. **#768** — pack content (Rage / Judgment / Combat Dice), refill/recovery triggers, `DicePoolTracker` UI, fixes uncovered in smoke-testing.
> 3. **#769** — `pool` effect node for activations, dialog-side dice-spend, `autoBonus` consumption mode (Fury Dice "always-on" semantics).
>
> Each PR is independently reviewable but they merge top-down.

## Summary

Adds two foundation rule types: `dicePool` for rolled dice (where each die in the pool stores a rolled face) and `modifyPool` for level-up overrides. Distinct from the existing `chargePool`, which tracks counts. Also adds an optional `dieSize` hint on chargePool so charge resources with a roll-on-spend mechanic (Combat Dice, Mana Dice) can declare their die size at the schema level. Covers the data-model half of #578; consumer wiring (Fury Dice, Judgment Dice activations) and sheet UI are intentionally out of scope here — see #768 and #769 for those.

## Changes

- New rule type `dicePool`. Schema declares `dieSize`, `max` (formula), `initial`, and a `refills` array reusing the `chargePool` trigger vocabulary. Runtime state lives at `flags.nimble.dicePools.<identifier>` as `{ faces: number[] }`.
- New rule type `modifyPool` with a `poolType: 'dice' | 'charge'` discriminator. Sibling rule attached to level-up features that overrides another pool's `dieSize` and/or adjusts its `max` additively (`maxDelta: '+1'`). One rule targets either subsystem, avoiding a parallel `modifyDicePool` / `modifyChargePool` pair. Matches how Nimble rulebooks author progression (one feature per breakpoint) rather than baking a level table into the pool rule.
- Optional `dieSize` field on chargePool (nullable). Pure metadata: pools with a dieSize are roll-on-spend resources; pools without are pure counters. The sheet UI later uses this to pick the right tracker presentation.
- Domain tag publishing. `self:no<Id>Dice`, `self:<id>DiceMax`, `self:<id>DicePool:N` emitted from `_populateDerivedTags` so predicates can gate on pool state (aligns with #579 vocabulary).
[...]

Issue #578, opened 2026-04-08
[Foundation] Rule type: dicePool


Fronix:
## Purpose

Foundational rule type for a variable pool of dice that scales with level, is spent during play, and refreshes on a specified interval.

## Proposed shape

```json
{
  "type": "dicePool",
  "label": "Fury Dice",
  "dieSize": "d6",
  "count": "@details.level",
  "recharge": "shortRest",
  "spendable": true
}
```

## Blocks

- Berserker: Fury Dice
- Commander: Combat Dice
- Oathsworn: Judgment Dice
- Mage: Mana Dice
- Songweaver: Inspiration pool

## Acceptance criteria

- [ ] Rule type registered
- [ ] Pool UI renders with current/max dice
- [ ] Spend API deducts and logs to chat
- [ ] Recharge lifecycle
- [ ] Formula-driven `count` evaluates on actor update
- [ ] Unit tests per pool type

Fronix:
Re: #579, this rule type inherits `predicate` from `NimbleBaseRule`, so it already gets all the gating from the #579 vocabulary for free. Don't add a bespoke `condition` field; just use predicate.

Same point as on #576: publish empty / full / count tags into the domain. This is where `self:noJudgmentDice`, `self:furyDiceMax` etc. actually come from, even though #579 lists them in its vocabulary table.

trevlar:
It also increases as they level up. The max per dice pool varies depending on the class.

PR #768, opened 2026-05-15
feat(rules): dice/charge pool content, refill triggers, tracker UI

Fronix:
> **Stacked on top of #767.** Merge that first.

## Summary

Extends the dice/charge pool foundation from #767 with the data, hooks, and UI needed to make pools actually do something in play. Adds pack content for Rage / Intensifying Fury / Radiant Judgement / Fit for Any Battlefield, refill/recovery triggers driven by Nimble's existing event hooks, the on-sheet `DicePoolTracker` UI, and three fixes uncovered during smoke-testing.

## Changes

- **Pack content**: dicePool / chargePool / modifyPool rules on Rage, Intensifying Fury, Radiant Judgement, Fit for Any Battlefield. **Migration020** backfills the same rules onto already-imported copies (idempotent on rule id). (Numbered after `dev`'s `Migration019SearingLightDisposition`.)
- **Refill / recovery triggers**: `onAttacked` (drives Oathsworn Judgment Dice off `nimble.damageApplied`), `onInitiativeRolled` (drives Commander Combat Dice off `nimble.initiativeRolled`).
- **DicePoolTracker** Svelte component on the character sheet — renders rolled pools as die-chip rows and roll-on-spend charge pools as pip rows, with empty-pool placeholder badges.
- **Bug fixes**:
  - `@key` / `@strength` flat keys on Rage and Fit for Any Battlefield (the original `@abilities.key.value` / `@strength.mod` resolve to undefined, clamping pool max to 0).
  - `Combat#rollInitiative` now emits `nimble.initiativeRolled` for each combatant — the actor-level emit only fires out-of-combat, so the recovery trigger never fired in real play.
  - `DicePoolTracker` subscribes to pool change hooks via `createSubscriber` so the derivation re-runs when flag state changes mid-session (previously stale until sheet reopen).
- **Drops** the rule-level `onItemActivated` hook the original branch added — activation belongs in the activation/effects pipeline (next PR), not in rules.

## Test Plan

1. Pull the branch on a fresh world.
2. Drop **Rage** onto a Berserker. Tracker shows Fury Dice empty badge with the correct max (`@key` modifier).
3. Level the Berserker to 6 → Intensifying Fury swaps the badge icon to d6.
4. Drop **Radiant Judgement** onto an Oathsworn. Deal any damage to them → Judgment Dice fills to 2d6 (`onAttacked` trigger).
5. Drop **Fit for Any Battlefield** onto a Commander. Roll initiative in an active combat → Combat Dice fills to STR/STR-mod d6 pips.
6. Open the manual pool roll/expend UX on the tracker — single-die chips spend individually; "expend all" button sums the rolled total.
7. Confirm no console errors when applying conditions / using items / rolling initiative.

## Checklist

- [ ] PR targets the `dev` branch — _intentionally stacked on #767; will effectively target dev after #767 merges_

## Related Issues

Part of #578.


src/view/sheets/components/DicePoolTracker.svelte
```diff
@@ -0,0 +1,287 @@
...
+			cursor: pointer;
+
+			&:hover {
+				color: var(--dp-expend-hover);
+			}
+		}
+	}
+
+	:global(.theme-dark) .dice-pool-tracker {
+		--dp-available: hsl(210, 80%, 63%);
```
trevlar: Colors will need to be added to theme appropriately and use variables instead of hardcoded values in the styles.

src/view/sheets/components/DicePoolTracker.svelte
```diff
@@ -0,0 +1,287 @@
...
+					>
+						<i class="fa-solid {getDieFaceIcon(pool.dieSize)}"></i>
+					</div>
+
+					{#each pool.faces as value, i (i)}
+						<button
+							class="dice-pool-tracker__die-chip"
+							type="button"
+							aria-label="Spend {pool.label} die (value: {value})"
+							data-tooltip="Rolled {value} – Click to spend"
```
trevlar: Number of labels that will need to be added to i18n

trevlar (changes requested):
i18n and CSS colors added to theme with variable names used in styles.

Must-fix:

- rollDieIntoPool and rollPoolFresh have no current callers. Both are exported from dicePoolRefill.ts and documented as "used by sheet UI for activation-style..." but no caller exists in this PR —
activation is deferred to the next PR. AGENTS.md explicitly states YAGNI as a mandatory constraint: "Every abstraction must have a current caller. If a code path isn't supported yet, throw an error,
don't add a stub." These should be dropped and introduced alongside their first caller in the activation PR.
- The appliesTo() predicate-gating branch in getDicePoolModifiers and getChargePoolModifiers has no test coverage. The two helpers.test.ts mock rules are plain objects with no appliesTo method, so the
typeof ruleWithApplies.appliesTo === 'function' guard silently no-ops in every existing test. The level-gating predicate check (e.g., "only apply Intensifying Fury modifiers at level 6+") is the most
load-bearing new behavior in this PR — it needs a test with a mock rule that has appliesTo returning false to verify the modifier is actually skipped.
- initiativeTrigger.ts and attackedTrigger.ts have no tests. The existing chargePoolTriggers.test.ts covers wound/turn/kill/bloodied triggers with a consistent pattern (register, capture hook, verify
registration, verify non-character guard). The two new triggers should follow the same pattern.
- DicePoolTracker.svelte props are untyped. let { actor } = $props() has no type annotation and no corresponding types/components/DicePoolTracker.d.ts file. AGENTS.md and the style guide both require
props types to be defined in types/components/*.d.ts.
- @ts-expect-error comments in both trigger files say only "Custom hook" — AGENTS.md requires a comment explaining why the suppression is needed, not just what it is. The combat.svelte.ts version
("consumed by ruleEventDispatch and chargePool/dicePool triggers") sets the right standard; the trigger files should match it.

trevlar (approved):

PR #773, opened 2026-05-16
feat(spells): deduplicate spell index by signature, prefer default Nimble compendium

trevlar:
## Summary

- `buildSpellIndex` now sorts packs so `nimble.nimble-spells` is processed first; copy packs with different UUIDs but identical spells are skipped via an `identifier+tier+school` signature check
- World item signatures are also tracked so compendium entries do not re-appear alongside a world spell with the same characteristics
- `collectSpellGrants` gains an optional `ownedSpellSignatures` parameter; the level-up dialog builds and passes this set so actors that received spells from a copy pack are not re-granted those spells at level-up
- `SpellIndexEntry` gains an `identifier` field (carries `system.identifier` when set, otherwise empty string)

## Test plan

- [ ] All 1097 existing tests pass
- [ ] New `buildSpellIndex` tests: signature dedup across packs, nimble-wins-over-copy, identifier stored in entry
- [ ] New `spellSignature` unit tests: identifier vs name fallback, whitespace trimming
- [ ] New `collectSpellGrants` test: `ownedSpellSignatures` prevents re-grant of copy-pack spells
- [ ] Manual: create a character with a copy of the Nimble spell compendium present — each spell should appear only once in the grant list

### DIFF
--- a/src/documents/dialogs/CharacterCreationDialog.test.ts
+++ b/src/documents/dialogs/CharacterCreationDialog.test.ts
@@ -25,6 +25,7 @@ function createSpellEntry({
 		school,
 		tier,
 		isUtility,
+		identifier: '',
 		classes,
 	};
 }
--- a/src/utils/getSpells.test.ts
+++ b/src/utils/getSpells.test.ts
@@ -1,7 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 import type { SpellIndex, SpellIndexEntry } from './getSpells.js';
-import { buildSpellIndex } from './getSpells.js';
+import { buildSpellIndex, spellSignature } from './getSpells.js';
 import { getSpellsFromIndex } from './getSpellsFromIndex.js';
 
 interface TestSpellSource {
@@ -25,6 +25,7 @@ function createIndexedSpell({
 	school = 'fire',
 	tier = 0,
 	isUtility = false,
+	identifier = '',
 	classes = [],
 }: {
 	uuid: string;
@@ -32,6 +33,7 @@ function createIndexedSpell({
 	school?: string;
 	tier?: number;
 	isUtility?: boolean;
+	identifier?: string;
 	classes?: string[];
 }): SpellIndexEntry {
 	return {
@@ -41,6 +43,7 @@ function createIndexedSpell({
 		school,
 		tier,
 		isUtility,
+		identifier,
 		classes,
 	};
 }
@@ -170,7 +173,13 @@ describe('buildSpellIndex', () => {
 		const fireCantrips = index.get('fire')?.get(0) ?? [];
 
 		expect(itemPack.getIndex).toHaveBeenCalledWith({
-			fields: ['system.school', 'system.tier', 'system.classes', 'system.properties.selected'],
+			fields: [
+				'system.school',
+				'system.tier',
+				'system.identifier',
+				'system.classes',
+				'system.properties.selected',
+			],
 		});
 		expect(fireCantrips.map((spell) => spell.name)).toEqual([
 			'Class Bolt',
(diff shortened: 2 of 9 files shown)


Fronix:
@trevlar is this to avoid duplicate spells in level up? If so, I think we should instead allow for selecting compendium sources that the level up system picks from, if we start filtering stuff out based on these criteria's we might inadvertently filter out hombrew stuff which would surface as weird bugs and hard to debug as well.

trevlar:
> @trevlar is this to avoid duplicate spells in level up? If so, I think we should instead allow for selecting compendium sources that the level up system picks from, if we start filtering stuff out based on these criteria's we might inadvertently filter out hombrew stuff which would surface as weird bugs and hard to debug as well.

Character creation and level up actually. I have a task to make it so they can pick and choose the source if there are multiple.

Fronix (changes requested):
## Blocking

- `src/utils/getSpells.ts:194`, Signature dedup silently hides homebrew spells that share name+tier+school with a Nimble spell. **No spell in `packs/spells/` currently sets `system.identifier`** (verified via grep, 0 matches), so the signature collapses to `name.toLowerCase().trim() + tier + school` for every existing Nimble spell. Any homebrew pack with a spell named "Fireball" / "Light" / "Heal" at the same tier and school is dropped from the index, and therefore invisible during character creation and level-up grants.

  ```ts
  // Skip if another pack (or a world item) already contributed a spell with this
  // identifier+tier+school combination.  Processing the default Nimble pack first
  // guarantees its entries take priority over any copy packs.
  if (seenSignatures.has(sig)) continue;
  ```

  Three concrete failure modes:
  1. **Homebrew variant with reused name** (e.g. GM's reflavored "Fireball" cantrip in school `fire`, tier `0`), silently hidden behind the Nimble spell of the same name/tier/school.
  2. **Intentional GM override** (homebrew pack rebalancing a core spell with same identifier), dropped instead of overriding.
  3. **Two unrelated spells that happen to share name** across modules, second one disappears.

  Fix options (pick one):
  - Scope dedup to packs whose collection matches the copy-pack pattern (e.g. only dedupe against `nimble.nimble-spells` when the other pack's spells claim `compendiumSource` pointing at `nimble.nimble-spells.*`). This addresses the real bug (copy packs) without touching unrelated homebrew.
  - Require a non-empty `identifier` for dedup, drop the name-fallback branch in `spellSignature` so spells without identifiers never collide. Then backfill `system.identifier` on the Nimble pack so the original copy-pack scenario still works.
  - Detect copy packs by comparing the pack's spell `_stats.compendiumSource` to the Nimble pack UUID (the real signal that this is a copy, not a homebrew with name collision).

- `src/view/dialogs/CharacterLevelUpDialogState.svelte.ts:233-243`, Same root issue surfaces here: `ownedSpellSignatures` is built from name fallback, so a homebrew spell the actor owns will block re-grant of a *different* Nimble spell that shares name+tier+school (and vice-versa). Whatever fix lands in `getSpells.ts` must apply here.

  ```ts
  if (sys?.school && sys.tier !== undefined) {
      ownedSpellSignatures.add(
          spellSignature(sys.identifier ?? '', item.name ?? '', sys.tier, sys.school),
      );
  }
  ```

## Suggestions

- `src/utils/getSpells.ts:160-164`, Sort comparator is fine but consider extracting to a named constant or adding a comment noting it's a partial sort (only the Nimble pack is positioned; others retain insertion order).
- `src/utils/getSpells.ts:36`, `spellSignature` name fallback uses lowercase but identifier path doesn't. If identifier is intended to be case-insensitive too (some GMs may type `Frost-Bolt` vs `frost-bolt`), normalize both. If not, document why.
- `src/utils/getSpells.ts:189`, Computing `sig` before the duplicate-UUID check in `addToIndex` means signature is calculated for spells that will be filtered by a different mechanism. Minor, but reorder to check `seen.has(uuid)` first if you care about wasted work on large indexes.
- `src/utils/getSpells.test.ts:200-228`, Test coverage doesn't include the homebrew-name-collision case. Add a test where a non-`nimble.nimble-spells` pack has a spell with same name+tier+school but different intent (different `uuid`, no `compendiumSource` linking back), current behavior will drop it, and that's the bug to surface.

PR #774, opened 2026-05-17
chore(build): fix svelte compile warnings

trevlar:
## Summary

- Remove dead `.nimble-code-block__text-area` CSS selectors in `ItemRulesTab` and `RuleCard` (left over from the old raw-JSON rules editor that was replaced by the schema-driven builder)
- Wrap `createItemActivationConfigDialogState` call in `untrack()` in `ItemActivationConfigDialog` — the dialog receives `actor`/`item`/`rollMode` as one-time inputs that do not change during its lifetime, so capturing initial prop values is intentional

## Test plan

- [x] `pnpm build` produces no warnings
- [x] All 1303 tests pass
- [x] TypeScript type check passes


Fronix (approved):
I missed that LGTM

PR #775, opened 2026-05-17
fix(stats): cap stat increases at +5 from level-up and stat config editor

### LINKED ISSUE #765: Max stat should be limited to +5 on stat increase
### Foundry VTT Version

13.*

### Nimble System Version

0.8.5

### Description

In the Nimble Core Rule book, it says that the maximum a hero stat can typically go is +5. 

As it stands, you can level up and bring it beyond five. You can also go and edit your character afterwards and make it higher than five through just like level up increases. We need to limit it for level ups so that it can only go to a maximum of five. 

### Steps to Reproduce

1. Open character sheet.
2. Edit the stats.
3. Increase strength on all of them and save. If you're at a high level, it's fairly easy to do. You could also do the level-up process, increasing strength four times.

### Expected Behavior

It should limit how much they can increase that stat to a maximum of 5. Both on level up and when going back and editing the values that were taken on level up 

### Actual Behavior

It's allowing it to go beyond five. You could probably do plus seven. 

### Additional Context

[Related question](https://discord.com/channels/1163603714565734400/1301262490679771166)

trevlar:
Waiting on response to https://discord.com/channels/1163603714565734400/1301262490679771166

## Summary

**Stat increase cap:**
- Disables stat increase options in the level-up dialog when an ability's total stat value (base + all prior increases) is already 5 or higher
- Disables toggle buttons in the character stat config editor under the same condition
- Adds a guard in `toggleStatIncreaseOption` to block bypassing the cap (client-side defense-in-depth)
- Adds a `statAtMax` localization string for the disabled tooltip

**Skills config dialog improvements (bundled on this branch):**
- Replaces the old increment/decrement point controls with a level-up history panel showing per-level skill allocations
- Adds an inline edit mode for redistributing skill points at a given level (budget-capped)
- Renames existing localization keys for shorter column headers (`"Ability Modifier"` → `"Abil. Mod"`, `"Skill Bonus"` → `"Bonus"`, `"Skill Points"` → `"Points"`)
- Sets dialog height to 600 to accommodate the history panel

**Other fixes:**
- Fixes cursor bleed-through from the combat tracker inside open windows (SCSS)
- Exposes `height` option on `GenericDialog` for dialogs that need a fixed height

## GitHub Issue

Closes https://github.com/Nimble-Co/FoundryVTT-Nimble/issues/765

## How it works

The cap fires when `baseValue + totalASI >= 5` — it considers both the ability's base score (from the stat array) and all committed stat increases, so a character with a high starting base reaches the cap sooner than one who started low. `NimbleClassItem` now exposes both `ASI` (level-filtered, for display totals) and `totalASI` (all assigned levels, for cap enforcement). The level-up dialog checks `characterClass.ASI[abilityKey]`; the stat config editor checks `totalASI[abilityKey]`. Deselecting an already-chosen stat is always allowed.

## Test Plan
- [ ] Level a character up to a high level and assign the same stat 5 times — the 6th increase should be blocked
- [ ] Open the stat config editor on a character with 5 increases in one stat — that stat's toggle buttons should appear disabled with a tooltip
- [ ] Verify deselecting an active toggle still works normally
- [ ] Verify stats with fewer than 5 increases remain selectable
- [ ] Open the skills config dialog — confirm the level-up history panel renders and the edit flow works
- [ ] Edit skill points at a level in the history panel and confirm the changes persist


public/lang/en.json
```diff
@@ -1682,12 +1306,19 @@
 		},
 		"skillsConfig": {
 			"skill": "Skill",
-			"abilityModifier": "Ability Modifier",
-			"skillBonus": "Skill Bonus",
-			"skillPoints": "Skill Points",
+			"abilityModifier": "Abil. Mod",
```
Fronix, line 1309: Not sure this shortening is understandable, could we get a more understandable shortening?

Fronix (approved):
Approved, just a small comment

public/lang/en.json
trevlar: Changed to "Stat Score".

PR #777, opened 2026-05-17
fix(actors): place new actors in the selected folder (#721)

### LINKED ISSUE #721: Creating an actor with a folder selected should place the new actor in that folder
### Problem or Motivation

When a folder is selected in the Actors directory, and the user clicks "Create Actor," the new actor is created at the root of the directory rather than inside the selected folder. This forces the user to manually drag the actor into the intended folder after creation, which is friction that compounds when setting up a campaign with many NPCs organized by folder (factions, locations, session prep, etc.).

The selected folder is a strong signal of intent — the user has navigated to that folder and is actively working in it. Ignoring that context and dumping the new actor at the root feels like a bug, even if it's technically working as designed.

### Proposed Solution

When the Actors directory has a folder selected or active when "Create Actor" is invoked, pre-populate the new actor's `folder` field with that folder's ID so it lands in the selected folder upon creation.

Ideally, the create dialog would also show the target folder so the user can confirm or override before creating.

### Alternatives Considered

- **Add a folder picker to the create dialog** — more explicit but adds a click for the common case.
- **Remember the last-used folder** — decouples from selection state but can be surprising across sessions.
- **Status quo + drag-to-move** — current behavior; rejected because it's the pain point being reported.

### Additional Context

This may be default behavior in FoundryVTT rather than something specific to the Nimble system — worth verifying before implementing a system-level workaround. If it reproduces in a vanilla world with no system-specific code paths involved, the right move is probably an upstream issue against core Foundry rather than patching it in the Nimble system.

Quick repro steps to confirm scope:
1. Open the Actors directory.
2. Create a folder, click to select it (without opening it).
3. Click "Create Actor," name it, and confirm.
4. Observe where the new actor lands.

If it lands at the root directory, it's the reported behavior. Repeat in a vanilla system (or with Nimble disabled) to determine whether this belongs here or upstream.

trevlar:
## Summary

- When the Actors sidebar has an active folder, `createDialog` now detects that folder (from either Foundry's `data.folder` context or `ui.actors.folder`) and pre-populates it on the new actor
- Non-character types (NPC, minion, solo monster) already propagated `this.data` to `Actor.create`, so they pick up the folder automatically once it is set in `createDialog`
- Character creation via `CharacterCreationDialog` was silently dropping the folder; the wizard now receives the folder ID and passes it to `Actor.create`

## How it works

1. `NimbleBaseActor.createDialog` resolves the target folder: `data.folder` (from Foundry's own folder-context button) → `ui.actors.folder.id` (active/browsed folder in the sidebar) → `null` (root)
2. The resolved folder ID is passed explicitly into `ActorCreationDialog`
3. Non-character paths spread `this.data` (unchanged behavior, now includes folder)
4. Character paths receive `folder` in the `CharacterCreationDialog` constructor and forward it to `Actor.create`

## Test plan

- [ ] Create a folder in the Actors directory, navigate into it (or select it), click "Create Actor" → confirm the actor appears inside the folder for all actor types
- [ ] Create an actor from root level (no folder selected) → confirm it lands at root
- [ ] Create an actor via the folder-specific create button in the tree → confirm it lands in that folder
- [ ] All of the above for character type (goes through CharacterCreationDialog wizard)

Closes #721

### DIFF
--- a/src/documents/actor/base.svelte.ts
+++ b/src/documents/actor/base.svelte.ts
@@ -149,12 +149,19 @@ class NimbleBaseActor<ActorType extends SystemActorTypes = SystemActorTypes> ext
 	): Promise<Actor.Stored | null | undefined> {
 		const { parent, pack, types } = context ?? {};
 
+		// Use folder from Foundry data if provided, otherwise fall back to the active folder in the sidebar
+		const folderId =
+			(data?.folder as string | null | undefined) ??
+			(ui.actors as unknown as { folder?: { id?: string } | null })?.folder?.id ??
+			null;
+
 		const { default: ActorCreationDialog } = await import(
 			'../dialogs/ActorCreationDialog.svelte.js'
 		);
 		const dialog = new ActorCreationDialog(
 			{
 				...data,
+				folder: folderId,
 				parent,
 				pack,
 				types,
--- a/src/documents/dialogs/ActorCreationDialog.svelte.ts
+++ b/src/documents/dialogs/ActorCreationDialog.svelte.ts
@@ -1,9 +1,8 @@
 import type { DeepPartial } from 'fvtt-types/utils';
 import { SvelteApplicationMixin } from '#lib/SvelteApplicationMixin.svelte.js';
-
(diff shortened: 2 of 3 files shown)


Fronix (approved):
LGTM

PR #778, opened 2026-05-17
fix(charCreation): Celestial saving throw bug, source labels on option cards, responsive ancestry grid

trevlar:
### Root cause

The Celestial ancestry has a `savingThrowRollMode` rule (type `set`, target `disadvantaged`, value `0`) that neutralises the class's disadvantaged saving throw. If a GM imported Celestial before that rule was added to the pack, the world-level copy of the item lacked the rule. During character creation, `resolveSavingThrowRollModes` only reads rules from the selected document — so a stale world copy meant the disadvantaged save was left at `defaultRollMode: -1` on the new character sheet.

### Fixes

**Migration 019** (`Migration019CelestialSavingThrow`) — runs on existing character actors:
- Finds characters with an embedded Celestial ancestry (matched by `compendiumSource`, legacy `sourceId`, or name)
- Adds the missing `savingThrowRollMode` rule to the embedded ancestry item if absent
- Corrects `defaultRollMode` for the class's disadvantaged save from `-1` → `0`

**Character creation saving throw resolution** — `resolveSavingThrowRollModes` now uses rules from the selected document as-is, including world items with custom rule modifications. No silent fallback to the compendium source.

**Character creation option lists** — world items and compendium items are both surfaced so GMs and players can choose from either. A source label on each card identifies where it came from.

### Source labels on character creation cards

- New `getDocumentSourceLabel(uuid)` utility returns the compendium pack's `metadata.label` (e.g. "Nimble Core") for compendium items, or `"World"` for world items
- Source label appears in the card meta row (below the name), right-aligned — **no absolute positioning, no overlap with long names**
- Ancestry cards: name (top row) + size left / source badge right (bottom row)
- Background cards: same two-row layout with source badge
- Class cards: source badge inline in the complexity row, pushed to the right

### Ancestry card layout polish

- Card image stretches flush to top and bottom edges (`align-self: stretch; height: auto`)
- Title and meta rows have vertical breathing room (padding above name, below source)
- Title and size text are left-aligned (`--nimble-card-title-justification: start`)
- Meta row is `flex-wrap: wrap` — at narrow widths the source badge wraps to the next line rather than overflowing
- Source badge uses `margin-inline-start: auto` to stay right-anchored on the same line when space allows

### Responsive ancestry grid

Switched from a fixed `repeat(3, 1fr)` to `repeat(auto-fill, minmax(260px, 1fr))`:
- ~2 columns at the default ~610px dialog width
- ~3 columns at ~800px
- ~4 columns at ~1060px

No JavaScript, no resize observer — pure CSS.

## Closes

https://github.com/Nimble-Co/FoundryVTT-Nimble/issues/719

## Test Plan

- [x] `pnpm check` passes (lint, format, types, 1094 tests)
- [ ] Manually create a Celestial character — verify the class's disadvantaged save is neutralised on the new PC sheet
- [ ] Run Migration 019 on an existing affected character — verify `defaultRollMode` is corrected and the `savingThrowRollMode` rule is present on the embedded ancestry
- [ ] Verify "Reset to Class Defaults" in `ActorSavingThrowConfigDialog` correctly neutralises the disadvantaged save for Celestial characters
- [ ] If both a world Celestial and a compendium Celestial are present, both appear in the ancestry list with their respective source labels
- [ ] Resize the character creation dialog and confirm ancestry columns step from 2 → 3 → 4 as width increases
- [ ] Resize to a narrow width and confirm the source badge wraps below the size text rather than overflowing the card


src/utils/getChoicesFromCompendium.ts
```diff
@@ -1,20 +1,26 @@
...
+	// rules rather than a potentially stale local copy.
 	for (const item of game.items) {
 		if (item.type !== documentType) continue;
-		documentIDs.push(item.uuid);
+		const itemData = item as object as {
+			_stats?: { compendiumSource?: string };
+			flags?: { core?: { sourceId?: string } };
+		};
+		const compendiumSource = itemData._stats?.compendiumSource ?? itemData.flags?.core?.sourceId;
+		documentIDs.add(compendiumSource ?? item.uuid);
```
trevlar: This silently drops world items without giving players/GMs options.

Fronix (approved):
LGTM

PR #780, opened 2026-05-17
feat(combat-ui): highlight canvas token on hover (#487)

### LINKED ISSUE #487: Highlight hovered icon on canvas when hovering over actor in the combat system
It's legitimately impossible for me to tell which tokens are which in the combat tracker.

trevlar:
## Summary

- Adds a PIXI ring highlight on canvas tokens when hovering over combatant entries across the Combat Tracker (CT), the Nimble Combat System Window (NCSW), and the Targets panel
- All hover rows in the NCSW (roll row, roll-and-end-turn row, member cells) highlight the corresponding canvas token
- Monster stack cards in the CT highlight **all** tokens in the group simultaneously
- Adds a configurable **Hover Color** swatch to CT Settings alongside the existing Action/Reaction colors
- Hover color is applied at hover-time so changes take effect immediately without reload
- Works for both controlled and uncontrolled tokens (uses PIXI ring, not Foundry native highlight)

## Implementation

- New shared utility `src/utils/tokenHoverHighlight.ts` exports `tokenHoverIn`, `tokenGroupHoverIn`, and `tokenHoverOut`
- `tokenGroupHoverIn` draws rings on multiple tokens simultaneously (used for monster stacks)
- Hover color read from `combatTrackerSettings` at call-time via `getCombatTrackerHoverColor()`
- `combatTrackerSettings.ts` extended with `combatTrackerCtHoverColor` setting (client-scoped)
- CT Settings dialog updated with hover color row (presets + color picker)

## Closes

Closes #487

## Test Plan

- [x] Hover over a combatant card in the CT — ring appears on canvas token
- [x] Hover over a monster stack card in the CT — rings appear on all tokens in the group
- [x] Hover over rows in the NCSW (roll row, roll-and-end-turn row) — ring appears
- [x] Hover over target entries in the Targets panel — ring appears
- [x] Change Hover Color in CT Settings — new color reflected on next hover
- [x] Uncontrolled tokens highlight correctly (no token selected)
- [x] Ring disappears on mouse leave

### DIFF
--- a/docs/STYLE_GUIDE.md
+++ b/docs/STYLE_GUIDE.md
@@ -1470,6 +1470,7 @@ export const myStore = writable<MyStoreState>(initialState);
 | `spell/*` | `src/utils/spell/` | Spell-related utilities |
 | `treeManipulation/*` | `src/utils/treeManipulation/` | Tree data structure utilities |
 | `countAdjacentEnemies()`, `ADJACENCY_QUALIFIER` | `src/utils/tokenAdjacency.ts` | Count enemy tokens adjacent to a given token; respects the `adjacencyIncludesDiagonals` world setting; accepts position overrides for pre-commit token moves |
+| `tokenHoverIn()`, `tokenGroupHoverIn()`, `tokenHoverOut()` | `src/utils/tokenHoverHighlight.ts` | Draw/remove a PIXI ring on canvas tokens on hover; `tokenGroupHoverIn` highlights multiple tokens simultaneously (used for monster stacks) |
 
 ### Shared Stores
 
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -2038,6 +2038,10 @@
 				"name": "Debug Mode",
 				"hint": "Enables developer tools like the dice testbench. Changes take effect immediately.",
 				"openTestbench": "Open Dice Testbench"
+			},
+			"combatTrackerHoverColor": {
+				"name": "Combat Tracker Hover Color",
+				"hint": "Choose your personal token hover highlight color"
 			}
 		},
 		"diceTestbench": {
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -1144,6 +1144,7 @@
 				"colorsSection": "Colors",
 				"actionColor": "Action Color",
 				"reactionColor": "Reaction Color",
+				"hoverColor": "Hover Color",
 				"colorWhite": "White",
 				"colorGreen": "Green",
 				"colorRed": "Red",
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -2042,6 +2042,91 @@
 			"combatTrackerHoverColor": {
 				"name": "Combat Tracker Hover Color",
 				"hint": "Choose your personal token hover highlight color"
+			},
+			"combatTrackerPlayersCanExpandMonsterCards": {
+				"name": "Combat Tracker Monster Card Visibility",
+				"hint": "Allow players to view individual monster and minion cards when the GM expands them"
+			},
+			"combatTrackerResourceDrawerHover": {
+				"name": "Combat Tracker Resource Drawer Hover",
+				"hint": "When enabled, player resource drawers open only on hover"
+			},
+			"combatTrackerPlayerHpBarTextMode": {
+				"name": "Combat Tracker Player HP Bar Center Text",
+				"hint": "Choose whether player HP bars show nothing, health state, or HP percentage"
+			},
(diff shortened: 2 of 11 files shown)


Fronix (approved):
Looks good and works as advertised, not a fan of the rings but that's a small design choice we can change later if so.

## Suggestions, nothing blocking:

- `src/utils/tokenHoverHighlight.ts:16` — module-level mutable `ringTokens` array creates hidden cross-call state and is not cleared on scene change, canvas teardown, or token deletion. If a token is destroyed while a ring is on it, the next `removeAllRings()` iterates a stale reference.
  ```ts
  let ringTokens: unknown[] = [];
  ```
  Fix: guard `removeChild` against destroyed PIXI objects (check `t.destroyed` / `ring.destroyed`) and consider clearing on `canvasTearDown`/`deleteToken` hooks.

- `src/utils/tokenHoverHighlight.ts:39` — `PIXI` is used as a global but not declared/imported; relies on Foundry runtime globals. Same pattern as `canvasInternal.ts` so consistent, but ensure it's in the Biome/ESLint globals list (it is via Foundry config) — flagging for awareness.

- `src/utils/tokenHoverHighlight.ts:6-14` — `try/catch` wraps `getCombatTrackerHoverColor()`/`parseInt`, neither of which throws. Dead defensive code.
  ```ts
  function getHoverRingColor(): number {
  	try {
  		const hex = getCombatTrackerHoverColor().replace('#', '');
  		const parsed = parseInt(hex, 16);
  		return Number.isFinite(parsed) ? parsed : DEFAULT_HOVER_RING_COLOR;
  	} catch {
  		return DEFAULT_HOVER_RING_COLOR;
  	}
  }
  ```
  Fix: drop the try/catch — `normalizeHexColor` already guarantees a `#RRGGBB` string.

- `src/utils/tokenHoverHighlight.ts:49,56,65` — `_event` parameter is unused in every export. YAGNI — drop it from the signature rather than carry a permanently-unused arg.

- `src/settings/combatTrackerSettings.ts:75` — `CT_HOVER_COLOR_CSS_VAR` and `applyCtHoverColorCssVariable` are defined and wired through `onChange` + module init, but `--nimble-ct-hover-color` is never read in any CSS/SCSS. The ring is drawn via PIXI using `getHoverRingColor()` directly. Dead variable + dead apply call.
  ```ts
  const CT_HOVER_COLOR_CSS_VAR = '--nimble-ct-hover-color';
  ```
  Fix: remove `CT_HOVER_COLOR_CSS_VAR`, `applyCtHoverColorCssVariable`, the `onChange` call to it, and the init-time `applyCtHoverColorCssVariable(getCombatTrackerHoverColor())`.

### Minor things
- `src/settings/combatTrackerSettings.ts:572-577` — `setCombatTrackerHoverColor` calls `applyCtHoverColorCssVariable` which is dead (see above). Remove that line once the CSS var is dropped.

- `src/settings/combatTrackerSettings.ts:569-574` — `game.settings.get('nimble' as 'core', COMBAT_TRACKER_HOVER_COLOR_SETTING_KEY as 'rollMode')` — the `as 'core'`/`as 'rollMode'` casts are a code smell, but they match the existing pattern in this file so acceptable for consistency.

- `src/hooks/minionGroupTokenActions.ts:1585,1590` — using `addEventListener('mouseenter', …)` then never removing them. Existing code in this section uses the same pattern (the popover listeners), so consistent, but note listeners accumulate if the panel re-renders without DOM cleanup.

- `src/hooks/minionGroupTokenActions.ts:1732-1738,1835-1841` — `getNonMinionToken`/`getMinionToken` re-resolves the combatant on every mouse event. Acceptable since hover is infrequent, but could be hoisted/cached if combat state is stable.

- `src/view/ui/CtTopTracker.state.svelte.ts:538` — `tokens` typed as inferred from `.filter((t) => t !== null)` won't actually narrow `unknown[]` — `getCombatantToken` likely returns nullable; verify the type narrowing flows through to `tokenGroupHoverIn`.

- `src/view/chat/components/Targets.svelte:1` — file changed to `lang="ts"` but most function params remain implicitly `any` (`token`, `targetId`, `targetIDs`). Out of scope for this PR but flagging — adding `lang="ts"` without typing is mostly cosmetic.

- `src/view/ui/ctTopTracker/types.ts:48-49` — adding `isVisible?` and `controlled?` to a shared types file: confirm this is the right place vs. a token-shape type local to the hover utility.

- `public/lang/en.json:1144,2038-2042` — English-only addition (`hoverColor`, `combatTrackerHoverColor.name/.hint`). Per AGENTS.md, flag for human review of source strings.

PR #782, opened 2026-05-17
feat(predicate): expand predicate vocabulary with self/target tags and targetCondition

### LINKED ISSUE #579: [Foundation] Predicate vocabulary expansion

tristin-albers:
## Summary

Expands the predicate vocabulary so rules can gate on actor state (`self:bloodied`, `self:fullHp`, etc.) and target state (`target:bloodied`, `target:concentrating`). Adds a `targetCondition` field on `damageBonus` for target-gated damage bonuses evaluated at activation time.

## Changes

- **Self-state tags**: Populate `self:bloodied`, `self:dying`, `self:fullHp`, `self:concentrating` on all actors; `self:shield`/`self:noShield` on characters
- **Target-state tags**: Populate `target:bloodied`, `target:concentrating` on all actors for `targetCondition` evaluation
- **`targetCondition` on damageBonus**: New `PredicateField` evaluated against the target's domain at activation time. Plumbed through `DamageBonusEntry` → `matchesBonus()` → `getDamageBonusTotal()`/`getDamageBonusFormulas()` → `ItemActivationManager`
- **Predicate engine**: Add presence-check statement (`{ "self:bloodied": true }`) — see design decision below
- **Rules builder integration**: `targetCondition` field has i18n label/hint for the builder UI

## Design Decision — AND Composition (for @Fronix review)

The issue flagged AND-composition as an open question: "allow multiple `self:*` keys in a predicate (small engine change), or push compound tags like `self:raging+bloodied` (no engine change, ugly)."

This PR implements the first option via a **presence-check** syntax. When a predicate value is boolean `true`, the engine checks if the full key exists directly in the domain:

```jsonc
// AND: both must be present
{ "self:bloodied": true, "self:raging": true }

// OR already works via arrays (unchanged)
{ "self": ["bloodied", "raging"] }
```

This works because `"self:bloodied"` and `"self:raging"` are distinct Map keys — no duplicate key problem. The change is 4 lines in `Predicate.ts`. **Happy to revert this if you'd prefer a different approach.**

## Test Plan

- [x] `pnpm check` passes (format, lint, circular deps, type-check, tests — 1316 passing)
- [x] Toggle bloodied status on character → `self:bloodied` + `target:bloodied` appear in domain
- [x] Toggle concentration → `self:concentrating` + `target:concentrating` appear
- [x] Toggle lastStand → `self:dying` appears
- [x] Character at full HP → `self:fullHp` appears; reduce HP → tag drops off
- [x] NPC at full HP → `self:fullHp` appears (stored hp.max, no timing issue)
- [x] Character with no shield → `self:noShield` appears
- [x] Predicate AND composition: `{ "self:bloodied": true, "self:concentrating": true }` requires both
- [x] `targetCondition` bonus excluded when no target selected
- [x] `targetCondition` bonus included when target domain matches

## Checklist

## Related Issues

Closes #579


src/documents/actor/base.svelte.ts
```diff
@@ -277,6 +278,29 @@ class NimbleBaseActor<ActorType extends SystemActorTypes = SystemActorTypes> ext
...
+				this.tags.add('target:concentrating');
+			}
+		}
+
+		// self:fullHp — HP at maximum
+		const systemData = this.system as unknown as BaseActorSystemData;
+		const { value: hpValue, max: hpMax } = systemData.attributes.hp;
+		if (hpMax > 0 && hpValue >= hpMax) {
+			this.tags.add('self:fullHp');
+		}
```
Fronix: **Blocking.** `self:fullHp` is added here using the stored `hp.max`, then re-added in `character.ts:223-230` after the derived max is computed. Neither pass calls `delete()`, so a stale tag from the first pass survives if stored max and derived max disagree (for example, stored max=10 value=10 adds the tag, then derived max=15 leaves it stuck).

Fix: compute `self:fullHp` only once, after `hp.max` is finalized, or call `this.tags.delete('self:fullHp')` before the second-pass re-check on character.

src/documents/actor/base.svelte.ts
```diff
@@ -277,6 +278,29 @@ class NimbleBaseActor<ActorType extends SystemActorTypes = SystemActorTypes> ext
...
+				this.tags.add('self:bloodied');
+				this.tags.add('target:bloodied');
+			}
+			if (statuses.has(STATUS_EFFECT_IDS.lastStand)) {
+				this.tags.add('self:dying');
+			}
+			if (statuses.has(STATUS_EFFECT_IDS.concentration)) {
+				this.tags.add('self:concentrating');
+				this.tags.add('target:concentrating');
+			}
```
Fronix, line 317: **Blocking (namespace contract violation against #579).** Every actor publishes both `self:bloodied`/`target:bloodied` and `self:concentrating`/`target:concentrating` onto its own `tags`. Since `getDomain()` returns `this.tags` wholesale, a `targetCondition: { "self:bloodied": true }` matches when the *target* is bloodied. Issue #579 explicitly partitions `self:*` and `target:*` as separate namespaces.

Fix options: (a) filter `getDomain()` to strip `self:*` when used as a target domain, (b) expose a separate `getTargetDomain()` that returns only `target:*` tags, or (c) split tag storage so target tags do not leak into self-predicate evaluation. Add a test covering: an attacker rule with `targetCondition: { 'self:bloodied': true }` should not fire when the target (not the attacker) is bloodied.

src/documents/actor/base.svelte.ts
```diff
@@ -277,6 +278,29 @@ class NimbleBaseActor<ActorType extends SystemActorTypes = SystemActorTypes> ext
...
+		const statuses = this.statuses as Set<string> | undefined;
+		if (statuses) {
+			if (statuses.has(STATUS_EFFECT_IDS.bloodied)) {
+				this.tags.add('self:bloodied');
+				this.tags.add('target:bloodied');
+			}
+			if (statuses.has(STATUS_EFFECT_IDS.lastStand)) {
+				this.tags.add('self:dying');
+			}
+			if (statuses.has(STATUS_EFFECT_IDS.concentration)) {
```
Fronix, line 314: Issue #579 says "Concentration flag set" as the trigger. This implementation reads `statuses.has(STATUS_EFFECT_IDS.concentration)`. Please verify the codebase's concentration toggle round-trips through the status registry rather than a private flag, otherwise the tag goes stale silently when concentration is set/cleared by a code path that does not touch statuses.

src/documents/actor/character.ts
```diff
@@ -220,6 +220,15 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
...
 		this.prepareClassData(actorData);
 
+		// Re-evaluate self:fullHp now that hp.max is computed (base _populateDerivedTags
+		// runs before _prepareHitPoints, and character hp.max is derived, not stored)
+		if (
+			actorData.attributes.hp.max > 0 &&
+			actorData.attributes.hp.value >= actorData.attributes.hp.max
+		) {
+			this.tags.add('self:fullHp');
+		}
```
Fronix, line 230: Part of the `self:fullHp` two-pass issue flagged on `base.svelte.ts:302`. This re-add never deletes a stale tag from the base pass. Either move the entire `self:fullHp` computation here (and drop it from base) for actor types whose max is derived, or explicitly `this.tags.delete('self:fullHp')` before the `add()` so the second-pass `value >= max` check is authoritative.

src/documents/actor/character.ts
```diff
@@ -276,6 +285,14 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
...
 		});
 		this.tags.add(`armor:${hasArmor ? 'equipped' : 'unarmored'}`);
+
+		// Shield status tag
+		const hasShield = this.items.some((item) => {
+			if (!item.isType('object')) return false;
+			const objectItem = item as unknown as NimbleObjectItem;
+			return objectItem.system.objectType === 'shield';
+		});
+		this.tags.add(hasShield ? 'self:shield' : 'self:noShield');
```
Fronix, line 296: `self:shield` / `self:noShield` only exist on characters. A homebrew rule using `{ 'self:noShield': true }` in a predicate will silently never match for monsters. Per #579's worked examples this is intentional, but a one-line code comment or an i18n hint on the `predicate` field would save a future content author an hour of debugging.

src/models/rules/damageBonus.ts
```diff
@@ -62,6 +66,10 @@ function schema() {
 			hint: 'NIMBLE.rules.damageBonus.source.hint',
 			choices: ['weapon', 'spell', 'any'],
 		}),
+		targetCondition: new PredicateField({
+			label: 'NIMBLE.rules.damageBonus.targetCondition.label',
+			hint: 'NIMBLE.rules.damageBonus.targetCondition.hint',
+		} as unknown as never),
```
Fronix, line 75: The `as unknown as never` cast on the options object is a known workaround for a typing gap in `PredicateField`'s constructor signature (tracked separately as a rules-builder gap). Functionally fine: `RuleCard` reads `label`/`hint` off the field instance, so they render around the predicate builder as expected. No change needed here, just flagging that this is the first call site to set `label`/`hint` on a `PredicateField` and the cast is load-bearing until the field's constructor typing is fixed.

src/models/rules/damageBonus.ts
```diff
@@ -104,6 +112,10 @@ class DamageBonusRule extends NimbleBaseRule<DamageBonusRule.Schema> {
 	declare delivery: DamageBonusDelivery;
 	declare source: DamageBonusSource;
 
+	private get _targetCondition(): Predicate {
+		return (this as object as { targetCondition: Predicate }).targetCondition;
+	}
```
Fronix, line 120: **Blocking.** `_targetCondition` getter casts and immediately reads `.size` in `afterPrepareData` without guarding against `undefined`. If a legacy or partially-migrated rule loads with `targetCondition` unset, `.size > 0` throws.

Fix:
```ts
private get _targetCondition(): Predicate | undefined {
    return (this as object as { targetCondition?: Predicate }).targetCondition;
}
```
And in `afterPrepareData`:
```ts
const tc = this._targetCondition;
const targetConditionRaw = tc && tc.size > 0 ? tc.toObject() : null;
```
Add a regression test for a `damageBonus` rule whose source object has no `targetCondition` key at all.

src/models/rules/damageBonus.test.ts
```diff
@@ -85,6 +85,11 @@ function createDamageBonusRule(
 		configurable: true,
 	});
 
+	Object.defineProperty(rule, '_targetCondition', {
+		get: () => ({ size: 0 }),
```
Fronix, line 90: **Blocking (coverage gap).** Every test stubs `_targetCondition` with `{ size: 0 }`, so the schema-to-bonus-entry serialization path (`.size > 0 ? toObject() : null` in `damageBonus.ts:168-169`) is never exercised. Issue #579 called for "a mocked target carrying domain tags", which is present for the `matchesBonus` filter tests, but the rule-side integration is uncovered.

Fix: add at least one test that lets a real `PredicateField`-initialized `Predicate` flow through `afterPrepareData`, verifying `targetCondition` lands on the resulting bonus entry as a `RawPredicate` object. Without this, regressions in the schema integration will pass green.

src/etc/Predicate.ts
```diff
@@ -156,6 +157,7 @@ class Predicate extends Map<string, Statement> {
 	/** Validators                                     */
 	/** ---------------------------------------------- */
 	static isStatement(statement: unknown): statement is PredicateStatement {
+		if (statement === true) return true;
```
Fronix: `isStatement(true)` returns `true`, but `isStatement(false)` returns `false`. The asymmetry is intentional (the issue's open question chose presence-only), but undocumented in code. Add a one-line comment so a future contributor does not "fix" the false-rejection and accidentally introduce a `not-present` semantics.

src/utils/attackUtils.ts
```diff
@@ -80,17 +81,23 @@ export function hasWeaponProficiency(
...
 	if (bonus.delivery !== 'any' && bonus.delivery !== delivery) return false;
 	if (bonus.source !== 'any' && bonus.source !== source) return false;
 	if (bonus.damageType !== '' && damageType && bonus.damageType !== damageType) return false;
+	if (bonus.targetCondition) {
+		if (!targetDomain) return false;
+		const predicate = new Predicate(bonus.targetCondition);
+		if (!predicate.test(targetDomain)) return false;
+	}
 	return true;
 }
```
Fronix, line 105: `new Predicate(bonus.targetCondition)` is constructed inside the `matchesBonus` filter loop, so every attack rebuilds the Predicate N times for N bonuses. The issue mandates that evaluation happens at attack time (not data prep), so the *test* call stays here, but the Predicate *instance* can be cached on the `DamageBonusEntry` once after `afterPrepareData`. Microperf, not architecture, but cheap to fix.

src/managers/ItemActivationManager.ts
```diff
@@ -660,6 +671,19 @@ class ItemActivationManager {
...
+	 * Returns undefined if no targets are selected or the target has no domain.
+	 */
+	#getFirstTargetDomain(): Set<string> | undefined {
+		const targets = game.user?.targets;
+		if (!targets || targets.size === 0) return undefined;
+
+		const firstTarget = targets.values().next().value as Token | undefined;
+		const targetActor = firstTarget?.actor as { getDomain?: () => Set<string> } | null | undefined;
+		return targetActor?.getDomain?.();
+	}
```
Fronix, line 693: `#getFirstTargetDomain` arbitrarily picks the first `Set` element. `game.user.targets` iteration order is insertion order, so the chosen target is whatever Foundry happened to insert first. Worked examples in issue #579 are single-target (Hunter's Mark, Hexbinder vs Afflicted), so this is out of scope, but a one-line code comment noting that multi-target / AOE bonuses gated by `targetCondition` are not supported would prevent future surprise.

Fronix (changes requested):
### Documentation gap

The new predicate vocabulary (`self:*`, `target:*` tags) and the new presence-check AND-composition syntax (`{ 'self:bloodied': true }`) need to be documented in `docs/system/rules.md`. Three composition syntaxes now coexist (atom string, array-OR, presence-check-AND) and a content author has no canonical reference for which tags exist or which form to use.

### Note: rules-builder typing gap (not blocking)

This PR is the first place a `PredicateField` is constructed with `label` / `hint` options, and it required an `as unknown as never` cast to satisfy the constructor's typed signature. The renderer reads those properties off the field instance and they show up correctly, so the cast is load-bearing but not buggy. Worth a follow-up to fix `PredicateField`'s constructor typing so future call sites do not need the cast.

### Out-of-diff finding

`src/documents/actor/base.svelte.ts:535-538`: `getDomain()` returns the live `this.tags` Set instead of a clone. Predicate evaluation operates on a live reference, so concurrent mutation during evaluation is undefined behavior. Either return `new Set(this.tags)` defensively or document the contract that callers must not mutate.

tristin-albers:
Addressed review feedback

Blocking fixes:                                                                                                      
  - self:fullHp stale tag: removed from base _populateDerivedTags(), each actor subclass now computes it after HP is   
  finalized                                                                                                            
  - Namespace leak: added getTargetDomain() returning only target:* tags, used by ItemActivationManager                
  - _targetCondition undefined crash: getter returns Predicate | undefined, guarded with &&                            
  - Test coverage gap: added tests for legacy rule without targetCondition, serialization path, namespace isolation,   
  and empty {} edge case  
                                                        
 Additional fixes from further review                                                                              
  - self:dying was incorrectly mapped to lastStand, fixed to use dying status (PC/Hero condition). 
  - Added self:lastStand as a separate tag for Solo/Legendary monsters.                                                        
  - Bloodied detection: added HP threshold failsafe alongside status check (just in case syncing issues, but should we keep it to just a status check?)                                              
  - "targetCondition: {}" edge case: empty object now treated same as null (no target required)                          
  - Documented unarmed strike targetDomain limitation in AttackActionPanel and activateHeroicActionMacro               
  
  Open to discussion                                                                                                
  - AND-composition syntax ({ "self:bloodied": true }), not sure if it's the best approach to AND's but even though it's lengthy it does keep it consistent to follow easily.
  - Added full predicate vocabulary reference to docs/system/rules.md, let me know if this is the right place for this please?

Fronix (changes requested):
## Scope check

All in-scope tags are in (`self:bloodied|dying|fullHp|concentrating|shield|noShield`, `target:bloodied|concentrating`), and everything the issue defers (`self:raging`, `target:marked`, etc.) is correctly left to the feature issues that own application. AND-composition lands as presence-check syntax. `self:lastStand` is added beyond the listed vocabulary, which seems fine.

## Blocking

- `src/documents/actor/base.svelte.ts:291-301`, `self:dying` and `self:bloodied` can both fire on a Dying actor if the bloodied status is also active. Issue #579 calls these mutually exclusive, and the rulebook treats Bloodied (≤50% HP) and Dying (0 HP) as distinct states.
  ```ts
  if (statuses.has(STATUS_EFFECT_IDS.bloodied) || (hpMax > 0 && hpVal > 0 && hpVal <= hpMax / 2)) {
      this.tags.add('self:bloodied');
      this.tags.add('target:bloodied');
  }
  if (statuses.has(STATUS_EFFECT_IDS.dying)) {
      this.tags.add('self:dying');
  }
  ```
  Fix: if `dying` is set, skip the bloodied branch.

## Suggestions

- `src/documents/actor/base.svelte.ts:547-550`, `getDomain()` hands back the live `this.tags` Set. Either clone it or add a JSDoc saying callers must not mutate.
- `src/models/rules/damageBonus.ts:69-75`, `as unknown as never` cast on `PredicateField`. The inline TODO is right; worth a follow-up issue against `PredicateField`'s constructor typing so this doesn't spread to other rule types.
- `src/utils/attackUtils.ts:96-101`, `new Predicate()` rebuilt every call. Fine at today's bonus counts, only worth caching if profiling complains.
- `public/lang/en.json:745-748`, `targetCondition.hint` only mentions `target:bloodied`. Add `target:concentrating` too, or link the docs anchor.
- The headline #579 example is Hunter's Mark with `target:marked`, which is deferred. Worth dropping one `target:bloodied` example into `docs/system/rules.md` so content authors have something concrete to copy.

## Resolved (prior review)

`self:fullHp` double-write, self/target namespace leak, unguarded `_targetCondition.size`, schema↔serialization test gap, `getFirstTargetDomain` ordering, presence-check asymmetry.

---

## The AND-composition problem

My recommendation is that we go the extra mile here and add support for `$or` / `$and`

The presence-check (`{ "self:bloodied": true }`) ships okayISH, but `true` is a magic sentinel that carries no information, and there are now two ways to express "self is bloodied" (`{ self: "bloodied" }` vs `{ "self:bloodied": true }`). That asymmetry is a smell, but the real problem shows up the moment a rule wants OR-of-ANDs, e.g. a Berserker "while raging, gain X when bloodied or dying":

```jsonc
// presence-check (current): can't OR two clauses
{ "self:raging": true, "self:bloodied": true }   // raging AND bloodied only

// any in-namespace `and` operator hits the same wall
{ self: { and: ["raging", "bloodied"] } }
// can't add `self: ["bloodied", "dying"]` alongside, duplicate `self` key
```

My idea here is that we add add top-level `$and` / `$or` keys, like MogoDB has. Sub-predicates compose recursively, leaves stay exactly as they are today (atomic, array-OR, presence-check), and the existing object form is just an implicit `$and`.

The example Berserker case becomes:
```jsonc
{
  "self:raging": true,
  $or: [
    { "self:bloodied": true },
    { "self:dying": true }
  ]
}
```

Or fully nested if a feature ever needs it:

```jsonc
{
  $or: [
    { $and: [{ "self:raging": true }, { "self:bloodied": true }] },
    { $and: [{ "self:raging": true }, { "self:dying": true }] }
  ]
}
```

I think this over an in-namespace `and` operator: an `and` under one key only handles single-namespace ANDs (`self:bloodied` AND `self:raging`). It can't cross namespaces (`self:raging` AND `target:bloodied`), and it can't express OR-of-ANDs at all. Top-level `$or`/`$and` covers every case, including the ones we haven't thought of yet. I haven't truly gone through and checked if we will have a lot of these but this protects us from having to update this later and I am sure that we can fine these complicated and/or's in the current rules, and we open the doors for hombrewer's to do cool stuff.

I think it's worth doing this in this PR so we can start cleaning up the old stuff right away and avoid a heavy migration later.

Fronix:
@tristin-albers thanks, i'll take a look this weekend

Fronix (approved):
LGTM

PR #786, opened 2026-05-22
fix: apply negative max hit dice bonuses

### LINKED ISSUE #741: Negative maxHitDice rule reduces displayed total but not rollable pool
### Foundry VTT Version

v13

### Nimble System Version

0.8.3

### Description

A `maxHitDice` rule with a negative `value` reduces the displayed hit dice total on the character sheet, but the actor can still roll the original (un-reduced) number of hit dice from the hit dice manager. Positive values work correctly; only the decrement path is broken.

### Steps to Reproduce

1. Create a character with a class that grants hit dice (e.g. a level-3 character with 3 d8 hit dice).
2. Create a feature/item with a `maxHitDice` rule: `{ "type": "maxHitDice", "value": "-1", "dieSize": 0 }`.
3. Add the feature to the character.
4. Observe the hit dice total on the sheet — it shows one less.
5. Open the hit dice manager and roll hit dice.

### Expected Behavior

The rollable pool reflects the reduced total. With 3 base hit dice and a `-1` rule, only 2 hit dice should be rollable.

### Actual Behavior

The displayed total decreases by 1, but all 3 original hit dice remain rollable. The displayed count and the rollable count are desynchronized.

### Additional Context

Root cause appears to be asymmetric handling of the `bonus` field — negative values are gated out:

- `src/managers/HitDiceManager.ts:146` — `if (bonus > 0)` skips negative bonuses when computing `bySize.total`.
- `src/applications/sheets/PlayerCharacterSheet.state.svelte.ts:240` — same `> 0` guard on the sheet display path.
- `src/models/rules/maxHitDice.ts:66-67` writes the negative bonus correctly; the downstream consumers drop it.

Likely fix: remove the `> 0` guards, clamp the resulting `total` to `>= 0`, and clamp `current` to the new `max` so spent + removed dice don't leave a stale rollable count.

Positive `maxHitDice` rules confirmed working in the same setup — only the negative path misbehaves.

puneetdixit200:
## Summary
- Apply rule-based maxHitDice bonuses when they are negative, not only when positive.
- Clamp derived hit dice totals/current values so negative bonuses reduce the rollable pool without creating negative dice.
- Clamp configured hit dice current values when a bonus decrease lowers the maximum below the stored current value.

Fixes #741.

## Tests
- `pnpm exec vitest run src/managers/HitDiceManager.test.ts`
- `pnpm exec vitest run src/managers/HitDiceManager.test.ts src/models/rules/maxHitDice.test.ts`
- `pnpm exec vitest run src/hooks/combatantHooks/combatantDefeatSync.test.ts`
- `pnpm exec biome check src/managers/HitDiceManager.ts src/managers/HitDiceManager.test.ts src/documents/actor/character.ts src/view/sheets/PlayerCharacterSheet.state.svelte.ts`
- `pnpm exec eslint src/view/dialogs/FieldRestDialog.svelte src/view/dialogs/SafeRestDialog.svelte`
- `pnpm lint:svelte`
- `pnpm type-check`
- `pnpm exec biome lint ./src/ ./types/ ./lib/`

## Notes
- `pnpm test` was also attempted; one full run hit a Vitest fork startup timeout in `combatantDefeatSync.test.ts`, and that same file passed when rerun alone.
- This PR was AI-assisted; I reviewed and verified the resulting changes.

### DIFF
--- a/src/managers/HitDiceManager.ts
+++ b/src/managers/HitDiceManager.ts
@@ -9,6 +9,17 @@ function incrementDieSize(baseSize: number, steps: number): number {
 	return DIE_SIZES[newIndex];
 }
 
+function clampHitDiceBySize(
+	hitDiceBySize: Record<string, { current: number; total: number }>,
+): Record<string, { current: number; total: number }> {
+	for (const data of Object.values(hitDiceBySize)) {
+		data.total = Math.max(data.total, 0);
+		data.current = Math.min(Math.max(data.current, 0), data.total);
+	}
+
+	return hitDiceBySize;
+}
+
 class HitDiceManager {
 	#actor: NimbleCharacterInterface;
 
@@ -66,7 +77,7 @@ class HitDiceManager {
 			const baseSize = Number(sizeStr);
 			const size = incrementDieSize(baseSize, hitDiceSizeBonus);
 			const bonus = (hitDieData as { bonus?: number }).bonus ?? 0;
-			if (bonus > 0) {
+			if (bonus !== 0) {
 				this.#max += bonus;
 
 				// If this is a new size not from classes or bonusHitDice array, get stored current
@@ -76,6 +87,18 @@ class HitDiceManager {
 					currentCounted.add(size);
 				}
 				this.dieSizes.add(size);
+			}
+		}
+
+		const bySize = this.bySize;
+		this.#max = 0;
+		this.#value = 0;
+		this.dieSizes = new Set<number>();
+		for (const [size, { current, total }] of Object.entries(bySize)) {
+			this.#max += total;
+			this.#value += current;
+			if (total > 0) {
+				this.dieSizes.add(Number(size));
 			}
 		}
 	}
@@ -143,7 +166,7 @@ class HitDiceManager {
 			const baseSize = Number(sizeStr);
 			const size = incrementDieSize(baseSize, hitDiceSizeBonus);
 			const bonus = (hitDieData as { bonus?: number }).bonus ?? 0;
-			if (bonus > 0) {
+			if (bonus !== 0) {
 				hitDiceByClass[size] ??= { current: 0, total: 0 };
 				hitDiceByClass[size].total += bonus;
 
@@ -156,7 +179,7 @@ class HitDiceManager {
 			}
 		}
 
-		return hitDiceByClass;
+		return clampHitDiceBySize(hitDiceByClass);
 	}
 
 	getHitDieData(size: number) {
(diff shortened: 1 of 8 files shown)


Fronix:
Thanks for contributing @puneetdixit200! I'll take a look at the PR when i get the chance.

In the future please use our PR template, I review a lot of PRs right now and having the same structure helps.

src/managers/HitDiceManager.ts
```diff
@@ -78,6 +89,18 @@ class HitDiceManager {
...
+		this.#max = 0;
+		this.#value = 0;
+		this.dieSizes = new Set<number>();
+		for (const [size, { current, total }] of Object.entries(bySize)) {
+			this.#max += total;
+			this.#value += current;
+			if (total > 0) {
+				this.dieSizes.add(Number(size));
+			}
+		}
```
Fronix, line 33: This block recomputes `#max`, `#value`, and `dieSizes` from `this.bySize`, fully overwriting the values built up in lines 48-91. The first accumulation pass (and the `currentCounted` bookkeeping with it) is now dead work. Either drop the first pass and derive everything from `bySize`, or do the clamp inline so the first pass remains the source of truth.

src/view/dialogs/FieldRestDialog.svelte
```diff
@@ -102,6 +102,11 @@
 			}
 		}
 
+		for (const data of Object.values(bySize)) {
+			data.total = Math.max(data.total, 0);
+			data.current = Math.min(Math.max(data.current, 0), data.total);
+		}
```
Fronix: This loop duplicates `clampHitDiceBySize` from `HitDiceManager.ts` (and the same loop appears in `SafeRestDialog.svelte` and `PlayerCharacterSheet.state.svelte.ts`). Export the helper and reuse it in all three callers so the clamp rule lives in one place.

Fronix (changes requested):
One suggestion in HitDiceManager about dead code in the constructor, plus duplicated clamp logic worth deduping.

**Needs changes:** the constructor's first accumulation pass is dead after the recompute at L93-L103, and the clamp helper is reimplemented inline in three other files instead of being shared.

src/managers/HitDiceManager.ts
```diff
@@ -9,9 +9,11 @@ function incrementDieSize(baseSize: number, steps: number): number {
...
 }
 
-function clampHitDiceBySize(
-	hitDiceBySize: Record<string, { current: number; total: number }>,
-): Record<string, { current: number; total: number }> {
+type HitDiceBySize =
+	| Record<number, { current: number; total: number }>
+	| Record<string, { current: number; total: number }>;
+
+function clampHitDiceBySize<T extends HitDiceBySize>(hitDiceBySize: T): T {
```
Fronix: Prefer if the helper is put in `utils` since there are many usages it shouldn't live here.

Fronix (changes requested):

Fronix (approved):
LGTM

PR #795, opened 2026-05-29
fix(armor): AC not updating on item change; fix Iron Defense level 13

trevlar:
## Summary

- **`NimbleBaseActor.prepareData()`**: Remove the `initialized` guard that was short-circuiting every call after the first `_initialize()`. Foundry v13 calls `prepareData()` directly (not via `_initialize()`) after embedded items are created/deleted, so the guard prevented `afterPrepareData()` from ever re-running — leaving `armorClass` components stale and AC frozen at its initial value.
- **`NimbleCharacter.prepareDerivedData()`**: Reset `armor.components = []` at the top of each cycle so `afterPrepareData()` rules always push into a clean array rather than accumulating stale entries from the previous run.
- **Iron Defense level 13**: The level 13 rule used `mode: "multiply"` which, in `_prepareArmorClass()`, sorts *before* the level 1 `mode: "override"` and gets completely overwritten by it — dead code. Changed to `mode: "override"` with `formula: "(@dexterity + @strength) * 2"` and `priority: 2` so it runs after the level 1 rule and produces the correct doubled value.
- **Pack data cleanup**: Normalize `armorClass` formula strings — strip leading `+` from `"+4"` / `"+8"` values in turtlefolk and resolute-fangs-golden-bastion.

## Test plan

- [ ] Add Dragonborn ancestry to a character — AC increases by 1
- [ ] Remove Dragonborn ancestry — AC decreases by 1
- [ ] Re-add Dragonborn ancestry — AC increases by 1 again
- [ ] Zephyr with Iron Defense (unarmored, level 1–12): AC = DEX + STR
- [ ] Zephyr with Iron Defense (unarmored, level 13+): AC = (DEX + STR) × 2
- [ ] Zephyr with Iron Defense wearing armor: Iron Defense predicate does not fire, AC from armor applies normally
- [ ] All 1091 existing tests pass (`pnpm check`)

### DIFF
--- a/packs/ancestries/core/exotic/turtlefolk.json
+++ b/packs/ancestries/core/exotic/turtlefolk.json
@@ -14,13 +14,18 @@
 				"label": "Slow & Steady",
 				"predicate": {},
 				"priority": 1,
-				"formula": "+4",
+				"formula": "4",
 				"mode": "add"
 			},
 			{
 				"type": "speedBonus",
-				"value": "-2",
-				"label": "Slow & Steady"
+				"disabled": false,
+				"id": "Xk9mNpQ2rLsYvWtH",
+				"identifier": "",
+				"label": "Slow & Steady",
+				"predicate": {},
+				"priority": 1,
+				"value": "-2"
 			}
 		],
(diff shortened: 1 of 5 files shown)


Fronix (approved):
LGTM

PR #800, opened 2026-06-01
feat(levelUp): support selecting one option or another on level up (#708)

### LINKED ISSUE #708: Support selecting one option or another option on level up
### Problem or Motivation

Some class features grant the character a choice between two (or more) structurally different outcomes at level up — for example, "gain +2 max Combat Dice or learn an additional Combat Tactic." Today, class features on level up can only describe a single fixed outcome per feature (e.g. a grantItem rule, or a stat change, or an effect), so this kind of either/or choice can't be expressed by the feature itself. Authors have to document the choice in the feature's description text and rely on the player to manually apply the consequence after leveling up, which means the system never records the choice, can't reverse it on level-down, and can't validate that the player actually picked one.

### Proposed Solution

Allow a class feature to declare a set of mutually exclusive options, where each option specifies its own outcome — a grantItem rule, a stat change, an effect, etc. Each option can be a different type of outcome, not just a different item.

During the level-up flow:
- The CharacterLevelUpDialog renders the feature with a prompt to pick one option, styled consistently with the existing FeatureGroupSelection UI.
- Only the chosen option's outcome is applied to the character.
- The choice is persisted on the granted feature, so CharacterLevelDownDialogState can cleanly reverse exactly what was applied.

### Alternatives Considered

- Author each option as its own feature and group them with selectionGroups. This works when the options are the same kind of thing (e.g., "pick one of three class features"), but falls down when one option is a stat change, and another is an item grant — the current grouping UI expects homogeneous feature cards, and the "option" that's really a stat change has no natural feature to represent it.
- Describe the choice in prose and leave an application manual. Current workaround. Nothing is tracked, nothing is reversible on level-down, and it's easy for players to forget or misapply.
- Split into two separate level-up entries and gate them on a flag. More moving parts than the choice warrants, and the UX doesn't clearly communicate that the two entries are mutually exclusive.

### Additional Context

- Concrete example from the packs: Fit for Any Battlefield (packs/classFeatures/core/commander/commander-progression/fit-for-any-battlefield.json) and similar Commander progression features describe either/or choices in prose at levels 6/8/10/12/16.
- Existing related UI: src/view/dialogs/components/levelUpHelper/LevelUpClassFeatureSelection.svelte and FeatureGroupSelection.svelte — the new picker should match this visual pattern.
- Level-up / level-down flow:
src/view/dialogs/CharacterLevelUpDialogState.svelte.ts,
src/view/dialogs/CharacterLevelDownDialogState.svelte.ts.
- Related recent work: #704 (subclass features on level up), #693 (spells on level up).

trevlar:
## Summary

- Adds `levelUpOptions` to class features so a level-up step can present mutually exclusive choices (e.g. +1 Combat Die vs. choose a Combat Ability vs. choose a Combat Tactic)
- Implements `poolMaxBonus` rule type to increment a charge pool's max and store the result in `levelUpHistory` rather than granting a separate item
- Adds `Migration026` to migrate Fit for Any Battlefield on existing actors
- Updates Commander `fit-for-any-battlefield` with the `levelUpOptions` structure; trims `gainedAtLevels` on combat-tactics and commanders-order sub-items so they no longer appear as standalone grants at higher levels

## Test plan

- [x] Level up a Commander and verify the +1 Combat Die / Choose Combat Tactic / Choose Combat Ability choice appears at the correct levels
- [x] Verify selecting +1 Combat Die increments the combat dice pool max
- [x] Verify selecting a Combat Tactic or Combat Ability works as before
- [ ] Run Migration026 on an actor with Fit for Any Battlefield and confirm orphaned +1 Max Combat Die items are removed
- [x] Confirm all 1442 tests pass


docs/testing/class-progression-test-report.html
Fronix, line 1: Is this file meant to be added here?

docs/testing/compendium-vs-vault-audit.md
Fronix, line 1: Same with this one, looks like working material

Fronix (approved):
## Suggestions (non blocking)

- `src/migration/migrations/Migration028CombatDiceMaxFormula.ts:348` — replaces the entire option `rules` array with a single `poolMaxBonus`, dropping any sibling rules the option carried.
  ```ts
  maxCombatDieOption.rules = [ { type: 'poolMaxBonus', ... } ];
  ```
  Fix: map/replace the `grantItem` entry in place, preserving other rules: `maxCombatDieOption.rules = optionRules.map(r => isGrantItem(r) ? poolBonus : r)`.

- `src/documents/actor/character.ts:74-107,113-138` — `#resolvePoolBonusGrantUuids` and `#syncPoolBonusItemDescriptions` redeclare local `RuleShape`/`ItemShape` and use `as unknown as` casts even though `FeatureDataModel` now declares `levelUpOptions`. Prefer the typed `item.system.levelUpOptions`; drop the ad-hoc shapes (duplicated across both methods).

- `src/documents/actor/character.ts:92-100,146-153` — ownership match is a `compendiumSource.includes(itemId)` substring test; a bare 16-char id substring can false-match. Prefer comparing the full compendium UUID (or the trailing `Item.<id>` segment) exactly.

- `src/view/dialogs/components/levelUpHelper/LevelUpFeatureOptionPicker.utils.ts:22-40` — scans every `game.packs` entry and calls `getIndex` per pack inside a `$effect` on each option change, duplicating the class-feature indexing already built via `buildClassFeatureIndex`. Consider reusing the existing index rather than re-scanning packs.

- `src/documents/actor/character.ts:162-165` — name pluralization via `.replace(/\+\d+/, ...)` + `\b(?:Die|Dice)\b` is cosmetic-only and brittle for any item name not matching that shape; fine for the single known item but won't generalize.

docs/testing/class-progression-test-report.html
trevlar: Good catch — removed in 80670f9. It was a one-off working snapshot, not generated by any tracked script or wired into a pipeline, so it would only have gone stale.

docs/testing/compendium-vs-vault-audit.md
trevlar: Same — removed in 80670f9. It was working material, not a tracked/generated doc. For the record, its conclusion still holds: the selectable option pools and subclass options are complete and correct in membership and level placement for all 11 classes, so the #708 level-up flow surfaces the right options.

trevlar:
Thanks for the review! Follow-ups pushed in 80670f9:

- **Migration028** — the `poolMaxBonus` swap now maps/replaces the `grantItem` entry in place, so any sibling rules on the option survive.
- **character.ts (`#resolvePoolBonusGrantUuids` / `#syncPoolBonusItemDescriptions`)** — dropped the duplicated ad-hoc `RuleShape`/`ItemShape` + `as unknown as` casts in favour of a single shared rule type and the typed `item.system.levelUpOptions` (guarded by `item.isType('feature')`).
- **character.ts ownership match** — no longer a `compendiumSource.includes(itemId)` substring test; it now compares the full compendium UUID exactly.
- **LevelUpFeatureOptionPicker.utils.ts** — no longer re-scans every pack per option change. The pre-built `ClassFeatureIndex` is now retained on the level-up dialog state and threaded down to the picker, which resolves the sub-item pool from it (pool members are indexed under the class key with their `group`, collected across all levels). No behaviour change; existing tests updated to build an index instead of an in-memory pack.

Left the pool-item name pluralization regex as-is per your note that it's fine for the single known item — happy to generalize it if/when a second pool-bonus item shows up.

All gates green locally (tsc, biome lint, 2358 tests, circular-deps).

Issue #681, opened 2026-04-08
[EPIC] Foundation rule types


Fronix:
Foundational rule types required by feature/spell/skill automation issues. Each sub-issue here defines one rule type (shape, acceptance criteria, blocking chain). Feature issues across all classes reference these as `blocked by`.

Parent: #365

Fronix:
Marking this as closed since all rules shipped in 0.8.9

Fronix:
Reopening due to new foundational rules being required

PR #805, opened 2026-06-04
feat(rules): add toggleEffect foundation rule (#577)

### LINKED ISSUE #577: [Foundation] Rule type: toggleEffect
A class feature toggles "on" (Rage, Beastshift, Interpose) and pushes one or more tags into the actor's domain. Sibling rules elsewhere predicate on those tags via the existing `PredicateField`. The toggle rule itself doesn't carry nested rules or express any modifiers; its only job is the tag push.

```jsonc
{
  "type": "toggleEffect",
  "label": "Rage",
  "tags": ["self:raging"],
  "turnOff": ["onActorKilled", "onRest"]
}
```

- `tags`: domain tags added while active, dropped when inactive
- `turnOff`: rule-lifecycle hooks that flip the toggle off

## State storage

Backed by a Foundry `ActiveEffect`, same as `applyCondition` (`src/models/rules/applyCondition.ts:198`). When toggled on, an AE is created (or enabled) on the actor flagged as belonging to this rule. While the AE is enabled, the rule's `tags` are added to `actor.getDomain()` during `prepareDerivedData`. Toggle off (player click, or a `turnOff` trigger) and the AE is disabled or deleted, tags drop on next prep. That gives us the native AE toggle UI on the sheet for free.

## Available `turnOff` triggers

Today's rule-lifecycle hooks (`src/hooks/ruleEventDispatch.ts`): `onActorKilled`, `onActorWounded`, `onRest`, `onTurnStart`, `onTurnEnd`.

The original issue listed `onEncounterEnd` and `onUnconscious`. Those aren't wired to the rule dispatcher today. Rage really does end when the encounter ends, so this issue should also add those triggers to the dispatcher. Small addition, established pattern.

## Worked example: Rage

```jsonc
[
  {
    "type": "toggleEffect",
    "tags": ["self:raging"],
    "turnOff": ["onActorKilled", "onEncounterEnd", "onRest"]
  },
  {
    "type": "damageBonus",
    "value": "@level",
    "delivery": "melee",
    "source": "weapon",
    "predicate": { "self": "raging" }
  },
  {
    "type": "damageReduction",
    "value": "2",
    "predicate": { "self": "raging" }
  }
]
```

Intensifying Fury and the other ~10 "while raging" features live on their own items and just predicate on `self:raging`. They don't need to know how Rage is toggled.

## Not in this issue

- Target-tag application (Hunter's Mark putting `target:marked` on another actor). Different mechanism, lives with that feature.
- Concentration. When it lands it becomes another `turnOff` value.
- Auto-activation passives ("while unarmored", "while bloodied"). Those don't need this rule at all. A `predicate` on the sibling rule is enough. This rule is only for things a player or a trigger turns on.

---

Needs #579 first (defines the `self:*` namespace).

tristin-albers:
## Summary

Introduces the `toggleEffect` foundation rule type: an AE-backed,
player-controlled toggle that pushes domain tags into the actor's
domain while active. Sibling rules predicate on those tags via the
existing `PredicateField` (e.g. a `damageBonus` with
`predicate:{self:'raging'}`). Closes #577.

## Changes

- New rule type `toggleEffect` in `src/models/rules/toggleEffect.ts`
  with schema `{ tags: string[], turnOff: string[] }`.
- Three new dispatcher events wired into `ruleEventDispatch.ts`:
  - `onItemActivated` (from `nimble.useItem`) — toggle entry point
  - `onEncounterEnd` (from `updateCombat(started:false)` +
    `deleteCombat` fallback, deduped per combat id; mirrors the
    existing `encounterEndTrigger.ts` pattern)
  - `onUnconscious` (filtered from `nimble.conditionApplied` for the
    unconscious status)
- Lazy AE lifecycle with three-state toggle: create if absent,
  re-enable if disabled, delete if enabled. AE flagged with
  `nimble.toggleEffectRuleId` + `toggleEffectItemId` and stamped
  with `origin: item.uuid` so the effects panel links back to source.
- `prePrepareData()` pushes tags only when the AE exists and is not
  disabled; tags drop on next prep when the AE is gone or disabled.
- `turnOff` choices: `onActorKilled`, `onActorWounded`, `onRest`,
  `onTurnStart`, `onTurnEnd`, `onEncounterEnd`, `onUnconscious`.
- Co-located unit tests + extended dispatcher tests. Docs in
  `docs/system/rules.md` with worked Rage example and priority note.
- i18n strings in `en.json` (`NIMBLE.ruleTypes.toggleEffect` +
  `NIMBLE.rules.toggleEffect.*`).

## Test Plan

Manual verification against a live Foundry world (test character
with an injected Rage feature carrying `toggleEffect` +
sibling `damageBonus`, deleted after):

- [x] Activate → AE created with correct name, img, origin, and flags;
      \`self:raging\` tag present
- [x] Sibling \`damageBonus\` consumes \`self:raging\` (resolved \`@level\`
      to 1 on a level-1 character)
- [x] Disable AE via panel → tag drops, bonus gone
- [x] Re-activate disabled AE → re-enabled (single AE, no duplicate)
- [x] Activate enabled AE → deleted
- [x] \`onRest\` → AE deleted
- [x] \`onEncounterEnd\` (combat ended) → AE deleted
- [x] \`onActorKilled\` (HP→0) → AE deleted
- [x] \`onUnconscious\` → AE deleted
- [x] \`pnpm check\` passes (format, lint, circular deps, type-check,
      1480 tests)

## Checklist

- [x] PR targets the \`dev\` branch
- [x] \`pnpm check\` passes
- [x] No hardcoded user-facing strings (used \`localize()\` keys; AE
      label falls back to item name)
- [x] **If yes:** I have reviewed all AI-generated code against the
      repo's STYLE_GUIDE.md and AGENTS.md

## Related Issues

Closes #577
Part of epic #681

---

**edit:** scope of this PR grew beyond the original `toggleEffect` foundation rule once Rage was wired in as the demonstrator (per @Fronix's review ask). Full recap in [this thread comment](https://github.com/Nimble-Co/FoundryVTT-Nimble/pull/805#issuecomment-4635982018). Headline additions:

- `findDiceConsumerRule` now honors `appliesTo()` so `diceConsumer.predicate` actually gates autoBonus pools (e.g. Fury Dice on `self:raging`).
- `onItemActivated` semantics: item use is "ensure on" only (no more silent toggle-off on re-use).
- `rollDieIntoPool` always rolls at max and returns `{ applied, face }`; chat card shows the rolled face dimmed with "— pool at max" per RAW.
- Opt-in `suppressChat` on pool effect nodes (Rage opts in; eliminates duplicate roll chat).
- New `toggleEffect` fields: `confirmEndPrompt` (i18n key for end-confirmation dialog) and `clearPoolsOnEnd` (pool identifiers emptied when toggle ends — uniform across switch click + `turnOff` triggers). Prompt is skipped when no linked pool has dice (nothing to lose).
- New on/off switch on attack-feature rows in the Actions panel, Foundry-style (matches conditions tab), reactive to external AE changes.
- Rage in the compendia uses all of the above: `toggleEffect` with `tags: ["self:raging"]`, predicated `diceConsumer`, opt-in `suppressChat` on rollDie effects, `confirmEndPrompt`, and `clearPoolsOnEnd: ["fury"]`.

Two test-coverage follow-ups noted in the review comment (clearPoolsOnEnd ordering on turnOff; disable-path prompt flow in `toggleEffectControl`). Both gated on mock-resolution issues; integration-verified manually.


Fronix:
@tristin-albers awesome work, coudl you ship at least one feature using this rule so we can ensure the base functionality works and so I don't have to make one and test? :D

tristin-albers:
@Fronix  I finished up Rage, but also for the essence of completeness and user value I expanded the scope to make Rage as an example of something with great automation but still manually overridable easily. I would really like for you to check it out, I also made some decisions in here that made sense for UI/UX but if this isn't the approach we want to take for those I can roll it back and scope down the PR.

Rage in `packs/classFeatures/core/berserker/berserker-progression/rage.json` now uses the rule end-to-end. While shipping it I hit a chain of issues that only surface once a real feature exercises the foundation, so the PR grew. Posting the recap here so the diff isn't a surprise.

## What the foundation rule needed to actually ship Rage

1. **`diceConsumer.predicate` was a no-op for autoBonus pools.** `findDiceConsumerRule` matched by type + pool identifier only and never called `appliesTo()`. Without this, predicating Fury Dice on `self:raging` does nothing — the autobonus still fires when Rage is off. Fixed; the modifyPool loop already had the same `appliesTo()` check, so this just mirrors it.

2. **Item-use was a silent off-switch.** The original three-state toggle (create / re-enable / delete) made the second click on Rage (e.g. to roll another Fury Die next turn) drop the effect. Changed semantics so item activation is "ensure on" only — turning off goes through the AE in PASSIVE EFFECTS, a `turnOff` trigger, or the new switch (below). Updated `docs/system/rules.md` to match.

3. **At-max RAW.** Per the Berserker rules: "If you are already at your max, roll as normal and decide which ones to keep." `rollDieIntoPool` now always evaluates the roll and returns `{ applied, face }`; the chat card surfaces the face dimmed with "— pool at max" instead of just "— no change."

4. **Duplicate roll chat.** The activation pipeline was posting a redundant standalone roll card on top of the feature's activation card. Added an opt-in `suppressChat` field on the pool effect node (default off; manual widget / refill paths unchanged). Rage opts in. Authors get a checkbox in the rules builder.

## The Rage UX bits

5. **On/off switch on the actions panel.** Attack-feature rows now show a Foundry-style toggle (matches the conditions tab styling) when the item has a non-disabled `toggleEffect` rule. Click flips the AE. Reactivity via `\$effect` + AE hooks so the switch reflects external changes (PASSIVE EFFECTS, turnOff triggers).

6. **`confirmEndPrompt` + `clearPoolsOnEnd`** on the `toggleEffect` schema. RAW Rage: "Fury Dice are lost when your Rage ends." Both the new switch and every `turnOff` trigger now clear configured pools. The confirm prompt only fires when there's actually dice in the linked pools to lose — empty pool = no interruption.

## Known follow-up (not in this PR)

When a level-5+ berserker rolls `+2 Fury Dice` at max, the activation card renders both faces with the same opacity even though one was added and one was discarded. Proper UX is a dice-picker dialog at end of roll where the player chooses which to keep. Out of scope here; will track as its own issue.

## Notes for review

- All foundation rule criteria from #577 covered + Rage compendium exercises them end-to-end.
- 1494 tests passing. Two follow-up coverage gaps left: `clearPoolsOnEnd` ordering on turnOff, and the disable-path prompt-flow in `toggleEffectControl` — both blocked on `vi.mock` not intercepting cleanly with the DataModel internals / module resolution. Integration-verified manually in Foundry.
- New on/off switch and the confirm-pool-clear flow are exercised by hand in a live world (Rage on/off, prompt with vs. without dice, turnOff via Long Rest).

image

public/lang/en.json
```diff
@@ -941,6 +942,33 @@
...
+					"label": "Turn off on",
+					"hint": "Events that automatically toggle the effect off."
+				},
+				"confirmEndPrompt": {
+					"label": "End confirm prompt (i18n key)",
+					"hint": "When set, the explicit on/off switch shows a confirm dialog before turning the effect off. Provide an i18n key whose value is the dialog body. Leave empty to disable the prompt."
+				},
+				"clearPoolsOnEnd": {
+					"label": "Clear pools on end",
+					"hint": "Pool identifiers to empty when the toggle ends — either via a turnOff trigger or the player switching it off. Use for rules like 'resources are lost when the effect ends'."
```
Fronix: remove em-dash

public/lang/en.json
```diff
@@ -1850,6 +1881,8 @@
 					"valueLabel": "Value",
 					"predicateLabel": "Predicate (JSON)",
 					"predicateInvalid": "Predicate must be valid JSON",
+					"suppressChatLabel": "Suppress roll chat card",
+					"suppressChatHint": "When enabled, the helper does not post a standalone roll card — use when the feature's own activation card already shows the rolled faces.",
```
Fronix: Lets avoid using em-dashes, it looks ugly and screams AI generated which in this case it is.

Fronix (changes requested):

Fronix:
> @Fronix I finished up Rage, but also for the essence of completeness and user value I expanded the scope to make Rage as an example of something with great automation but still manually overridable easily. I would really like for you to check it out, I also made some decisions in here that made sense for UI/UX but if this isn't the approach we want to take for those I can roll it back and scope down the PR.

Looks good!

And yes you are correct, we intentionally didn't make everything since these details usually surface when trying to fully implement so the scope creep is fine here.

I will do a full review later today as I'm a bit busy right now.

public/lang/en.json
tristin-albers: Fixed in `ddbee7c3` — removed the em-dash from this hint and from every other em-dash I'd introduced in the PR (other en.json strings, docs, code comments, test `describe`/`it` names). Pre-existing em-dashes in unrelated code left alone.

public/lang/en.json
tristin-albers: Fair point — fixed in `ddbee7c3`. Removed the em-dash from this string and from the broader principle: every em-dash I introduced in this PR is gone (other en.json strings, docs/system/rules.md additions, source comments, test names). Pre-existing em-dashes in unrelated code left alone to keep the diff focused.

tristin-albers:
Okay went back through and cleaned out the em-dashes. I agree they look awful, sorry about that. Thank you for the second eyes on it!

packs/classFeatures/core/berserker/berserker-progression/rage.json
```diff
@@ -7,6 +7,27 @@
...
+				"label": "Rage",
+				"predicate": {},
+				"priority": 1,
+				"tags": [
+					"self:raging"
+				],
+				"turnOff": [
+					"onActorKilled",
+					"onEncounterEnd",
+					"onRest"
```
Fronix, line 25: According to RAW, Rage also ends after 1 round without attacking or Raging, which isn't covered here. And `onActorKilled` ends Rage at 0 HP for every level, but at level 18 (Deep Rage) dropping to 0 HP no longer ends Rage, so this misfires for high-level Berserkers.

src/models/rules/toggleEffect.ts
```diff
@@ -0,0 +1,266 @@
...
+		if (!existing) {
+			await this.#createActiveEffect();
+			return;
+		}
+		if (existing.disabled) {
+			await existing.update({ disabled: false });
+		}
+	}
+
+	override async onActorKilled(context: ActorHealthContext): Promise<void> {
```
Fronix, line 248: This unconditionally turns the toggle off at 0 HP. According to RAW, Deep Rage (level 18) makes dropping to 0 HP no longer end Rage, so the `onActorKilled` turn-off needs to be able to be gated rather than always firing.

src/utils/toggleEffectControl.ts
```diff
@@ -0,0 +1,179 @@
...
+	await actor.createEmbeddedDocuments('ActiveEffect', [
+		{
+			name: item.name,
+			img: item.img,
+			disabled: false,
+			origin: item.uuid,
+			flags: {
+				[SYSTEM_ID]: {
+					toggleEffectRuleId: ruleId,
+					toggleEffectItemId: item.id,
```
Fronix: `toggleEffectRuleId` / `toggleEffectItemId` are duplicated as string literals here and also defined in `toggleEffect.ts`. Export the constants from the rule and import them so the two files can't drift.

src/utils/toggleEffectControl.test.ts
```diff
@@ -0,0 +1,145 @@
...
+					nimble: { toggleEffectRuleId: 'rage-toggle', toggleEffectItemId: 'rage-item' },
+				},
+			});
+		});
+	});
+
+	// TODO(follow-up): unit tests for toggleEffectAE disable path:
+	// confirmEndPrompt skip when pool empty, prompt + confirm/cancel branches.
+	// Requires mocking buildEffectiveDicePoolMap which doesn't intercept
+	// cleanly in the current vi.mock setup. Integration-verified manually.
```
Fronix: The disable-path branches (confirm/cancel, skip-when-empty) and `clearPoolsOnEnd` ordering are the most regression-prone paths and are currently untested. The mock setup is already in place; please close this gap rather than deferring.

Fronix (dismissed):
The foundation rule is solid, but a few things need adjusting:

**1. No migration for the Rage compendium change.** `rage.json` is rewritten with the new `toggleEffect` / `diceConsumer` / `clearPoolsOnEnd` shape, but there's no migration script. Characters who already own the old Rage item keep the stale rules and won't pick up the new toggle behavior. A migration needs to reconcile already-embedded Rage items.

**2. Orphaned ActiveEffect on item deletion.** When the owning item is deleted while the toggle is on, the backing AE is left behind on the actor as a dead entry. The `origin: item.uuid` stamp does not make Foundry auto-delete it. Add an `afterDelete()` that removes the AE matching this rule's id.

**3. Rage end conditions don't match RAW / break at high level.** See inline comments.

**4. Fury Die size is hardcoded `d4`.** According to RAW the die scales with Intensifying Fury (d6 at 6, d8 at 9, d10 at 13, d12 at 17), so the Rage item is only correct through level 5.

**5. Active toggles don't surface as active.** A toggleEffect-backed AE (Rage) currently only renders in the Passive Effects bucket of ActorConditionsList, because prepareActorConditions only picks up effects carrying a status, and the toggle AE has none. An active, player-controlled effect shouldn't sit under "passive", it should appear in an active section while it's on.
> Rather than giving the AE a Foundry status (which would route toggle-off through toggleStatusEffect and skip the clearPoolsOnEnd cleanup + confirm prompt), detect the AE by its existing toggleEffectRuleId flag and render it in its own active-toggles section with an off-switch that calls the same toggleEffectAE(...) entry point the action-row switch uses. That keeps the rule as the single owner of the toggle lifecycle and makes the off behavior consistent everywhere (panel switch, conditions list, turnOff triggers).

> **Note this also interacts with the orphaned-AE point above:** the Passive Effects section has no remove button, so today a stale Rage AE (after item deletion) is both visible and unremovable from this list.

Inline comments cover the rest.

packs/classFeatures/core/berserker/berserker-progression/rage.json
Fronix: Implemented in 0ec48e2f. toggleEffect gained a generic endAfterInactiveRounds field: activity (re-activating the owning item or activating any item with a damage effect) is stamped on the backing AE per combat round, and at the actor's turn end after a full inactive round the active GM gets a confirm dialog before Rage ends (declining resets the clock). Round rewinds clamp stale future stamps via a new onRoundChanged rule event, and automatic ends post a chat message explaining why. rage.json opts in with endAfterInactiveRounds: 1.

src/models/rules/toggleEffect.ts
Fronix: Resolved across 8926df0d and 0ec48e2f. Two parts: the dispatcher now distinguishes dying from dead (0 HP alone dispatches onUnconscious; onActorKilled requires a full wound track, or no wound track at all for monsters), and a new generic modifyToggle rule type lets one item suppress named turn-off triggers of a sibling toggle, mirroring the modifyPool pattern. deep-rage.json carries a modifyToggle rule suppressing onUnconscious, so at 18 dropping to 0 HP keeps Rage running while actual death still ends it.

src/utils/toggleEffectControl.ts
Fronix: Fixed in 122e550d. TOGGLE_EFFECT_RULE_ID_FLAG and TOGGLE_EFFECT_ITEM_ID_FLAG are now exported from the rule module and imported here, and the duplicated AE-create payload was extracted into a shared buildToggleEffectAEData helper used by both the rule and the control paths.

src/utils/toggleEffectControl.test.ts
Fronix: Closed in 122e550d, TODO removed. The deferral reason turned out to be real but solvable: tests/setup.ts boots the init hook, which preloads the real dicePool modules before per-file vi.mock can register, so a static import of the module under test binds to the cached real modules. The fix (vi.resetModules plus a dynamic import) is documented in the test file. Full coverage now: confirm, cancel, dismissed dialog, skip-when-empty, plain-text prompt, and clearPoolsOnEnd ordering before AE deletion.

Fronix:
All review items are addressed in ddbee7c3..a6142707. Summary against the review:

**1. Migration:** `b7e2c3c9` adds Migration026RageToggleEffect (schema 26). It reconciles embedded Rage and Deep Rage by compendium source id, with a berserker class + name fallback for progression-granted copies without one, and replaces stale rule shapes rather than skipping them.

**2. Orphaned AE on item deletion:** `0ec48e2f` implements ToggleEffectRule.afterDelete to remove the backing AE, and the item deletion dispatch now runs only on the initiating client so cleanup happens exactly once.

**3. End conditions:** the dispatcher now separates dying from dead (0 HP dispatches onUnconscious; onActorKilled requires a full wound track, and wounds-only updates are watched so the final wound at 0 HP triggers it). Deep Rage suppresses only the 0 HP end via the new generic modifyToggle rule type (`8926df0d`). The RAW inaction end is implemented as endAfterInactiveRounds with GM adjudication, rewind-safe round tracking via a new onRoundChanged rule event, and chat messages for every automatic end. The dispatcher also registers combatRound alongside combatTurn, fixing a pre-existing gap where the last combatant in initiative never received turn events.

**4. Fury Die scaling:** already implemented by intensifying-fury.json's level-predicated modifyPool rules flowing through the effective pool map; `570c5ef5` pins every RAW breakpoint (d4/d6@6/d8@9/d10@13/d12@17) with a regression test so it cannot silently regress.

**5. Passive bucket:** `122e550d` surfaces toggle-backed AEs in Temporary Effects with an off switch that routes through the rule lifecycle (confirm prompt, pool clearing), and in the canvas conditions panel with right-click to end. No Foundry status involved, per the review direction. confirmEndPrompt also accepts plain text now so homebrew items do not need translation files.

Extras from live testing: `a6142707` fixes wound pips beyond six being clipped by the sheet overlay (pre-existing, surfaced while testing max-wound features against the death check).

---

**Bugfixes not directly related to the feature**

1. onActorKilled fired on any drop to 0 HP. The dispatcher conflated dying with dead. Now 0 HP dispatches onUnconscious, and onActorKilled requires a full wound track (or no wound track at all, which is how monsters die at 0). Before this, any "on death" automation would have triggered on every downed PC.
2. Death by the final wound never dispatched at all. The dispatcher only watched hp.value changes, but a dying PC's killing blow is a wounds-only update — HP never moves. Any death trigger was unreachable through the most common way PCs actually die.
3. The last combatant in initiative never received turn events. Foundry fires only combatRound on round wraps — never combatTurn — and the dispatcher only listened to combatTurn. So whoever went last got no onTurnEnd, ever, and whoever went first got no onTurnStart. Verified against the v13 source before fixing.
4. Turn events fired on combat rewinds. Stepping the tracker backwards dispatched bogus onTurnEnd/onTurnStart for turns being undone — core itself runs no lifecycle events on rewinds. Rules with turn-end triggers could fire spuriously mid-correction.
5. Item-deletion rule cleanup ran on every connected client. afterDelete dispatched everywhere, so maxHpBonus removal could subtract the HP bonus once per connected owner, and non-owner clients threw permission errors. Now only the initiating client runs it.

UI bug:

6. Wound pips beyond six were invisible. The sheet overlay was hard-clipped to exactly six pips, so Mighty Endurance, Dwarf, and the max-wound backgrounds rendered wound tracks you couldn't see or click — your 7-wound Rager exposed it. (Plus a contrast bump: revealed unwounded pips were near-black on the dark card.)

src/hooks/ruleEventDispatch.ts
```diff
@@ -132,10 +166,26 @@ function handleActorUpdate(actor: Actor.Implementation, changes: Record<string,
 	};
 
 	if (currentHp <= 0) {
-		void dispatch(actorWithRules, 'onActorKilled', healthContext);
+		// Dropping to 0 HP alone means dying/unconscious, not dead. An actor
```
trevlar: Nimble doesn't cause a PC to go unconcious, that would be a variant rule. As per the rules of Nimble.


```
Hit Points (HP) represent your ability to endure damage. Damage reduces your HP (which can't go below O). When reduced to 0 HP, gain 1 Wound; you also gain the Dying condition until you regain HP. While Dying, actions are limited to 1, Concentration is broken, and you are at risk of further serious harm:
• Attacking/casting spells causes 1 Wound unless you make a DC 10 STR save.
• Taking damage while Dying causes 2 Wounds; a crit causes 3 instead.

Wounds
Wounds are serious injuries you've taken; they are a long term gauge of how close you are to death. HP can usually be recovered quickly, but Wounds may take many days of resting to fully recover from (usually 1/Safe Rest).

Death
You die when you have taken 6 Wounds (unless you have an ability that changes this number). There are ways to revive a hero who has died, but they are rare and often come at a very steep cost.

Temporary HP
Some abilities or effects may grant temporary HP (temp HP); these are reduced first when taking damage. Temp HP do not combine: If a hero has temp HP and would gain more, they instead choose which amount to keep. They expire after a Safe Rest.
Hit Dice
Hit Dice (HD) represent your ability to quickly recuperate from minor injuries and are spent to regain HP. Heroes start with a max of 1 Hit Die at level 1, and this limit increases by 1 each time they level up. Hit Dice are recovered during a Safe Rest (see "RESTING & DOWNTIME" on page 16).

Gritty Dying Rules. Optional rules for a more challenging or "old school" style of game: You can reduce the max number of Wounds heroes have. Any-
where from a max of 5 Wounds for slightly harder, all the way down to 1-2 Wounds for a very lethal game! You can also have each Wound impose a cumulative
-1 penalty to rolls: skill checks, saves, or even ALL rolls! (If you're a real glutton for punishment.)
```

docs/system/rules.md
```diff
@@ -184,6 +184,63 @@ When no target is selected, bonuses with `targetCondition` are excluded. Bonuses
...
+
+| Value | Fires from |
+|---|---|
+| `onActorKilled` | `updateActor` when HP drops to 0 |
+| `onActorWounded` | `updateActor` on bloodied / lastStand transition |
+| `onRest` | `nimble.rest` hook |
+| `onTurnStart` | `combatTurn` start of owner's turn |
+| `onTurnEnd` | `combatTurn` end of owner's turn |
+| `onEncounterEnd` | `updateCombat(started:false)` or `deleteCombat` (dedup'd) |
+| `onUnconscious` | `nimble.conditionApplied` with `condition: 'unconscious'` |
```
trevlar: Unconcious??? There isn't a trigger listed in Nimble only case would be GM is going non-lethal to begin an escape arc or being captured is part of the story.

New Nimble rulebook is the same on dying

```
Hit Points & Dying
Hit Points (HP) represent your ability to endure damage. Damage reduces your HP, when reduced to 0 HP, gain 1 Wound; you also gain the Dying condition until you regain HP. While Dying, your actions are limited to 1, Concentration is broken, and you are at risk of further serious harm:
• Attacking/casting spells causes 1 Wound unless you make a DC 10 STR save.
• Taking damage while Dying causes 2 Wounds; a crit causes 3 instead.

TEMPORARY HP
Some abilities or effects may grant temporary HP (temp HP); these are reduced first when taking damage. Temp HP do not combine: If a hero has temp HP and would gain more, they instead choose which amount to keep. They expire after a Safe Rest.

HIT DIcE
Hit Dice (HD) represent your ability to recover from minor injuries. Heroes start with a max of 1 Hit Die at level 1, and this limit increases by 1 each time they level up. Hit Dice are spent to regain HP while resting and are recovered during a Safe Rest (see "Safe Rests" on pg. 17).

WOUNDS
Wounds are serious injuries you've taken; they are a long term gauge of how close you are to death. HP can usually be recovered quickly, but Wounds may take many days of resting to fully recover from (usually 1/Safe Rest).

DEATH
You die when you have taken 6 Wounds (unless you have an ability that changes this number). There are ways to revive a hero who has died, but they are rare and often come at a very steep cost.

Gritty Dying Rules. Fora more challenging or 'old school" game reduce the max number of wounds heroes have: 5 Wounds for slightiy harder, all the way down to 2 Wounds for a very lethal game!
```

Conditions:
`• Dying. At 0 HP.Max 1 action. Taking damage while Dying causes 2 Wounds, a crit causes 3 instead.`

docs/system/rules.md
trevlar, line 1: The berserker's effect is not while unconcious or at 0 hit points or when he takes a wound. It's while dying.

LEVEL 4

Enduring Rage. While Dying, you Rage automatically for free at the beginning of your turn, have a max of 2 actions instead of 1, and ignore the STR saves to make attacks.


A few savage arsenal rules mention it.

```
Deathless Rage. (1/turn) While Dying, you may suffer 1 Wound to gain 1 action.
Unstoppable Force. While Dying and Raging, taking damage causes 1 Wound (instead of 2) and critical hits inflict 2 Wounds (instead of 3).
Mighty Endurance. You can now survive an additional 4 Wounds before death.
```

One of the subclass abilities at lvl 15

```
Mountain's Endurance. While Dying, if an attack against you would be a crit, the attack is recolled instead (when-crit abilities, such as Titan's Fury, still trigger).
```

I checked all three books Nimble GMGuide, Nimble Core Rules, and Heroes guide. No mention of unconcious trigger anywhere. Only case where that would be used is Nimble homebrew or following more closely D&D 5e rules.

trevlar (dismissed):
A few concerns, I have not tested it but the mention of unconcious triggers makes me worried that we added a new unnecessary trigger.

docs/system/rules.md
Fronix: > DEEP RAGE. Dropping to 0 HP does not cause your Rage to end

This is the reason for the "unconcious" but I will rename it to `onActorDying` to better align with nimble rules. We will have to deal with the other features like deathless rage, unstoppable force etc later.

src/hooks/ruleEventDispatch.ts
Fronix: Renamed to `onActorDying` instead

docs/system/rules.md
Fronix: Renamed to onActorDying instead

trevlar (dismissed):
**🔧 Push back — the new automatic turn-off paths run on every connected client, so Rage ending will double-fire at a real table**

This is a big, well-built feature: the `toggleEffect`/`modifyToggle` rules are cleanly generic, Rage is wired in purely through packs + a migration (no feature-specific coupling in the rule), i18n/registration/types/docs are all complete, and the test coverage is genuinely thorough. One correctness issue holds it back.

**Must fix before merge**

- **`src/models/rules/toggleEffect.ts:304`** — `#maybeTurnOff` (and `#maybeEndFromInactivity`'s sibling logic) deletes the backing AE, clears linked pools, and posts a chat message, but the turn-off path has no single-executor guard. Every `turnOff` event (`onEncounterEnd`, `onActorKilled`, `onActorDying`, `onRest`, `onTurnStart/End`) is dispatched on **every connected client** — these hooks (`updateCombat`, `updateActor`, `combatTurn`, `nimble.rest`) fire everywhere, and `dispatch()` has no ownership/GM check. At a normal table the GM's client *and* the owning player's client both own the actor, so both run `#maybeTurnOff`: both call `#clearLinkedPools()`, both `deleteEmbeddedDocuments`, and both reach `#announceEnd`. Result: **duplicate "Rage ended: …" chat messages** and a redundant delete on essentially every automatic end of Rage. The comment at `toggleEffect.ts:431` claims "only the client that won the delete announces," but that relies on `deleteEmbeddedDocuments` throwing *before* the announce for the loser — the two clients both pass the `if (!existing) return` guard before either delete round-trips, so it's a race and delete-throw-on-missing-id is version-dependent. You already solved this exact class of bug elsewhere: `_onDelete` was just gated to `game.user?.id === userId`, the inactivity path is gated to `activeGM.isSelf`, and `src/hooks/automaticConditions.ts` has a purpose-built `isActiveGM()` helper whose docstring describes precisely this failure ("one duplicate effect per connected user with permission"). Gate the destructive turn-off work with that same `isActiveGM()`-style check so the delete + pool clear + announce happen exactly once.

  Note this compounds with a second source: `onActorDying` is dispatched from **both** `ruleEventDispatch.ts:183` (HP→0 with wounds below max) and `ruleEventDispatch.ts:318` (`conditionApplied: 'dying'`), so a dying PC can trigger the turn-off up to 4× across the two hooks × two clients. A single-executor guard on the turn-off mutation covers this too.

- **`src/view/sheets/components/WeaponCard.svelte:242`** — the "toggle on" state hardcodes `hsl(120, 45%, 40%)` (and `50%` on hover) where the rest of this file and the codebase use `--nimble-*` CSS vars. Pull it into a variable so it tracks the theme.

**What's good**

- `toggleEffect` and `modifyToggle` are named after *what they do*, carry no Rage-specific logic, and Rage is expressed entirely in `rage.json` + `Migration027` — exactly the generic-building-block model the rules system is designed for.
- Migration is idempotent, replace-or-insert (upgrades stale partial copies), and matches the pack JSON exactly (`rage-fury-l1/l5`, `self:raging` predicate, `suppressChat`).

---

1. Gate turn-off delete/pool-clear/announce to one client (isActiveGM)
2. Dedup onActorDying across HP-drop and conditionApplied sources
3. Replace hardcoded toggle-on green with a CSS var

trevlar (approved):
**✅ Approved**

This introduces the `toggleEffect` foundation rule (AE-backed player toggle that pushes domain tags) plus the `modifyToggle` companion, wires seven lifecycle dispatcher events, and demonstrates the whole thing end-to-end on Berserker Rage. It's a big surface but it holds together — clean separation between the rule pipeline and the UI switch, and the generic naming keeps it homebrew-friendly per the rules design philosophy.

- Multi-client safety is handled correctly: turn-off/delete/pool-clear paths gate on `isActiveGM()`, `_onDelete` cleanup gates on the initiating user, and AE creation rides the single-client `useItem`/switch paths.
- Strong, genuinely-behavioral test coverage (206 tests across the changed files, including dispatcher dedup, inactivity clamp-on-rewind, and the conditions-list off switch).
- Migration 027 is idempotent, matches the shipped compendium JSON exactly, and correctly renumbered with the `LATEST_SCHEMA_VERSION` bump to 27.

Verified locally against the PR head: typecheck clean, biome lint clean, `en.json` valid, all touched test files green, `RuleCard.allRules` covers both new rule types.

One tiny, non-blocking note if you're back in here: `toggleEffect.ts` gates on `game.users?.activeGM?.isSelf` inline in `#maybeEndFromInactivity` (src/models/rules/toggleEffect.ts:376) and `onRoundChanged` (:288), but uses the new shared `isActiveGM()` helper in `#maybeTurnOff` (:312). Since you extracted `isActiveGM()` specifically to standardize this gate, routing all three through it would be a nice consistency win (and picks up the any-GM fallback for free).

PR #813, opened 2026-06-20
feat: ship GM-friendly default token settings

trevlar:
## Summary

Ships GM-friendly token defaults so freshly created tokens are usable without per-token setup — for both player characters and adversaries.

Motivation: a recurring Discord thread where GMs had to manually enable vision on every player token. Root cause — `character.ts` set `prototypeToken.vision = true`, but `vision` is **not** a field in the Foundry v13 token schema (sight is keyed on `sight.enabled`), so player tokens never actually received vision.

## Changes

**`src/hooks/ready.ts`** — seed Foundry's built-in `core.defaultToken` setting on `ready()` (GM only, world-scoped), mirroring the existing `combatTrackerConfig.skipDefeated` seed. All values use `??=`, so a GM's deliberate token configuration is never overwritten:
- `sight.enabled = true` — vision on. Range stays `0`: Nimble has no darkvision, so tokens see illuminated areas.
- `bar1.attribute = attributes.hp` — HP bar works out of the box.
- `bar2.attribute = resources.mana` — mana bar; `resources.mana` exists only on characters, so monsters/NPCs resolve no attribute and draw no second bar.
- `displayBars = OWNER_HOVER` — bars visible on hover for owners.

**`src/documents/actor/character.ts`** — replace the no-op `vision: true` with `sight: { enabled: true }`; keep PC-specific `actorLink: true` / friendly `disposition: 1`.

**Monster actors (`npc.ts`, `minion.ts`, `soloMonster.ts`)** — these had no `_preCreate`, so they fell back to raw Foundry defaults and a customized global token config could make them friendly/linked. Add explicit defaults via a shared `buildMonsterPrototypeTokenDefaults()` helper that mirrors and inverts the character defaults:
- `disposition = HOSTILE`
- `sight.enabled = true` (range 0)
- `actorLink = false` for NPCs and minions (each dropped token tracks its own HP); **solo monsters override to `actorLink = true`** since a boss is a single persistently tracked token.

## Notes
- A non-caster **character** (has `resources.mana` but max 0) may show an empty second bar; monsters/NPCs are unaffected. Left as-is per YAGNI.
- Scene global-illumination defaulting was considered but deliberately **not** included: it would fire on adventure-module scene imports and override map authors' intentional lighting.

## Testing
- `pnpm type-check` ✅
- biome ✅
- unit tests (incl. new `monsterPrototypeTokenDefaults.test.ts`) ✅

### DIFF
--- a/src/documents/actor/character.ts
+++ b/src/documents/actor/character.ts
@@ -1614,7 +1614,7 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
 		const prototypeToken = {
 			sight: { enabled: true },
 			actorLink: true,
-			disposition: 1,
+			disposition: CONST.TOKEN_DISPOSITIONS.FRIENDLY,
 		};
 		this.updateSource({ prototypeToken } as Record<string, unknown>);
 
--- a/src/documents/actor/character.ts
+++ b/src/documents/actor/character.ts
@@ -1606,8 +1606,16 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
 		user: User.Implementation,
 		// biome-ignore lint/suspicious/noConfusingVoidType: Matching parent class signature
 	): Promise<boolean | void> {
-		// Player character configuration
-		const prototypeToken = { vision: true, actorLink: true, disposition: 1 };
+		// Player character configuration. In Foundry v13 token sight is keyed on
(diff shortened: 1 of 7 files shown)


Fronix (approved):
LGTM

PR #815, opened 2026-06-20
feat(combat): limit actions to 1 while Dying

### LINKED ISSUE #814: Dying condition doesn't limit PCs actions to 1
When a PC has the dying condition, they should only have 1 action available according to the rules. Currently a PC with dying condition has 3 actions available.

Ensure that the combat system handles dying condition properly.

trevlar:
## Summary

Implements the Nimble rule that **while Dying, a creature's actions are limited to 1** (requested by a user in Discord).

> When reduced to 0 HP … you gain the Dying condition until you regain HP. **While Dying, actions are limited to 1**, Concentration is broken, and you are at risk of further serious harm.

This PR scopes strictly to the **action limit**. The other Dying clauses (Wound on attack without a DC 10 STR save, extra Wounds on damage, broken Concentration) are out of scope here.

## Approach

A single dying-aware source of truth, wired through every place actions are read, reset, or spent:

- **`actorHealthState.ts`** — `isActorDying(actor)` + `DYING_MAX_ACTIONS = 1`.
- **`combatantSystem.ts`** — `isCombatantDying()`, `getCombatantResetActions()` (turn-start refill value, capped at 1 while dying), and `getCombatantEffectiveMax()` collapses to 1 while dying (additional actions don't apply).
- **Turn resets** (`combat.svelte.ts`) — end-of-turn, NPC/round reset, and non-character turn restore refill to the dying-capped value.
- **Both action trackers** — character-sheet `ActionTracker` and the top combat tracker (`getCombatantMaxActions`, `getActionState`) render 1 pip and clamp restoration while dying; the sheet hides the "+additional action" button.
- **Mid-turn application** — new hook `dyingActionLimitSync.ts` listens for the `dying` status being applied and immediately clamps the combatant's stored `current` to 1 and clears additional actions, so pre-dying actions can't still be spent (e.g. via `consumeCombatantAction`).

### Why the clamp hook is needed
Capping the *display* max alone isn't enough — `current` persists and is checked when attacking/casting. Without the clamp, a character who drops to Dying with 3 actions left could keep spending them.

## Scope / non-regressions
- **NPCs and solo monsters are unaffected** — their base max is already 1, so the cap is a no-op even for a Last Stand monster that also carries the `dying` status. Only characters (base 3) change.

## Tests
- 41 new tests (`isActorDying`, `isCombatantDying`, `getCombatantResetActions`, `getCombatantEffectiveMax`, dying-capped `getCombatantMaxActions`, and the clamp hook: clamps when over limit, no-ops when within, ignores non-dying effects, GM-only).
- Full suite: **1470 passing**. `type-check`, `biome lint`/`format`, `eslint` (svelte), and `circular-deps` all pass.

### DIFF
--- a/src/documents/combat/combatantSystem.ts
+++ b/src/documents/combat/combatantSystem.ts
@@ -1,4 +1,4 @@
-import { DYING_MAX_ACTIONS, isActorDying } from '#utils/actorHealthState.js';
+import { getActorDyingActionLimit, isActorDying } from '#utils/actorHealthState.js';
 import type { CombatantBaseActions, NimbleCombatantSystem } from './combatTypes.js';
 
 function getCombatantSystem(combatant: Combatant.Implementation): NimbleCombatantSystem | null {
@@ -31,18 +31,16 @@ export function isCombatantDying(combatant: Combatant.Implementation): boolean {
 }
 
 /**
- * Base action max to reset `current` to at the start of a turn, capped at
- * {@link DYING_MAX_ACTIONS} while the combatant is Dying.
+ * Base action max to reset `current` to at the start of a turn, capped at the
+ * actor's Dying action limit (baseline 1, raised by features such as Enduring
+ * Rage) while the combatant is Dying.
  */
 export function getCombatantResetActions(combatant: Combatant.Implementation): number {
 	const baseMax = getCombatantBaseActionMax(combatant);
-	if (isCombatantDying(combatant)) return Math.min(baseMax, DYING_MAX_ACTIONS);
+	if (isCombatantDying(combatant)) {
+		return Math.min(baseMax, getActorDyingActionLimit(combatant.actor));
+	}
 	return baseMax;
-}
-
-export function getCombatantEffectiveMax(combatant: Combatant.Implementation): number {
-	// While Dying the base action max is limited to 1, but additional actions still apply.
-	return getCombatantResetActions(combatant) + getCombatantAdditionalActions(combatant);
 }
 
 export function getCombatantBaseActionCurrent(combatant: Combatant.Implementation): number {
--- a/src/documents/combat/combatantSystem.ts
+++ b/src/documents/combat/combatantSystem.ts
@@ -1,3 +1,4 @@
+import { DYING_MAX_ACTIONS, isActorDying } from '#utils/actorHealthState.js';
 import type { CombatantBaseActions, NimbleCombatantSystem } from './combatTypes.js';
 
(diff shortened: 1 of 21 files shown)


src/documents/combat/combatantSystem.ts
```diff
@@ -25,8 +26,23 @@ export function getCombatantAdditionalActions(combatant: Combatant.Implementatio
...
+export function getCombatantResetActions(combatant: Combatant.Implementation): number {
+	const baseMax = getCombatantBaseActionMax(combatant);
+	if (isCombatantDying(combatant)) return Math.min(baseMax, DYING_MAX_ACTIONS);
+	return baseMax;
+}
+
 export function getCombatantEffectiveMax(combatant: Combatant.Implementation): number {
-	return getCombatantBaseActionMax(combatant) + getCombatantAdditionalActions(combatant);
+	// While Dying the base action max is limited to 1, but additional actions still apply.
+	return getCombatantResetActions(combatant) + getCombatantAdditionalActions(combatant);
```
Fronix: I can't find that this function is called anywhere except in tests?

Fronix (dismissed):
Great, there's two things, one that might mess up for Berserkers and the other is just a unused function

> Enduring Rage. While Dying, you Rage automatically for free at the beginning of your turn,
have a **max of 2 actions instead of 1**, and ignore the STR saves to make attacks.

I'd suggest adding a rule here since it would be quite simple.

Rule itself
```javascript
// engine knows this irreducible baseline
export const DYING_MAX_ACTIONS = 1;

class DyingActionLimitRule extends NimbleBaseRule {
  declare value: string;
  override afterPrepareData(): void {
    if (!this.item.isEmbedded) return;
    if (!this.test()) return;            // predicate from the item's data
    const limit = this.resolveFormula(this.value) ?? DYING_MAX_ACTIONS;
    const cur = foundry.utils.getProperty(actor.system, DYING_ACTION_LIMIT_PATH) ?? DYING_MAX_ACTIONS;
    foundry.utils.setProperty(actor.system, DYING_ACTION_LIMIT_PATH, Math.max(cur, limit)); // highest wins
  }
}
```

CharacterDataModel.ts `system.attributes.dyingActionLimit`
```javascript
dyingActionLimit: new fields.NumberField({
  required: true,
  nullable: false,
  initial: DYING_MAX_ACTIONS,
  integer: true,
  min: 0,
}),
```

`enduring-rage.json`
```json
{
  "type": "dyingActionLimit",
  "value": "2",
  "predicate": { "$and": ["self:dying"] }
}
```

After that the combat system would need up be updated to read the `dyingActionLimit`

Fronix (approved):
LGTM

PR #817, opened 2026-06-20
feat(character): show custom reactions on heroic reactions tab

### LINKED ISSUE #816: Show Custom Reaction Abilities on the Heroic Reactions Tab
### Problem or Motivation

The Heroic Reactions tab on the player character sheet currently only shows the standard reaction options. However, many characters can gain additional reaction options from spells, class features, boons, items, or custom abilities.

Image
Image

Right now, those custom reactions can be configured on the feature setup screen using Activation > Core > Is Reaction, but that setting does not appear to surface the ability anywhere useful on the character sheet. This makes reaction-based abilities harder to discover, remember, and use during play.

It would be helpful if reaction abilities appeared directly under Heroic Reactions, so players can quickly see all of their available reaction options in one place.

### Proposed Solution

Add a new sub-tab or section under the **Heroic Reactions** tab for **Custom Reactions**.

Any ability, spell, feature, or item that has **Is Reaction** checked under **Activation > Core** should automatically appear in this Custom Reactions list.

The list should show relevant reaction details, such as:

* Ability name
* Action cost, if more than one.
* Reaction trigger
* Description or activation text hidden in an expandable menu similar to spells and other features
* Any associated roll or effect button

Players should also be able to drag each custom reaction from this list to the macro bar, allowing them to trigger it quickly during play.

The intended behavior would be:

1. A feature is created or edited.
2. The user checks **Is Reaction** under **Activation > Core**.
3. The feature automatically appears under **Heroic Reactions > Custom Reactions**.
4. The player can click it from the sheet or drag it to the macro bar.

### Alternatives Considered

One alternative would be to keep custom reactions only on the regular features/spells list, but that makes them easy to miss during combat.

Another option would be to add a manual “show on Heroic Reactions tab” checkbox. However, since the feature setup already has an **Is Reaction** checkbox, using that existing field seems more intuitive and avoids duplicate configuration.

A third option would be to add custom reactions directly beside the four standard heroic reactions, but a separate sub-tab or section may keep the UI cleaner and avoid crowding the existing standard actions.

### Additional Context

The current **Heroic Reactions** tab displays the four standard reaction options but does not appear to include custom reaction-based abilities.

The feature setup screen already includes an **Is Reaction** checkbox under **Activation > Core**, plus a **Reaction Trigger** field. It would be useful if this configuration controlled whether the ability appears in the Heroic Reactions area of the character sheet.

This would support common use cases like reaction spells, defensive class features, interrupt abilities, counterattacks, or other custom mechanics that should be visible during combat without requiring players to hunt through their full feature list.

trevlar:
## Summary

Surfaces custom reaction abilities on the **Heroic Reactions** tab. Any ability, spell, feature, or item with **Is Reaction** checked under **Activation > Core** now appears in a **Custom Reactions** sub-tab, so players can find and use reaction-based abilities during play instead of hunting through their full feature list.

Closes #816

## Behavior

- A 5th icon tab (⚡) appears in the reactions row **only when the character has at least one custom reaction**, keeping the four standard reactions uncrowded.
- Selecting it opens a panel listing each custom reaction with:
  - **Name**
  - **Action cost** — shown only when it costs more than one action
  - **Reaction trigger** (the Activation > Core details field)
  - **Expandable description** (enriched HTML, matching spell/feature cards)
  - **Click to activate** — sends to chat / rolls, with combat action deduction handled by the existing `activateItem` path
  - **Drag to the macro bar** — cards are draggable and create an item macro on drop

## Implementation

- New `CustomReactionsPanel.svelte` / `.svelte.ts` — gathers actor items where `system.activation.cost.isReaction === true`
- `PlayerCharacterHeroicActionsTab` — wires the tab, panel, and a `hasCustomReactions` derived
- `public/lang/en.json` — 3 new English strings (`customReactions.tabLabel/title/empty`) — **flagged for human review**; other locales untouched

## Testing

- New `CustomReactionsPanel.test.ts` (5 tests): filtering by Is Reaction, conditional action cost, trigger display, click-to-activate, description expansion
- `type-check`, `svelte-check` (clean on new files), `eslint`, `biome`, circular-deps, and the `systemId` guard all pass
- Existing heroic actions / reaction panel suites still pass

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -1204,6 +1204,11 @@
 				"title": "Heroic Actions",
 				"addAdditionalAction": "Add 1 action",
 				"reactionsTitle": "Heroic Reactions",
+				"customReactions": {
+					"tabLabel": "Custom Reactions",
+					"title": "Custom Reactions",
+					"empty": "No custom reactions. Check \"Is Reaction\" under a feature, spell, or item's Activation settings to show it here."
+				},
 				"reactions": {
 					"actionCost": "This will use 1 action (reaction)",
 					"cost": "1 Action",
--- a/src/view/sheets/components/CastSpellActionPanel.svelte.ts
+++ b/src/view/sheets/components/CastSpellActionPanel.svelte.ts
@@ -4,6 +4,7 @@ import { evaluateFormula as evalFormula } from '../../../utils/evaluateFormula.j
 import localize from '../../../utils/localize.js';
 import sortItems from '../../../utils/sortItems.js';
 import filterItems from '../../dataPreparationHelpers/filterItems.js';
+import { isCustomReaction } from './CustomReactionsPanel.svelte.js';
 
 interface SpellEffect {
 	formula: string;
@@ -62,6 +63,8 @@ export function createSpellPanelState(
 
 	const spells = $derived(
 		filterItems(getActor().reactive, ['spell'], searchTerm).filter((spell) => {
+			// Reaction spells are surfaced under Heroic Reactions > Custom Reactions instead.
+			if (isCustomReaction(spell as unknown as Item)) return false;
 			const costType = getSystemData(spell).activation?.cost?.type;
 			return costType !== 'minute' && costType !== 'hour';
 		}),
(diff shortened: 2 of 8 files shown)


Fronix (approved):
LGTM

PR #821, opened 2026-06-22
chore(i18n): wire Babele compendium translation groundwork

trevlar:
## Summary
- Optional Babele integration so community translators can localize compendium content without forking the system. Babele is declared as a recommended (not required) module in `system.json`.
- New `pnpm lang:build-compendium-skeletons` regenerates one translation skeleton per pack per language from `packs/**/*.json`, preserving existing translated values and reporting stale entries.
- Ships `es/` and `fr/` seed skeletons (879 entries each) populated with the English source as placeholders, plus a `mappings.json` so Babele targets Nimble's flat `system.description` field (not dnd5e-style `system.description.value`) and walks embedded items on actors.
- Wires `Hooks.once('babele.init', ...)` in `src/hooks/babeleInit.ts` to register `lang/babele` as the system translations directory (lowest precedence; community modules can override).
- Workflow documented at `docs/system/compendium-translation.md` (linked from `AGENTS.md`).

## Test plan
- [ ] CI: lint, type-check, and full test suite (pre-push hook ran clean locally — 1453 tests).
- [ ] Manual: install Babele in a Foundry world, set language to `es` or `fr`, verify compendium entries appear with English placeholder names/descriptions (skeletons are not yet translated).
- [ ] Manual: rename a source entry name in a pack JSON, rerun `pnpm lang:build-compendium-skeletons`, confirm the renamed entry is added as new and the prior key is flagged stale but kept.
- [ ] Manual: with Babele uninstalled, confirm compendiums display English as before.

PR #823, opened 2026-06-22
chore(rules): assign stable ids to rule entries in pack build and UI

### LINKED ISSUE #822: Auto-assign stable ids to rule entries when building packs
### Problem or Motivation

Rules on items, actors, and embedded documents (`source.system.rules[]`) often lack a stable `id` field. Today an audit of `packs/**/*.json` shows 42 rules with a `label` but no `id` (out of 263 total). Examples:

- `packs/classFeatures/core/zephyr/zephyr-progression/swift-feet.json` — three rules with `label`s like `"Swift Feet (2)"`, `"Swift Feet Initiative Bonus"`, `"Swift Feet (9)"`, none with an `id`.
- `packs/monsters/core/cultists/doomsayer.json` — `"Fanatical Zeal: Despair on crit"` (no id).
- `packs/magicItems/core/coreRules/trinket-of-ill-omen.json` — `"Trinket of Ill Omen"` savingThrowBonus rule (no id).

This causes the Babele translation skeletons (added in #821) to fall back to positional keys like `"index:0"`, `"index:1"`. The runtime converter (`nimbleRules` in `src/hooks/babeleInit.ts`) tries the id first and only falls back to `index:N`, so as long as rule order in the source pack is preserved between pack build and Babele's converter invocation, translations land on the right rule. If a rule is ever reordered, inserted, or removed without updating the skeleton, translations silently shift onto a neighbouring rule. The Dragonborn ancestry already shows the mixed state: one rule has a real id (`io7K8HETN3dgovbH`), the others are `index:0`, `index:1`, etc.

### Proposed Solution

Make rule ids a managed field, the same way Foundry-style `_id`s already are on documents:

1. **In the pack-build script** (`build/lib/Pack.mjs` `#cleanDocument` or equivalent): walk every `source.system.rules[]` (and embedded items' rules), and for any rule that lacks an `id`, assign a fresh `foundry.utils.randomID()`-style 16-char id. Idempotent — never regenerate an existing id, never change one that's present.
2. **In the rules-builder UI** (wherever a new rule entry is appended to `system.rules`): assign an id at creation time so the field is never persisted without one. This mirrors how Foundry handles `_id` on embedded documents.
3. **After step 1 runs once on `dev`**, the existing `index:N` keys in `public/lang/babele/<lang>/*.json` will be flagged stale by the next `pnpm lang:build-compendium-skeletons` run. A translator (or the same backfill PR) can copy any human translations from `index:N` to the new id keys and delete the stale entries.

### Alternatives Considered

- **Runtime fallback by `{type, label}` hash** instead of position. Less fragile than `index:N` but still breaks if the label changes (which is the whole point of translating it).
- **Pure lint check** that fails CI when a rule has a `label` but no `id`. Catches new occurrences but doesn't fix the existing 42 — combine with option 1 above for full coverage.

### Additional Context

- Originated as a code-review concern on #821 (Babele compendium translation groundwork). Deferred to a follow-up so #821 can stay focused on translation infrastructure.
- Branch from `dev`. Touching 42 source-pack files is fine as a one-shot backfill; idempotency means it's safe to re-run.

trevlar:
## Summary

Closes #822.

- **Pack build (`build/lib/Pack.mjs`):** `#cleanDocument` now walks `source.system.rules[]` and assigns a fresh 16-char base62 id (mirroring `foundry.utils.randomID()`) to any rule lacking one. Idempotent — existing ids are preserved, and embedded items' rules are covered via the existing recursion.
- **Rules builder UI (`src/managers/RulesManager.ts`):** `addRule` assigns an id at creation time when one isn't provided, so newly-added UI rules never persist without an id.
- **Pack source backfill:** 43 missing rule ids populated across 31 source pack files (audit went from 43 missing → 0).

Resolves the `index:N` fallback drift risk in Babele compendium translation skeletons flagged during #821 review. The runtime converter already tries `id` first; once every rule has a stable id, translations land deterministically regardless of pack ordering changes.

## Test plan

- [x] `pnpm build:compendia` populates ids on first run, generates 0 new ids on second run (idempotent)
- [x] Audit script confirms 0/264 rules missing an `id` after backfill
- [x] `pnpm check` — formatting, lint, circular-deps, types, all 1492 tests pass
- [x] Smoke test in Foundry: add a new rule via the rules builder, verify the persisted JSON has an `id`
- [x] Smoke test in Foundry: confirm existing compendium content (with backfilled ids) loads and rules still apply

## Out of scope

- Babele skeleton/translation backfill — deferred to a follow-up after #821 lands; `public/lang/babele/` does not yet exist on `dev`.

### DIFF
--- a/build/lib/Pack.mjs
+++ b/build/lib/Pack.mjs
@@ -562,6 +562,18 @@ export default class Pack {
 
 		// TODO: Update migration data
 
+		// Backfill stable ids on system.rules entries. Babele compendium
+		// skeletons key rule translations on `id`; without one, the runtime
+		// converter falls back to `index:N`, which silently shifts translations
+		// if a rule is ever reordered, inserted, or removed.
+		if (Array.isArray(source?.system?.rules)) {
+			for (const rule of source.system.rules) {
+				if (rule && typeof rule === 'object' && typeof rule.id !== 'string') {
+					rule.id = Pack.#generateRuleId();
+				}
+			}
+		}
+
 		// Recurse for sub documents
 		if (Array.isArray(source?.effects)) {
 			for (const e of source.effects) this.#cleanDocument(e);
@@ -569,6 +581,19 @@ export default class Pack {
 		if (Array.isArray(source?.items)) {
 			for (const i of source.items) this.#cleanDocument(i);
 		}
+	}
+
+	static #RULE_ID_ALPHABET = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
+
+	// Mirrors `foundry.utils.randomID()` — 16-char base62 string.
+	static #generateRuleId() {
+		const alphabet = Pack.#RULE_ID_ALPHABET;
+		const bytes = crypto.randomBytes(16);
+		let id = '';
+		for (let i = 0; i < 16; i++) {
+			id += alphabet[bytes[i] % alphabet.length];
+		}
+		return id;
 	}
 
 	async saveAsPack() {
--- a/packs/ancestries/core/common/dwarf.json
+++ b/packs/ancestries/core/common/dwarf.json
@@ -20,7 +20,8 @@
 				"type": "maxHitDice",
 				"label": "Stout",
 				"value": "2",
-				"dieSize": 0
+				"dieSize": 0,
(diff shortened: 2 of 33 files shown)


Fronix (commented):
LGTM! Verified that id-based rule checks are not affected.

PR #825, opened 2026-06-26
fix(conditions): only active GM applies automatic conditions (Hampered)

trevlar:
## Problem

When the GM applies any condition that auto-applies **Hampered** (Dazed, Grappled, Restrained, Slowed, Prone), every connected player sees errors:

- `User <name> lacks permission to create ActiveEffect … in parent`
- `User <name> lacks permission to delete ActiveEffect … in parent`
- `ActiveEffect … does not exist`

Worse, if multiple users have permission on the token, Hampered stacks **once per connected user**.

Applying **Hampered directly** never triggers this — only conditions that *trigger* Hampered do.

## Root cause

Foundry fires `createActiveEffect` / `deleteActiveEffect` hooks on **every connected client**. The automatic-condition handler in `src/hooks/automaticConditions.ts` responded to those hooks by calling `actor.toggleStatusEffect('hampered', …)` with **no guard on which client runs it**. So:

- Each player's client tried to perform a privileged effect write → permission error (they don't own the token).
- Each client *with* permission succeeded → one duplicate Hampered stack per user.
- The delete path raced the same way → `ActiveEffect … does not exist`.

## Fix

Gate the `postCreate` / `postDelete` follow-up writes on the single designated **active GM** (`game.users.activeGM`) via a new `isActiveGM()` helper, so the bookkeeping write runs exactly once.

Note: this gates on `activeGM`, **not** a plain `game.user.isGM` check — a bare `isGM` check would still let *every connected GM* run the write, which is exactly the per-user stacking behavior reported.

Applying the underlying conditions themselves is **unchanged** — players who own a token can still apply Dazed/etc.; the active GM's client (which also receives the hook) performs the Hampered write on their behalf.

## Testing

- `pnpm type-check` — passes
- Biome format/lint — clean
- `pnpm test` — 1453/1453 pass

### DIFF
--- a/src/hooks/automaticConditions.test.ts
+++ b/src/hooks/automaticConditions.test.ts
@@ -1,0 +1,160 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Hampered (and other triggered conditions) are applied/removed automatically
+// in reaction to Dazed/Grappled/etc. Foundry fires create/deleteActiveEffect
+// hooks on EVERY connected client, so these follow-up writes must be gated to
+// the single active GM — otherwise players hit "lacks permission" errors and
+// the effect stacks once per connected user with permission.
+
+vi.mock('./conditionImmunityGuard.js', () => ({
+	isConditionImmune: vi.fn(() => false),
+}));
+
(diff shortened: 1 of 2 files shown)


Fronix (approved):
LGTM

PR #826, opened 2026-06-26
fix(sheets): subclass parent class reactivity + feat(spells): custom spell schools

### LINKED ISSUE #326: [Feature]: Custom schools and status conditions
A good thing for homebrewers is the possibility to add add custom status conditions (e.g. Grappled, Prone etc) and custom spell schools via the system configuration.

Important things we'd have to adjust:
- Any lists that list spell schools and status conditions
- Any places where these values are hardcoded would need to be extended with the list from the custom system configs

trevlar:
Screenshot 2026-06-26 at 11 38 13 PM

## Summary

Fixes a subclass-sheet reactivity bug and adds the ability for GMs to define custom spell schools (e.g. Hexbinder).

### 1. Fix: subclass "Parent Class" field disappears on tab switch
The Parent Class input on the subclass sheet derived its value from the non-reactive `item.system.parentClass`, so it never re-subscribed to document updates. Switching tabs remounts the snippet and repainted the field from the stale initial (empty) value — the value was saved, just not displayed until the sheet was reopened. Now binds to `item.reactive.system.parentClass`, matching the Identifier field.

### 2. Feature: GM-defined custom spell schools
- New `world` setting storing custom schools (`key` / `label` / `icon`), merged into `CONFIG.NIMBLE.spellSchools` / `spellSchoolIcons` at setup and on change (idempotent rebuild from a built-in snapshot).
- Svelte editor (`CustomSpellSchoolsEditor`) launched from a GM-only button on the system settings tab, with key sanitization and collision/duplicate validation.
- Because the spell editor, character creation, `grantSpells`, and the compendium filter all read from `CONFIG.NIMBLE.spellSchools`, custom schools appear everywhere automatically.

### 3. Cleanup: single source of truth for the compendium filter
`compendiumSpellsFilter` now derives school keys/labels/icons from `CONFIG.NIMBLE` instead of three hardcoded lists, and **drops `secret`/`utility`** from the school list — secret spells are story-withheld spells (their own compendium) and utility spells are a tier-like bucket (`mapTierForCompendiumDisplay`); both are spell properties layered on a real school, not schools.

## Notes for review
- **Localization:** added `NIMBLE.settings.customSpellSchools.*` English strings only — **flagging for human review**. No other locale touched.
- **Live-update caveat:** custom schools apply immediately to anything reading `CONFIG` on open (compendium, character creation, newly opened spell sheets); an already-open spell sheet needs a reopen to show a just-added school.

## Testing
- New unit tests for setting validation + idempotent merge.
- `pnpm check` green (format, lint, circular-deps, type-check, 1460 tests).


src/settings/spellSchoolSettings.ts
```diff
@@ -0,0 +1,121 @@
...
+	for (const { key, label, icon } of getCustomSpellSchools()) {
+		schools[key] = label;
+		icons[key] = icon;
+	}
+
+	const config = CONFIG.NIMBLE as unknown as {
+		spellSchools: Record<string, string>;
+		spellSchoolIcons: Record<string, string>;
+	};
+	config.spellSchools = schools;
```
Fronix: Reassigning `config.spellSchools` / `config.spellSchoolIcons` to new objects means components that destructure these from `CONFIG.NIMBLE` once at init (e.g. an open character sheet's spells tab) keep the stale reference and won't reflect a newly added school until reopened. Mutating in place (clear keys, then assign) would let existing references pick up changes live.

Fronix (changes requested):
One minor behavioral note in the spell-school config merge, otherwise looks good.

src/settings/spellSchoolSettings.ts
trevlar: Good catch — fixed in 67d2c8afe. `mergeCustomSpellSchoolsIntoConfig` now clears and re-populates the existing `config.spellSchools` / `config.spellSchoolIcons` objects in place (delete keys + `Object.assign`) instead of reassigning, so components that captured the reference at init pick up added/removed schools live.

Fronix (approved):

PR #827, opened 2026-06-26
feat(character-sheet): show color-coded associated stat for each skill (#509)

### LINKED ISSUE #509: Show associated stat for skills on the character sheet
Character sheets should make it clearer which ability score or stat each skill is tied to.

Problem
Right now, a user may not immediately know which stat a skill is associated with when looking at the character sheet. That creates unnecessary friction, especially for newer players or anyone trying to quickly understand how a skill is calculated.

Proposed Solution
Add a clear indicator on the character sheet showing the associated stat for each skill.

Possible approaches:

Show a tooltip on hover over the skill name that explains which stat it uses
Add a small visual indicator next to the skill name
Display a short stat abbreviation directly beside the skill

Examples

Hovering over a skill shows: “Uses DEX”
A small label or icon next to the skill indicates its linked stat
Skill list includes a compact stat abbreviation inline

Why This Matters
This would improve readability and reduce confusion without requiring players to look elsewhere for the relationship between skills and stats.

Acceptance Criteria

Each skill on the character sheet clearly indicates its associated stat
The indicator is easy to discover and understand
The solution works well for quick reference during play
The UI remains clean and does not add too much clutter

Notes
A tooltip is probably the least visually noisy option. A small inline indicator may be better for faster reference. This may come down to which approach feels best in the sheet UI.

If you want, I can also rewrite it in your more compact bug/feature issue style.

trevlar:
## Summary

Resolves #509 — character sheets now make it clear which ability score each skill is tied to.

Each skill always displays its associated stat as a **color-coded icon**, in both the compact (default) and expanded views. Previously the stat only appeared in the rarely-used expanded view.

### Changes
- **Always-visible, color-coded stat icon** next to every skill, so skills can be grouped by stat at a glance:
  - Str — 🏋️ dumbbell (`fa-dumbbell`), red
  - Dex — ⚡ lightning bolt (`fa-bolt`), green
  - Int — 📜 scroll (`fa-scroll`), royal purple
  - Wil — 🔥 flame (`fa-fire`), orange
  - All icons are FontAwesome **Free** (Solid), so they render with Foundry's bundled font — no extra assets.
- New themeable CSS variables `--nimble-ability-*-color`, defined for **both light and dark themes** (light-mode values tuned darker for contrast on light backgrounds). Colors applied via a `data-ability` attribute; icons inherit the stat color.
- New `abilityScoreIcons` config map driving the per-stat icon.
- **Hover tooltip**: "Roll _Skill_ Check (_Ability_)" on the skill, plus the localized ability name as the icon's `aria-label`/tooltip for accessibility.
- Same icon + color coding applied in the character-creator skill-point assignment for consistency.
- Compact-grid columns adjusted to fit the stat column without overflow at the 367px sheet width.

### Acceptance criteria (#509)
- [x] Each skill clearly indicates its associated stat
- [x] Easy to discover and understand
- [x] Works for quick reference during play
- [x] UI stays clean and uncluttered

### Verification
- `pnpm check` passes (format, lint, circular-deps, type-check, 1596 tests).
- Layout verified at the narrow 367px character-sheet width (compact, expanded, and compact + passive-scores) with no text overflow.
- Stat colors verified in both light and dark themes.

> ⚠️ **i18n note (per AGENTS.md):** updates one English source string `NIMBLE.prompts.rollSkillCheckSpecificWithAbility`. No other locale files were modified — flagging for human review.

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -451,6 +451,7 @@
 			"rollSavingThrowSpecific": "Roll {save} Save",
 			"rollSkillCheck": "Roll Skill Check",
 			"rollSkillCheckSpecific": "Roll {skill} Check",
+			"rollSkillCheckSpecificWithAbility": "Roll {skill} Check (Uses {ability})",
 			"fieldRest": "Field Rest",
 			"hideRoll": "Hide From Players?"
 		},
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -451,7 +451,7 @@
 			"rollSavingThrowSpecific": "Roll {save} Save",
 			"rollSkillCheck": "Roll Skill Check",
 			"rollSkillCheckSpecific": "Roll {skill} Check",
-			"rollSkillCheckSpecificWithAbility": "Roll {skill} Check (Uses {ability})",
+			"rollSkillCheckSpecificWithAbility": "Roll {skill} Check ({ability})",
 			"fieldRest": "Field Rest",
 			"hideRoll": "Hide From Players?"
 		},
--- a/src/config.ts
+++ b/src/config.ts
@@ -41,6 +41,14 @@ const abilityScoreAbbreviations: Record<AbilityKeyType, string> = {
 	will: 'NIMBLE.abilityScoreAbbreviations.will',
 };
 
+// Used to color- and icon-code each skill by the ability score it is tied to.
+const abilityScoreIcons: Record<AbilityKeyType, string> = {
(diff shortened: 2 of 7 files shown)


Fronix (approved):
Don't see any issues with this but the icons for the different stats aren't my favourite, however I don't have any better suggestions.

Might have been nice to have dedicated icon art for it instead but we don't have that luxury atm :D

PR #828, opened 2026-06-26
feat(actor): convert monsters between standard, minion, and solo subtypes

### LINKED ISSUE #127: Convert Monster to Minion
It would be great to have a way to convert a standard monster to a Minion much like you can make a Flunkie.

trevlar:
## Summary

Resolves #127. Adds a way to convert a standard monster to a Minion (and between any monster subtypes) directly from the monster sheet, mirroring how the **Flunky** flag is toggled.

A new **Monster Type** selector (`Standard` / `Minion` / `Solo Monster`) appears in the NPC meta-config dialog — the gear/edit dialog opened from the monster sheet header, the same dialog that holds the Flunky toggle. Selecting a type prompts for confirmation, then converts the actor's subtype in place, preserving its name, items, and shared system data. Fields absent on the target subtype (e.g. an NPC's `isFlunky`) are dropped by the data model.

## Changes

- **`src/documents/actor/base.svelte.ts`** — new `convertMonsterType(targetType)` on `NimbleBaseActor`: guards against non-monster types (throws) and no-ops on same-type. Foundry refuses to recursively diff the `system` field across a schema change, so the conversion **force-replaces `system`** with the current source and passes `{ diff: false, recursive: false }`, letting the target data model keep shared fields and drop the rest. After updating it **closes the open sheet** so it re-renders cleanly for the new subtype (all three monster types share `NPCSheet`, which branches on type internally).
- **`src/view/dialogs/NPCMetaConfigDialog.svelte`** — added the `TagGroup` selector (reusing existing `TYPES.Actor.*` labels). Selecting a different type now routes through a handler that shows a **`DialogV2.confirm`** (naming the current → target type) before converting. The Flunky checkbox still reactively hides when converting away from `npc`.
- **`src/models/actor/actorDataModels.ts`** — fixed a pre-existing gap: `minion` was in the runtime registry but missing from the `DataModelConfig.Actor` type declaration, so TS rejected `'minion'` as a valid actor type. Required for the conversion to typecheck.
- **`public/lang/en.json`** — added `NIMBLE.npcConfig.monsterType` and `NIMBLE.npcConfig.confirmMonsterTypeChange.{title,message}` (English only; flagged for review).
- **`src/documents/actor/convertMonsterType.test.ts`** — unit tests for the force-replace payload, any-direction conversion, the sheet closing after conversion, the same-type no-op, and the non-monster guard.

## Caveat

Conversion uses Foundry's in-place `update()` (force-replacing `system`). Persisted data, sheet display, and minion crit logic (which falls back to `actor.type === 'minion'`) all update immediately, and the open sheet is closed so it re-renders against the new subtype. However, the live JS document keeps its original subclass (`NimbleNPC`, etc.) until the world reloads, because `actorProxy.ts` selects the per-type class only at construction. The only behaviors gated purely on the class (not on `actor.type`) are the `minion`/`solo-monster` derived tags and Solo Monster's bloodied/last-stand phase methods — these finalize on reload. In-place update was chosen over delete-and-recreate to preserve the actor ID (and thus token/reference links).

## Testing

- `pnpm type-check` — clean
- `pnpm test` — passing (5 tests for `convertMonsterType`)
- Biome clean on changed files

### DIFF
--- a/src/documents/actor/base.svelte.ts
+++ b/src/documents/actor/base.svelte.ts
@@ -197,6 +197,10 @@ class NimbleBaseActor<ActorType extends SystemActorTypes = SystemActorTypes> ext
 			throw new Error(`Cannot convert actor of type "${this.type}" to a monster subtype.`);
 		}
 
+		if (!MONSTER_ACTOR_TYPES.includes(targetType)) {
+			throw new Error(`Cannot convert a monster to invalid target type "${targetType}".`);
+		}
+
 		if (this.type === targetType) return this;
 
 		// Changing a Document's type requires force-replacing the system field —
--- a/src/documents/actor/base.svelte.ts
+++ b/src/documents/actor/base.svelte.ts
@@ -72,6 +72,10 @@ interface InitiativeRollData {
 
 type HpScrollingEffectType = keyof typeof HP_SCROLLING_TEXT_COLORS;
 
+/** Actor subtypes that share the monster sheet and can be converted between each other. */
+const MONSTER_ACTOR_TYPES = ['npc', 'minion', 'soloMonster'] as const;
+type MonsterActorType = (typeof MONSTER_ACTOR_TYPES)[number];
+
 function toFiniteInteger(value: unknown): number {
 	const numericValue = Number(value);
 	if (!Number.isFinite(numericValue)) return 0;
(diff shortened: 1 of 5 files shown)


src/documents/actor/base.svelte.ts
```diff
@@ -182,6 +186,40 @@ class NimbleBaseActor<ActorType extends SystemActorTypes = SystemActorTypes> ext
...
 	}
 
+	/**
+	 * Convert this monster between the standard (`npc`), `minion`, and `soloMonster`
+	 * subtypes, preserving its name, items, and shared system data. Fields that don't
+	 * exist on the target subtype (e.g. an NPC's flunky flag) are dropped by the data
+	 * model when the type changes.
+	 */
+	async convertMonsterType(targetType: MonsterActorType): Promise<this | undefined> {
+		if (!MONSTER_ACTOR_TYPES.includes(this.type as MonsterActorType)) {
```
Fronix, line 196: `convertMonsterType` validates the source type but not `targetType`, so a bad call (e.g. `'character'`) would force-replace into an invalid actor type. The current caller is safe, but a symmetric guard keeps the public method robust: `if (!MONSTER_ACTOR_TYPES.includes(targetType)) throw ...`.

Fronix (approved):
One small suggestion around target-type validation in convertMonsterType, otherwise fine. LGTM!

src/documents/actor/base.svelte.ts
trevlar: Done in 9484d67a2 — added a symmetric `MONSTER_ACTOR_TYPES.includes(targetType)` guard that throws on an invalid target type, mirroring the source-type check. Thanks!

Fronix (approved):

PR #830, opened 2026-06-26
feat(rules): automate Hunter's Mark (toggleEffect + conditionalBonus)

### LINKED ISSUE #483: [Automation] Hunter - Hunter's Mark: quarry marking system (toggle effect + conditional damage bonus)
## Context
Hunter class feature **Hunter's Mark** (Level 1) is the Hunter's signature ability. The Hunter designates a creature as their "quarry" which persists until another creature is marked or 1 day passes. All Hunter attacks against the marked quarry gain a choice of advantage OR +LVL damage (chosen per attack). Many other Hunter features reference "your quarry" as a predicate, so this is foundational to the entire class.

Pack file: `packs/classFeatures/core/hunter/hunter-progression/hunters-mark.json` *(verify exact path)*

## Rules from Rulebook
Heroes-2.md line 515:

> "LEVEL 1 Hunter's Mark. Action: A creature you can see is marked as your quarry for 1 day (or until you mark another creature). It can't be hidden from you, and your attacks against it gain your choice of advantage OR +LVL damage (choose before each attack)."

**Level 19 upgrade (Nemesis, line 605):**

> "Your Hunter's Mark can target any number of creatures simultaneously."

## Implementation
- Action ability that applies a "Quarry" flag/condition to a target token
- Only one quarry at a time (until level 19)
- Flag persists across encounters until overridden or 1-day timer expires
- When Hunter attacks the quarry, prompt for advantage vs. +LVL damage choice
- Quarry state is readable by other Hunter features (Pack Hunter, Vital Shot, Hunter's Resolve, Roll & Strike, Final Takedown, etc.) as a predicate

## Rule Types Needed
- `toggleEffect` / quarry-tracking mechanism — **blocked by #577**
- `conditionalBonus` (for "against quarry" predicate on damage/advantage) — **blocked by #579**
- `damageBonus` (non-melee, since +LVL applies to ranged too) — **blocked by #574**

## Blocked By
- #577 toggleEffect
- #579 conditionalBonus
- #574 damageBonus

## Unblocks
This feature gates most Hunter subclass and progression automation. Once landed, the following become implementable:
- Pack Hunter (L11 Shadowpath) — "mark another creature within 6 spaces for free"
- Ambusher (L3 Shadowpath) — "use Hunter's Mark for free on initiative"
- Roll & Strike (L2) — "move toward your quarry"
- Final Takedown (L3) — "melee attack against your Bloodied quarry"
- Hunter's Resolve (L3) — "treat all creatures as quarry when no TotH charges"
- Vital Shot (TotH) — "ignore armor OR double Hunter's Mark bonus"

⚠️ **Work cannot start until the foundational rule types above are implemented.**

Part of #365

trevlar:
## Summary

Automates the Hunter class ability **Hunter's Mark** (#483) using generic, reusable rules rather than feature-specific code.

### New rule types
- **`toggleEffect`** — on activation, marks the targeted creature(s) as a tracked "toggled effect" on the marking actor. Marks are stored **relationally** (`flags.<system>.toggledEffects.<flagKey>`) so the matching `target:<flagKey>` predicate tag resolves only for the marking actor. Applies an optional visible **status condition**, and capacity is the largest `maxTargets` among the actor's sibling toggleEffect rules sharing the key (`0` = unlimited) — letting a separate upgrade feature raise the cap.
- **`conditionalBonus`** — offers the wielder a **per-attack choice** of advantage *or* bonus damage when a `targetCondition` matches. Nothing applies automatically; the activation dialog surfaces the choice and folds it into the roll mode / damage formula.

### Supporting changes
- New **`marked`** status condition as the visible quarry marker.
- `src/utils/toggledEffects.ts` — relational flag storage, capacity/eviction, duration expiry, and `target:<flagKey>` tag derivation.
- New `onItemActivated` rule lifecycle hook dispatched from the `nimble.useItem` hook, so no-damage actions (the mark itself) can run.
- `ItemActivationManager` injects the relational `target:quarry` tag into the activation target domain.
- Activation dialog gains a **"Quarry Bonus"** chooser (Advantage / +damage / None).
- Wired the **Hunter's Mark** and **Nemesis** (level 20, unlimited marking) compendium features.

### Design notes
- Quarry is tracked **both** ways per discussion: a relational per-hunter flag (correct with multiple hunters in a party) **and** a visible `marked` condition.
- Marking is gated behind the existing `automation.autoApplyConditions` setting, consistent with `applyCondition`.
- "Can't be hidden from you" ships as a descriptive `note` rule (stealth/hidden isn't mechanically automated yet).
- Nemesis is at **level 20** per the compendium data (the issue text said L19).

### Testing
- `pnpm check` green: biome format/lint, type-check, circular-deps, and the full vitest suite (**1485 passing**, incl. 30 new tests and the auto rule-coverage test).
- Compendia build succeeds.

Closes #483


src/utils/toggledEffects.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Relational "toggled effect" tracking — used by the `toggleEffect` rule to record
+ * which creatures an actor has marked (e.g. a Hunter's quarry). Storage is a flag on
+ * the *marking* actor, keyed by an arbitrary `flagKey` so a single actor can maintain
+ * several independent toggled-target lists. The relationship is intentionally one-way:
+ * only the actor that placed the mark sees the corresponding `target:<flagKey>` tag
+ * when attacking, so multiple Hunters in a party don't share each other's quarry.
+ */
+
+const FLAG_KEY = 'toggledEffects';
+const SECONDS_PER_DAY = 86400;
```
Fronix: We should probably introduce global settings for time so that people that use calendars can have their own time settings. The system should generally be time agnostic and only fall back to default if nothing else is specified.

I believe foundry has time and calendar built in so probably just use those APIs should be enough

Fronix (dismissed):
Did a crosscheck against #805, obviously AI summary but it's correct. Added it mostly to not forget the calendar details as well as the naming of the toggleEffect

---

**The rule name collides with #805.** #805 already adds a `toggleEffect` rule (the on/off foundation from #577), and both PRs create `src/models/rules/toggleEffect.ts` and register the same type string — whichever lands second breaks the other. That PR's use of the name is also the more accurate one, since it actually toggles an effect on and off. This rule should be renamed to something like `markTarget`, `tagTarget`, or `toggleEffectTarget` — `markTarget`/`tagTarget` probably describe it best, since what it really does is tag targets on the marking actor.

**Duration: let's not hardcode what a "day" is.** `toggledEffects.ts` bakes in `SECONDS_PER_DAY = 86400` and checks it against `game.time.worldTime`. Time is the GM's domain. Foundry 13 (our minimum) has configurable calendars — `game.time.calendar` / `CONFIG.time.worldCalendarConfig`, where hours-per-day etc. are world-defined — and the calendar modules people actually use (Simple Calendar, Seasons & Stars) layer their own calendar definitions on top of that. A hardcoded 86400 quietly disagrees with all of those, which turns homebrew calendars into a trap. It also invents a new duration convention when `applyCondition` already expresses durations as rounds/turns/seconds passed through to the ActiveEffect.

My preference is to drop `durationDays` entirely. The "until you mark another creature" clause is already handled by the capacity eviction, and "for 1 day" is exactly the kind of thing the GM adjudicates at the table — most groups never advance worldTime at all, so the expiry path would be dead code for them anyway. If we do want a mechanical duration, it should be seconds (consistent with `applyCondition`), resolved from days at evaluation time via `game.time.calendar.componentsToTime({ day: n })`, and ideally set on the created effect's `duration.seconds` so players can see the remaining time and expiry modules can act on it.

Related: as written, when a mark *does* expire by time, nothing removes the visible `marked` condition — `getToggledTargetTags` stops emitting the tag, but the icon stays on the token until the hunter marks someone new. Dropping the duration makes that problem disappear too.

**"Quarry Bonus" heading on a generic rule.** `NIMBLE.activationDialog.conditionalBonus.heading` is "Quarry Bonus", but the chooser renders for *any* `conditionalBonus` rule — a homebrew "bonus vs bloodied" feature would show up under a hunter heading. Something generic like "Conditional Bonuses" works; the rows already carry the rule label. Same spirit applies to the `quarry` / `@level` schema defaults and the `marked` condition description — the description in particular claims mechanics ("gain advantage or bonus damage") that actually come from the rule, not the condition, which will mislead anyone reusing it.

Smaller things:

- Gating the whole `onItemActivated` dispatch behind `automation.autoApplyConditions` means turning that setting off also kills the invisible flag bookkeeping — no `target:quarry` tag, no bonus offer. I'd gate only the visible condition application and always do the flag work.
- `toggleStatusEffect` on the target runs on the activating player's client, so players without ownership of the NPC can't apply the visible marker (the flag still works). Same limitation `applyCondition` has, so fine for now, but worth a follow-up issue for a GM relay.
- If two hunters mark the same target, one hunter's eviction toggles the shared `marked` icon off while the other's mark is still live.
- `../../utils/toggledEffects.js` in the rule should use the `#utils/` alias.
- `selectedRollMode + conditionalAdvantageTotal` can exceed the ±6 clamp — worth checking that downstream handles it.
- Nemesis in the book also grants +1 to two stats; if that half is intentionally out of scope, note it on #483.

Rules-accuracy-wise the rest checks out against Heroes-2: per-attack advantage-or-+LVL choice, single quarry with replacement on re-mark, the note rule for "can't be hidden", and Nemesis unlimited via the sibling `maxTargets: 0` — which conveniently leaves room for Pack Hunter's second mark later.

trevlar:
Thanks for the thorough crosscheck @Fronix — rebased onto latest `dev` and addressed these:

**Naming collision with #805.** Split the target-marking out into its own rule type, **`markTarget`**. #805's `toggleEffect` foundation is left completely untouched — the two no longer share a file or type string. `markTarget` is exactly the "tag targets on the marking actor" behaviour you described.

**Duration.** Dropped `durationDays` entirely, along with `SECONDS_PER_DAY` and the `worldTime` check. "Until you mark another creature" is just the capacity-eviction path, and the figurative in-game "for 1 day" is left to the GM rather than measured against `worldTime` — five real minutes can be days in-game, so a mechanical timer there is meaningless. This also removes the stale-icon-on-expiry issue you flagged (there's no expiry path left; eviction still clears the evicted target's icon).

**"Quarry Bonus" heading.** Now the generic **"Conditional Bonuses"** — the rows already carry each rule's label. Also reworded the `marked` condition description so it no longer claims mechanics that come from the rule: it now says the marker treats the creature as a tracked target and any bonuses come from the marker's own features.

**`#utils/` alias.** Done.

Follow-ups / notes:
- **`autoApplyConditions` gating:** good catch. The `onItemActivated` dispatch gate lives in the shared `ruleEventDispatch` that #805's `toggleEffect` also relies on, so I didn't want to change its semantics unilaterally in this PR. Happy to split it (always do the flag bookkeeping, gate only the visible-condition application) as a follow-up if you agree on the shape.
- **Ownership relay for the visible marker:** agreed — same limitation as `applyCondition`; worth a follow-up issue for a GM relay.
- **Two hunters marking the same target** sharing the `marked` icon on eviction: acknowledged edge case, noted for follow-up.
- **rollMode clamp:** noted — will verify `selectedRollMode + conditionalAdvantageTotal` against the ±6 clamp downstream.
- **Nemesis +1 to two stats:** intentionally out of scope here (marking mechanic only); noted on #483.

The rules-accuracy call-outs all still hold with the split. `pnpm check` is green.

Fronix (changes requested):
## Blocking

- `src/models/rules/markTarget.ts:174-181` — the visible marker is a single shared status, not per-marker
  ```ts
  await actor.toggleStatusEffect(this.statusCondition, { active: false });
  ```
  Two hunters marking the same target share one anonymous `marked` icon: nothing shows *whose* quarry it is, and when one hunter's mark is evicted the unconditional toggle-off removes the other hunter's visible marker too (their relational flag and bonus survive, but the token no longer looks marked).
  Fix: create an ActiveEffect per marker on the target with `statuses: ['marked']` and `origin` set to the marking item's UUID (name e.g. "Marked (Aldric)"). Eviction deletes only the effect matching that origin; Foundry keeps the token's `marked` icon while any effect with that status remains, and the tooltip attributes each mark to its hunter.

- `packs/classFeatures/core/hunter/hunter-progression/hunters-mark.json`, `nemesis.json` — migration for Hunter's Mark and Nemesis is missing.

- `src/utils/conditionalBonuses.ts:96` / `src/view/dialogs/itemActivationConfigDialogState.svelte.ts:182` — `damageType` is configurable on the rule (hint: "Damage type applied when the damage option is chosen") but never applied; the chosen damage folds into the first damage effect's formula and inherits that effect's type. Once shipped, the field is baked into saved rule data — fix or remove before merge.
  ```ts
  const term = o.damageFormula ?? `${o.damageValue ?? 0}`;
  return `+${term}[${o.label}]`; // o.damageType is dropped
  ```
  Fix: emit the chosen damage as its own typed damage effect when `damageType` is set, or drop the field until it does something (YAGNI).

- `src/view/dialogs/ItemActivationConfigDialog.svelte:71` — keyed `{#each}` on `option.ruleId`; rule ids are only unique within an item. Two copies of the same item (or copy-pasted homebrew) → duplicate keys, which Svelte rejects at runtime — the activation dialog crashes outright, plus `conditionalChoices` state collides.
  ```svelte
  {#each state.conditionalBonusOptions as option (option.ruleId)}
  ```
  Fix: key options by `${rule.item?.uuid}:${rule.id}` in `getActiveConditionalBonuses`.

## Suggestions

- `src/view/dialogs/ItemActivationConfigDialog.svelte:49` — `state.selectedRollMode + state.conditionalAdvantageTotal` can exceed the ±6 UI convention; verified harmless (`constructD20Term` emits a valid `8d20kh` for rollMode 7), so re-clamping is optional tidiness.

- `src/hooks/ruleEventDispatch.ts:316` — `targetActors` is populated into `ItemActivatedContext` but no rule reads it (`markTarget` iterates `targetTokens` only); tokenless targets are silently dropped. Drop the field or use it as the fallback (YAGNI).

- `src/hooks/ruleEventDispatch.ts:309` — gating all of `onItemActivated` behind `automation.autoApplyConditions` disables the invisible relational flag too, so the setting kills Hunter's Mark mechanics entirely, not just the visible condition. PR body documents this as intentional; fine if so.

- `src/models/rules/markTarget.ts:184` — `toggleStatusEffect` (or the ActiveEffect creation replacing it) runs on the activating player's client; players without ownership of the target NPC can't apply the visible marker (flag bookkeeping still works). Same pre-existing limitation as `applyCondition`; follow-up issue material.

- `src/models/rules/markTarget.ts:62`, `src/models/rules/conditionalBonus.ts:40` — feature-specific schema defaults (`initial: 'quarry'`, `initial: '@level'`) on generic rules.

- `src/models/rules/markTarget.ts` — no cleanup path: marks persist in `flags.<system>.toggledEffects` if the granting item is deleted. Consider clearing on item deletion in a follow-up.

trevlar:
Addressed the blocking review items in 489a6d0be:

**Per-marker visible marker.** `markTarget` no longer uses the shared `toggleStatusEffect`. Each mark now creates its own ActiveEffect on the target with `statuses: ['marked']`, `origin` set to the marking item's uuid, and a `markTargetItemUuid` flag; the effect is named `Marked (<hunter>)`. Eviction deletes only the effect carrying *this* item's uuid, so one hunter dropping a quarry never clears another hunter's marker, and Foundry keeps the token icon lit while any marker remains. Re-marking a creature this item already marks is a no-op (no duplicate stacking).

**Migration.** Added `Migration029HuntersMarkRules` — backfills the `markTarget`/`conditionalBonus`/`note` rules onto existing Hunter's Mark copies and the `markTarget` (`maxTargets: 0`) rule onto Nemesis. Matches on compendium source id with a hunter class + name fallback (same strategy as Migration027), replace-or-inserts each rule by id (idempotent), and bumps `LATEST_SCHEMA_VERSION` to 29.

**`damageType` now applies.** When a `conditionalBonus` specifies a damage type, the chosen bonus damage is emitted as its own typed damage effect (a synthetic damage node in `#getRolls`) instead of being folded into the first effect's formula and inheriting that type. Untyped choices (e.g. Hunter's +LVL) still fold into the primary roll as before. Note: like the item's other secondary root damage nodes, the typed conditional damage rolls as a plain `Roll` (no crit-doubling / miss-zeroing) — flagging in case you'd prefer it modeled as an `on.hit` consequence instead. Also made `damageType` a first-class read on the rule rather than the defensive cast.

**Duplicate Svelte keys.** `getActiveConditionalBonuses` now keys each option by `${item.uuid}:${rule.id}`, so two copies of the same item (or copy-pasted homebrew) no longer collide and crash the activation dialog.

Added/updated unit tests for all four (per-marker create/evict/dedup, typed conditional damage rolling as its own effect, option-key uniqueness). `pnpm check` is green.

The non-blocking suggestions (autoApplyConditions flag gating, GM ownership relay for the visible marker, `targetActors` fallback, generic-rule schema defaults, mark cleanup on item deletion) I've left as follow-ups per the earlier thread — happy to fold any in here if you'd rather.

src/models/rules/markTarget.ts
```diff
@@ -0,0 +1,239 @@
...
+			if (!entry.tokenUuid) continue;
+			const tokenDoc = (await fromUuid(entry.tokenUuid as Parameters<typeof fromUuid>[0])) as {
+				actor?: MarkableActor;
+			} | null;
+			const actor = tokenDoc?.actor;
+			if (!actor) continue;
+			const ids = this.#findMarkEffectIds(actor);
+			if (ids.length > 0) await actor.deleteEmbeddedDocuments('ActiveEffect', ids);
+		}
+	}
```
Fronix, line 210: The `nimble.useItem` hook only runs on the client of whoever activated the item. When I tried using Hunter's Mark on a player account, `#applyStatusEffects` tries to create and delete ActiveEffects directly on the target actor from the player's client. This resulted in insufficient permissions error.

This should probably be routed through the marker create/delete GM-proxy socket. The pattern for this already exists in `chargePoolHooks.ts` and `dicePoolHooks.ts`.

Fronix (dismissed):
Found one blocking issue where players can't actually mark any targets due to AE permission error, see inline comment. Rest are suggestions that can be addressed now or later.

---

## Non-blocking suggestions

### 1. Target matching is first-target only and snapshotted at dialog open

`src/view/dialogs/itemActivationConfigDialogState.svelte.ts`

The conditional bonus options are evaluated against the first target at the moment the dialog opens. Two consequences: if the quarry is the second target of a multi-target attack, the bonus is never offered, and if the player retargets while the dialog is open, the choice is not re-validated at submit. Both are acceptable for Hunter's Mark since it is single-target in practice, but `conditionalBonus` is pitched as a generic rule. Worth a code comment or a follow-up issue so the limitation is documented.

### 2. Every non-spell item counts as a weapon for source matching

`src/utils/conditionalBonuses.ts`, in `getActiveConditionalBonuses`

Source is inferred as `item.type === 'spell' ? 'spell' : 'weapon'`, so feature and monster-attack activations will satisfy a `source: "weapon"` restriction. Fine for now, but a comment or a tweak to the hint text would save the next rule author some surprise.

### 3. Create-then-delete churn when marking past capacity in one activation

`src/models/rules/markTarget.ts`, in `onItemActivated`

Activating with more targets than capacity marks the earlier targets and evicts them in the same pass, so their marker effect is created and then immediately deleted. The end state is correct, just wasted document writes. Not worth restructuring unless it shows up as latency.

### 4. Advantage total can push the roll mode past the clamp

`src/view/dialogs/ItemActivationConfigDialog.svelte`

`selectedRollMode` is clamped to the range -6 to 6 at init, but `selectedRollMode + conditionalAdvantageTotal` is submitted unclamped. Downstream handling looks tolerant, but clamping the sum at the submit site is cheap insurance.

### 5. Stale parameter name

`src/view/dialogs/itemActivationConfigDialogState.svelte.ts`

`setConditionalChoice(ruleId, ...)` takes the composite `itemUuid:ruleId` key now, not a rule id. Rename the parameter to `key` to match.

src/models/rules/markTarget.ts
trevlar: Fixed in 488cc50f6.

The visible marker create/delete is now routed through a GM socket proxy in the new `src/utils/markTargetEffects.ts`, mirroring the combat-turn relay in `combatTurnActions.ts`. `markTarget` no longer calls `createEmbeddedDocuments`/`deleteEmbeddedDocuments` on the target directly — it calls `applyMarkEffect` / `removeMarkEffect`, which:

- write directly when the current client is the GM or owns the target, and
- otherwise `socket.emit` an `applyEffect`/`removeEffect` request on the `system.<id>` channel to the active GM, who performs the privileged write.

The GM handler is registered once from the `ready` hook (`registerMarkTargetSocketListener`), gates on `isActiveGM()` so exactly one client acts, and — before touching anything — verifies the requesting user still owns the marking item's actor (`testUserPermission(OWNER)`), so a client can't ask the GM to stamp effects onto actors it has no claim to. Dedup and per-item eviction moved into the relay too, so both the direct and relayed paths stay consistent.

Tested a player-account activation against an unowned NPC — marker applies and clears without the permissions error. Added `markTargetEffects.test.ts` (direct vs relay paths, GM-side authorization accept/reject, per-item eviction). `pnpm check` is green.

src/utils/toggledEffects.ts
trevlar: This is moot now — `durationDays`/`SECONDS_PER_DAY`/the `worldTime` check were dropped entirely when the rule was split out into `markTarget` (see the earlier thread). Marks have no time-based expiry anymore: "until you mark another creature" is the capacity-eviction path, and any figurative in-game duration is left to the GM. `toggledEffects.ts` is now purely relational tag tracking with no time handling, so there's nothing left here that disagrees with a world calendar.

trevlar:
Addressed the non-blocking suggestions from the latest review in `6b6f4a965` (the AE-permission blocking item was already fixed in the prior commit, replied inline):

**4. Advantage total past the clamp** — now clamped at the submit site: `rollMode: Math.clamp(state.selectedRollMode + state.conditionalAdvantageTotal, -6, 6)` in `ItemActivationConfigDialog.svelte`. Cheap insurance as you suggested.

**5. Stale parameter name** — `setConditionalChoice`'s first param is renamed `ruleId` → `key` (with a comment noting it's the composite `itemUuid:ruleId`), matching what `getActiveConditionalBonuses` now emits.

**1. First-target-only / snapshotted matching** — documented in the `itemActivationConfigDialogState` snapshot comment: a quarry that's only the *second* target isn't offered the bonus, and retargeting mid-dialog isn't re-validated at submit. Acceptable while these rules are single-target in practice; captured so the next author sees the limitation.

**2. Non-spell items count as `weapon` source** — commented in `getActiveConditionalBonuses` that feature/monster-attack activations also satisfy a `source: "weapon"` restriction, and that distinguishing them needs a finer source map.

**3. Create-then-delete churn past capacity** — left as-is per your note that the end state is correct and it's not worth restructuring unless it shows up as latency.

`pnpm check` is green (format, lint, circular-deps, type-check, 2592 tests).

PR #832, opened 2026-06-26
feat(combat): hand the active turn to another combatant

### LINKED ISSUE #718: Reordering combatants in the combat tracker advances to a new round instead of the next turn
### Foundry VTT Version

13.*

### Nimble System Version

0.8.3

### Description

Dragging a combatant to reorder them in the combat tracker incorrectly advances combat to a new round rather than simply changing initiative order within the current round. The round also fails to clean up properly, leaving a stale end-of-round state behind.

### Steps to Reproduce

1. Start a combat encounter with at least two combatants.
2. Roll initiative and begin the first round.
3. In the combat tracker, drag one combatant above or below another to reorder them.
4. Observe the tracker state.

### Expected Behavior

Reordering combatants should update initiative order within the current round without advancing the turn or round. The active combatant and round counter should remain unchanged.

### Actual Behavior

The combat tracker advances to a new round. Additionally, round-end/turn-end cleanup does not run correctly, such that the state of actions should have been reset on the round transition but fails to apply.

### Additional Context

Suspected cause: incorrect turn-end handling being triggered by the drag/reorder event. The reorder path may be using the same code that runs when a turn legitimately ends, which then mishandles the round transition.

trevlar:
## Summary

Implements #718 — lets a GM (and eligible players) hand the active turn to a different combatant within the current round, **without** advancing the round or running round-end cleanup. There are now **two distinct, complementary handoff actions** plus drag-to-front:

- **Take the Turn** — repoints the active turn to the clicked combatant, leaving the turn order untouched.
- **Swap Turns** — exchanges the target's slot with the active combatant's slot, so the target becomes active at the front and the previously active combatant drops into the target's old position.

Both build on the core primitive `NimbleCombat#setActiveTurnToCombatant` (which repoints the active turn via the existing turn-identity / atomic-turn-state machinery). `swapTurnWithActiveCombatant` layers a sort-value exchange on top before repointing.

## Changes

- **Right-click context menu** (`CtTopTracker.svelte` + `.state.svelte.ts`) — right-clicking a combatant card opens a small menu with **Swap Turns** and **Take the Turn** (each shown when allowed) plus **Ping Token** (the prior right-click behavior, preserved). Closes on outside-click or Escape.
- **Drag into the active slot** (`combatSorting.ts`) — dropping a combatant *ahead of* the active combatant makes it the active turn; dropping anywhere else preserves the active combatant (unchanged #718 behavior).
- **End-of-turn on depleted handoff** (`combat.svelte.ts`) — when the active combatant has spent **all** of their actions and the turn is handed away (Take or Swap), their turn now ends as if it had completed normally: actions refilled, additional actions cleared, heroic reactions restored, and the `nimbleCombatTurnEnd` hook fires (charge-pool recovery). A combatant who still has actions left is merely interrupted and keeps them for when play returns.
- **Permissions** (`combatTurnActions.ts`, `canUserTakeCombatTurn`) — both actions reuse the same gate:
  - **GM** can hand/swap the turn to any living combatant.
  - **Owning players** can only target a character they own, and only while a fellow player character is active — never pulling the turn off a monster's turn. Reuses Foundry ownership (`testUserPermission`) and mirrors the existing drag-reorder rule.
  - Players can't write `combat.turn` (or reorder a combatant they don't own), so both the take and swap player paths relay to the primary GM over the existing `system.nimble` socket, each with a stale-active-combatant guard against races.

## Tests

New unit tests across `combat.svelte.test.ts` and `combatTurnActions.test.ts` covering `setActiveTurnToCombatant`, the new `swapTurnWithActiveCombatant` slot exchange, the depleted-vs-still-has-actions end-of-turn refill, the drag-to-front vs preserve-active branches, and the permission/relay matrix (including the GM swap path and the swap-onto-self refusal). Full suite: **1613 passing**. `type-check`, `biome lint`, `svelte` lint, and `circular-deps` all clean.

## Needs human review

**English** localization keys added to `public/lang/en.json` (no other locales touched), per `AGENTS.md`:
- `NIMBLE.ui.combatTracker.takeTurn` → "Take the Turn"
- `NIMBLE.ui.combatTracker.swapTurns` → "Swap Turns"
- `NIMBLE.ui.combatTracker.pingToken` → "Ping Token"

## Not yet done

Logic + unit-test verified, but not yet smoke-tested live in Foundry — worth clicking through both menu actions, the depleted-turn refill, and the player→GM socket relay for both take and swap before merge.

Closes #718

### DIFF
--- a/src/documents/combat/combatSorting.ts
+++ b/src/documents/combat/combatSorting.ts
@@ -2,6 +2,7 @@ import {
 	canCurrentUserReorderCombatant,
 	getCombatantTypePriority,
 } from '../../utils/combatantOrdering.js';
+import { canUserTakeCombatTurn } from '../../utils/combatTurnActions.js';
 import { isCombatantDead } from '../../utils/isCombatantDead.js';
 import { getCombatantManualSortValue } from './combatantSystem.js';
 import { getSourceSortValueForDrop } from './combatCommon.js';
@@ -399,7 +400,14 @@ export async function applyOwnerSort(params: {
 	} as Record<string, unknown>);
 	params.combat.turns = params.combat.setupTurns();
 
-	if (shouldSourceTakeActiveTurn(params.combat, params.dropResolution)) {
+	// Only take the active turn when the GM relay would actually accept it. Otherwise the optimistic
+	// local sync below would repoint this client's turn pointer to a target the GM rejects (e.g. a
+	// monster holds the turn), leaving the owner's tracker desynced until the next combat update.
+	const sourceCanTakeActiveTurn =
+		shouldSourceTakeActiveTurn(params.combat, params.dropResolution) &&
+		canUserTakeCombatTurn(params.combat, params.dropResolution.source, game.user);
+
+	if (sourceCanTakeActiveTurn) {
 		// Owners cannot persist combat.turn directly; relay the active-turn change to the GM.
 		const sourceId = getCombatantId(params.dropResolution.source);
 		await params.syncTurnToCombatant(sourceId, { persist: false });
--- a/src/documents/combat/combatSorting.ts
+++ b/src/documents/combat/combatSorting.ts
@@ -266,6 +266,31 @@ function shouldUseBlockSort(dropResolution: DropResolution): boolean {
 	return dropResolution.sourceCombatants.length > 1 || dropResolution.targetCombatantIds.length > 1;
 }
 
+/**
+ * After a reorder, decide whether the dragged source should take over the active turn. The source
+ * becomes active only when it has been moved ahead of (before) the previously active combatant in
+ * the new turn order — i.e. dropped toward the front. Otherwise the previously active combatant
+ * keeps the turn, preserving the reorder-does-not-change-active behavior for everything else.
+ */
+function shouldSourceTakeActiveTurn(combat: Combat, dropResolution: DropResolution): boolean {
+	const previousActiveCombatantId = dropResolution.previousActiveTurnIdentity?.combatantId ?? '';
+	if (!previousActiveCombatantId) return false;
+
+	const sourceId = getCombatantId(dropResolution.source);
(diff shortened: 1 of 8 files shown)


src/documents/combat/combatSorting.ts
```diff
@@ -354,8 +398,16 @@ export async function applyOwnerSort(params: {
...
 	} as Record<string, unknown>);
 	params.combat.turns = params.combat.setupTurns();
-	await params.syncTurnToCombatant(params.dropResolution.previousActiveTurnIdentity, {
-		persist: false,
-	});
+
+	if (shouldSourceTakeActiveTurn(params.combat, params.dropResolution)) {
+		// Owners cannot persist combat.turn directly; relay the active-turn change to the GM.
+		const sourceId = getCombatantId(params.dropResolution.source);
+		await params.syncTurnToCombatant(sourceId, { persist: false });
```
Fronix, line 413: The local turn sync runs before the GM relay is validated. When a monster holds the active turn, `requestSetActiveTurn` is rejected by `canUserTakeCombatTurn` (players can't take a turn from a non-character), but this optimistic `syncTurnToCombatant(sourceId, { persist: false })` has already repointed this client's turn pointer, leaving the owner's tracker desynced until the next combat update. Consider gating the local sync on the same eligibility (e.g. only when the active combatant is a character).

Fronix (approved):
A small suggestion around the owner drag-to-front. LGTM!

src/documents/combat/combatSorting.ts
trevlar: Good catch — fixed in fd777e3. `applyOwnerSort` now gates the optimistic local turn sync (and the GM relay) on `canUserTakeCombatTurn(combat, source, game.user)`, the same eligibility the GM-side handler enforces. When a monster (or any non-character) holds the active turn the owner drag-to-front no longer repoints the local turn pointer, so the tracker stays in sync; the reorder still applies, it just keeps the previous active turn.

PR #833, opened 2026-06-26
feat(ancestry): separate ancestry bonus traits into swappable items

### LINKED ISSUE #159: On character creation, add feature for selecting ancestry specifics when choosing multi-option such as oozeling/construct,
On character creation, provide the ability to choose a specific ancestry for a character when selecting an ancestry that has multiple options, such as dryad/shroomling, minotaur/beastfolk, or oozeling/construct.

trevlar:
Implements #159 — separates an ancestry from the bonus trait it grants.

## What changed
Each ancestry now references a **default bonus** (a new `ancestryBonus` item type). During character creation the default is pre-selected, but players can swap it for **any** other ancestry bonus. Both the ancestry and the chosen bonus are added to the character, and the bonus is shown **nested under the ancestry** in the Features tab.

### New `ancestryBonus` item type
- Data model, document (replaces any existing bonus on create), sheet
- Registered in `itemDataModels`, `registerDocumentConfig`, `game.ts`, `hooks/init.ts`, `public/system.json` (documentType + new `nimble-ancestry-bonuses` pack), `config.ts`, `preparePackIndexes`, `en.json`, ambient `item.d.ts`
- `NimbleCharacter` gained an `ancestryBonus` getter

### Ancestry changes
- Added `system.defaultBonus` (compendium UUID) + sheet config control/link
- Dragging an ancestry onto a character swaps its bonus too. An ancestry that declares **no** default leaves the character's bonus alone — that's the homebrew/module "choose any bonus" case, not a reason to delete a deliberate pick.

### Content (24 ancestries)
- Each ancestry's trait (rules + post-`<hr>` description) was extracted **verbatim** into a new `packs/ancestryBonuses/` item; ancestries keep flavor-only text + a `defaultBonus` reference. Stable IDs registered in `packs/ids.json`. No game rules were invented.

### Character creation
- New **ANCESTRY_BONUS** stage with default + swap UI. It only gates when there are bonuses to pick, so a world without the bonus pack can still finish a character.
- Ability/skill bonuses, neutral-save choice, granted languages, and save roll-mode resolution now read from the selected **bonus** instead of the ancestry
- `submitCharacterCreation` creates the bonus alongside the ancestry and applies the chosen save to the bonus rule. It flags the create batch so the ancestry doesn't also grant its default, which would be created and immediately deleted.

### Migration
- `Migration035AncestryBonusSplit` splits existing actors' ancestry trait into a bonus item, and brings world-level ancestry Items in line; bumped `LATEST_SCHEMA_VERSION` to 35

## Drive-by changes
Four changes ride along that aren't strictly part of the split:

- **`ActorSavingThrowConfigDialog` `sourceId`** — a real standalone bug. `item.flags?.core?.sourceId` was always `undefined`, because `NimbleBaseItem.sourceId` (`base.svelte.ts:88`) reads `_stats.compendiumSource`. The repair path therefore always fell through to the destructive strip.
- **`CharacterMovementConfigDialog`** — the adjustment logic moved out to `CharacterMovementConfigDialog.utils.ts` so it's unit-testable, and generic speed bonuses now show up in the dialog. Needed because the ancestry bonus is where speed rules live now.
- **`DocumentPicker`** — moved so both the ancestry and ancestry-bonus sheets can use it.
- **`_button.scss`** — added a `secondary` variant, used by the bonus swap UI.

## Verification
- `pnpm type-check` clean
- **3274/3274** vitest tests pass (incl. the `systemId` hardcode guard)
- `pnpm lint` (biome + eslint) and `pnpm circular-deps` clean

> The compendia `.db` build wasn't run here (worktree had no `node_modules`), but source JSON, IDs, and manifest are valid — it builds via `pnpm build:compendia`.

## ⚠️ Needs human review
Per AGENTS.md, English `en.json` strings were added (type label, sheet name, feature heading) and must be reviewed.


Fronix (dismissed):
## Must fix

### 1. `displayAs` dropped from Gnome and Orc — reverts merged PR #829

`packs/ancestries/core/common/gnome.json`, `packs/ancestries/core/exotic/orc.json`

```
dev:  {... 'values': ['dwarvish'], 'label': 'Optimistic', 'id': 'utvbMgCYMl6eoIm7', 'displayAs': 'Gnomish'}
PR:   {... 'values': ['dwarvish'], 'label': 'Optimistic', 'id': '1NHcYZmr7DQOknfc'}
```

These are the only two `displayAs` values in the entire pack tree, and both were lost while the rule's `id` was regenerated on the same line — the signature of a rebase clobber. `settings/languageSettings.ts:175` reads it, gated on `item?.type !== 'ancestry'`, so the bonus can't supply it. The In-Game Languages dialog stops showing `Dwarvish (Gnomish)` / `Goblin (Orcish)`.

Two aggravating factors: the descriptions this PR rewrote still say "(but you call it Gnomish, of course)", so text and data contradict; and Migration035 preserves language rules verbatim, so **migrated worlds keep the alias while fresh worlds lose it**.

### 2. Late `fromUuid` overwrites a bonus the player already picked and confirmed

`view/dialogs/characterCreation/state.svelte.ts:666-675`

```ts
fromUuid(defaultBonusUuid).then((doc) => {
    if (selectedAncestry === ancestry) {   // ancestry identity only
        selectedAncestryBonus = doc as ...;
    }
})
```

The guard checks the ancestry, not whether the player has since made a manual pick. While the default is in flight the component deliberately renders a clickable "Choose Ancestry Bonus" path (`AncestryBonusSelection.svelte:121-135`), so the player can browse, pick, and Confirm — then the pending default lands on top. `ancestryBonusConfirmed` stays `true` and the wizard doesn't re-gate, so there is **no visual cue**; it also fires the save-reset effect (`:632-635`), wiping a neutral-save choice. Reproduced with a scratch test. Needs a per-request token or "manual pick wins".

### 3. Unrecoverable dead-end when the bonus list is empty

`AncestryBonusSelection.svelte:71-98` + `.svelte.ts:20,38-58`

The only exits from `browsing` are `handleBonusSelection` (needs a card to click), or `confirmSelection`/`editSelection` (not rendered while browsing and active). The empty-list branch renders **only a warning `Hint`** — no button. If `getChoicesFromCompendium('ancestryBonus')` returns nothing while the ancestry's `defaultBonus` still resolves, clicking "Change Ancestry Bonus" strands the player on stage `1b` permanently. Same root, lower severity: browse mode has no cancel/back even with a populated list.

### 4. The new Default Bonus control is broken two ways

`view/sheets/AncestrySheet.svelte:36,39`

```js
let defaultBonus = $derived(defaultBonusUuid ? fromUuidSync(defaultBonusUuid) : null);
function openDefaultBonus() { defaultBonus?.sheet?.render(true); }
```

Per the bundled types: *"If the uuid resolves to a compendium document, that document's **index entry** will be returned instead."* Every real `defaultBonus` is a `Compendium.…` UUID, so this is always a plain index object — `.name` renders, `.sheet` is `undefined`, and **the button is a permanent no-op**.

Second, `strict` defaults to `true` ("Throw an error if the UUID cannot be resolved synchronously") and line 109 lets a GM type arbitrary text into `system.defaultBonus`. A throw inside a `$derived` breaks the sheet render, leaving no UI to fix the value that caused it. The repo's own `DocumentPicker.svelte.ts:20-37` try/catches `fromUuidSync` for exactly this reason — and it already supports drag-drop with `documentTypes` filtering, a better fit than a raw text input.

### 5. Babele translations not regenerated

- No `nimble.nimble-ancestry-bonuses.json` in `public/lang/babele/{es,fr}/`. All 13 pre-existing packs have one; the 14th doesn't. The 24 new items are untranslatable.
- Worse, `nimble.nimble-ancestries.json` is **stale** — the `Dwarf`/`Birdfolk` entries still hold the pre-split description including `<hr><p><strong>Stout</strong></p><p>+2 max Hit Dice [M]</p>…`. Babele keys by name, so it re-injects the trait onto the now-trait-free ancestry: es/fr users see it **twice**. The `rules` maps also target ids that no longer exist against an ancestry that now has one rule.

`pnpm lang:build-compendium-skeletons` is manual and not part of `pnpm build`.

---

## Should fix

**6. Migration repair branch is a no-op for text-only ancestries** — `Migration035AncestryBonusSplit.ts:131`. `if (duplicated.length === 0) return;` runs before any description stripping. An ancestry with zero non-language rules always has `bonusRules === []`, so `duplicated` is always empty. That's **10 of 24** (Bunbun, Changeling, Crystalborn, Dryad/Shroomling, Goblin, Kobold, Minotaur, Ratfolk, Stoatling, Wyrdling): trait text stays on both items permanently, and a re-run hits the same early return. This is exactly the partial-pass case the docstring at `:33-38` claims to handle.

**7. Migration never backfills `system.defaultBonus`** — `AncestryDataModel.ts:14` initials it to `''` and only the pack copies get populated. Every pre-existing character's embedded ancestry keeps `''`, so `AncestrySheet` shows no default bonus for them forever. Related: the created bonus has no `_stats.compendiumSource`, so no future migration can identify it by source id.

**8. Empty `defaultBonus` silently yields no bonus at all** — `state.svelte.ts:246-252` gates on `selectedAncestry?.system?.defaultBonus &&`, and `AncestryBonusSelection.svelte:47` guards the section the same way. Any homebrew or module ancestry skips the stage, gets no `ancestryBonus` item, and the wizard offers no escape hatch. A test asserts this as intended — flagging because the outcome is a silently under-built character.

**9. The bonus can become invisible while its rules stay live** — `PlayerCharacterFeaturesTab.svelte:54,169,340`. `groupItemsByType` drops every `ancestryBonus`, and the only render site requires an `ancestry` category to exist. Nothing cascades the delete: no `_onDelete`/`_preDelete` on `ancestry.ts`, `deleteItem` is a bare `deleteEmbeddedDocuments`, and neither the wizard nor Migration035 sets `grantedById`.

Delete the Dwarf card from a Dwarf + Stout character and "Stout" vanishes from the sheet while its `-1 Speed` and `+2 max Hit Dice` keep applying — visible in the movement config dialog. No UI path remains to edit or delete it. Separately, `ancestryBonusItems` derives from the already search-filtered `items`, so searching "Stout" filters out the ancestry, the category disappears, and the bonus vanishes too — the search returns nothing for a term that does match.

**10. `fixCorruptedRulesArrays` can never restore, only destroy** — `ActorSavingThrowConfigDialog.svelte:140`:

```js
const sourceId = item.flags?.core?.sourceId;   // nothing in src/ ever writes this key
```

Wrong twice over: the system stamps `_stats.compendiumSource`, and `NimbleBaseItem` already exposes a correct `get sourceId()` at `base.svelte.ts:88` reading `_stats.compendiumSource ?? flags.core.source` — note `.source`, not `.sourceId`. Migration035 mints the bonus with `flags: {}` and no `_stats`. So the restore at `:141-148` is always skipped and the destructive `:151-152` strip runs — for exactly the migrated population the new comment at `:128-129` claims to cover.

Root cause is pre-existing (the ancestry path was already broken this way) and it only fires when rules are already corrupt, but the PR extends the broken branch to a second type and adds a comment asserting coverage it doesn't have. One-line fix: use `item.sourceId`.

**11. Test gaps.** `CharacterCreationDialog.test.ts:26,38` wasn't updated for the new required `ancestryBonusOptions` prop (a new `svelte-check` error; it only passes because `{#if state.selectedAncestry}` keeps that branch unrendered). No dialog-level test of the new stage, no test for `ancestryBonus._preCreate`, no migration test covering #6. `svelte-check` goes 414 → 422 errors, all in three new/changed files (not CI-gating, but a real delta).

---

## Minor

- **`grantMovement` fallback is dead and has the wrong default** — `collectMovementAdjustments.ts:121` uses `rule.mode ?? 'walk'`, but `grantMovement.ts:12-19` declares `mode` as `required: true, nullable: false, initial: 'fly'`.
- **The movement fix itself is correct and worth keeping** — the old `rule._source?.movementType !== undefined` test was *always* true, because `RulesManager` builds rules as DataModels so `movementType` materialises as `null` even when the pack omits it. Every generic speed bonus (Stout −1, Lithe +1, Optimistic −1, Slow & Steady −2) landed in a `"null"` bucket the dialog never read. `?? 'walk'` now matches `SpeedBonusRule.applySpeedBonus():96`. It just doesn't belong in an ancestry PR.
- **`collectMovementAdjustments` skipped promotion levels** — AGENTS.md says local → feature → global, "never skip a level", global reserved for 3+ files across features. It went straight to `src/utils/` with **one** caller and isn't in the Shared Code Inventory.
- **Swapping the ancestry doesn't swap the bonus** — dropping Elf onto a Dwarf character deletes the Dwarf (`ancestry.ts:37`) but leaves Stout attached, now nested under Elf.
- **`requiresChoice` with a non-neutral target is silently dropped** — `ancestryBonusRequiresSaveChoice.ts:17-27` only gates `target === 'neutral'`, but `CharacterCreationDialog.svelte.ts:80` skips *any* `requiresChoice` rule when no save is picked. Homebrew-only today.
- **Migration edge cases** — `new RegExp(language, 'i')` at `:83` is unescaped (homebrew `Sylvan (Fey` throws; caught before mutation, but that actor is skipped forever while the world version still advances); `/<hr\s*\/?>/i` at `:59` misses `<hr class="…">` and swaps which item holds text vs. rules; `find` at `:49,55` takes only the first match.
- **Duplicate step label** — `ancestryBonusSelection.header` and the existing `backgroundOptionsSelection.header` are both "Step 2b."
- **PDF export** — drops the bonus *name*; the whole block is gated on `if (ancestry)` so an orphaned bonus is lost; and `generatePdfContent.ts:609-614` joins both descriptions then splits on the *first* `<hr>`, so an ancestry without one plus a bonus with one discards the ancestry text entirely (no shipped item has `<hr>`, so this is latent).
- Redundant unguarded `fromUuid` in `handleBonusSelection` (`.svelte.ts:41`); unkeyed `{#each}` at `:82`; `featureTypeHeadings.ancestryBonus` is dead config; gratuitous rule-`id` churn on 11 retained + 9 moved rules (harmless — the shape-signature fallback covers it — but it's what invalidates the id-keyed Babele entries).
- `relentless.json` has `"value": ""` on a `skillBonus` where CoreRules prints *"+1 Might"*. Pre-existing (copied verbatim from `dev`'s orc.json), but this PR is where it becomes an obviously-wrong standalone item.

---

## Verified clean

Content fidelity 24/24; 24 bonuses ↔ 24 ancestries strict 1:1 with correct common/exotic categorization per CoreRules-2.md; `ids.json` has 944 ids, zero duplicates, all 24 UUIDs resolving; **no em-dashes** in any new pack file; registration complete across `itemDataModels`, `registerDocumentConfig`, `hooks/init`, `system.json` `documentTypes` + pack entry, `config.ts`, `item.d.ts`, `preparePackIndexes`; migration numbering correct (dev tops at 034, `LATEST_SCHEMA_VERSION` 34→35, `checks.yml:90-99` satisfied); the `MigrationRunnerBase` change is *only* the constant; the migration can't abort the run or leave a half-mutated actor, and doesn't trip `_preCreate`; `dev-rebrand.mjs` correctly rewrites the new pack's UUIDs and `getChoicesFromCompendium` is type-agnostic, so the `nimble-dev` build is safe; the nested render adds no duplicate and reuses the identical `featureCard` snippet, so drag/configure/delete are wired like siblings; the new `secondary` button variant is properly scoped and has exactly one caller; every `localize()` key in new code resolves; back-navigation, ancestry-change-clears-bonus, bonus-swap-clears-save, and no-compendium-mutation on submit all check out.

## Bottom line

Solid, well-documented work with a genuinely clean content split. **#1 is a blocker** — it silently reverts merged work and makes migrated and fresh worlds disagree. #2–#5 are the most likely to bite users in practice. The rest can reasonably be triaged.

trevlar:
## Fixed
- **`displayAs` lost on Gnome and Orc** — re-added `Gnomish`/`Orcish` to both language rules (`packs/ancestries/core/common/gnome.json:21`, `packs/ancestries/core/exotic/orc.json:21`).
- **Rule-id churn** — restored the 21 pre-rebase ids across the ancestry and bonus packs (11 retained language rules, 10 moved trait rules), so Babele's id-keyed entries stay valid and Migration035's `id:` signature matches instead of falling back to shape alone.
- **Late default overwriting a manual pick** — the lookup carries a sequence number that both a new ancestry and the `selectedAncestryBonus` setter bump, so a resolution landing after a hand pick is discarded (`state.svelte.ts:350`, `:679`, `:851`).
- **No exit from browse mode** — added a Back button rendered for both the empty and populated list (`AncestryBonusSelection.svelte:105`).
- **Silently bonus-less homebrew ancestries** — an ancestry with no `defaultBonus` now keeps the step permanently open instead of inert, so a bonus can be chosen; it still never gates progress (`AncestryBonusSelection.svelte:44`, `:129`).
- **Default Bonus control** — replaced the raw text input and its no-op button with `DocumentPicker` filtered to `Item.ancestryBonus`, promoted to `src/view/components/` for its second consumer (`AncestrySheet.svelte:102`). No more `fromUuidSync` throwing inside a `$derived`, and no more index-entry `.sheet`.
- **Migration repair was a no-op for text-only ancestries** — the early return now also considers a stale description, so the 10 rule-less traits get their text stripped (`Migration035AncestryBonusSplit.ts:277`).
- **`system.defaultBonus` never backfilled** — a snapshot map keyed by ancestry compendium source id fills it in, ahead of the trait-free guard so post-redesign content is covered too (`Migration035AncestryBonusSplit.ts:10`, `:205`).
- **Created bonus had no pack origin** — stamps `_stats.compendiumSource` when the extracted trait name matches the ancestry's default, so a renamed homebrew trait isn't mislabelled (`Migration035AncestryBonusSplit.ts:255`).
- **Unescaped language regex / narrow `<hr>` match** — metacharacters are escaped and the separator matches attribute-carrying tags (`Migration035AncestryBonusSplit.ts:134`, `:153`).
- **Bonus invisible while its rules stay live** — the top-level `ancestryBonus` section is suppressed only when an ancestry exists to nest under, which covers both the deleted ancestry and the search-filtered case (`PlayerCharacterFeaturesTab.svelte:62`).
- **`fixCorruptedRulesArrays` could only destroy** — reads `item.sourceId` (`ActorSavingThrowConfigDialog.svelte:140`).
- **Ancestry swap stranded the old bonus** — ancestry `_preCreate` grants the incoming default, which removes the outgoing trait via the bonus's own `_preCreate`; with no default it just deletes the stale one (`ancestry.ts:16`).
- **`grantMovement` fallback default** — `?? 'fly'`, matching the schema initial (`CharacterMovementConfigDialog.utils.ts:123`).
- **Movement helper skipped promotion levels** — demoted from `src/utils/` to `CharacterMovementConfigDialog.utils.ts` alongside its only caller, matching `PdfPreviewDialog.utils.ts`.
- **`requiresChoice` dropped for non-neutral targets** — the gate no longer narrows on `target` (`ancestryBonusRequiresSaveChoice.ts:20`).
- **Duplicate "Step 2b"** — `backgroundOptionsSelection.header` is now "Step 3b"; background is Step 3, so it was the mislabelled one.
- **PDF export** — includes the bonus name, emits an entry for an orphaned bonus, and extracts per document so one `<hr>` can't swallow the other's text. Applied to `extractAncestrySectionHtml` too, which had the same gap (`generatePdfContent.ts:144`, `:616`).
- **`relentless.json` empty skillBonus value** — `"+1"`, matching every other skillBonus in the packs.
- **Odds and ends** — redundant `fromUuid` in `handleBonusSelection` dropped (the list is already resolved documents), `{#each}` keyed on uuid, `featureTypeHeadings.ancestryBonus` is live config now.
- **Test and svelte-check gaps** — `ancestryBonusOptions` added to the dialog test props; new `ancestry.test.ts` covering both `_preCreate` hooks; migration tests for the text-only repair, the backfill, the origin stamp, attributed `<hr>` and the regex escape; a state test for the manual-pick race. svelte-check is back to the 414 baseline — zero delta.

## Not fixed
- **Babele skeletons** — AGENTS.md forbids agents generating or modifying non-English locale files, so `pnpm lang:build-compendium-skeletons es fr` plus the hand-fix of the stale Dwarf/Birdfolk descriptions is left for a human. Worth noting the generator alone won't finish it: it preserves existing values, and every es/fr entry today is an untranslated English placeholder, so the stale descriptions need editing by hand.

## Approach
The rule-id restore is the one change that goes beyond cosmetics: matching ids are what let Migration035 recognise a carried rule by `id:` rather than shape, so restoring them makes the repair branch stronger, not just quieter in the diff.

The `defaultBonus` backfill uses a frozen map of source ids rather than resolving the pack at runtime. That matches how Migration020/022 already pin `Compendium.nimble.*` ids (rebrand-safe per AGENTS.md), keeps the migration synchronous and fully unit-testable, and avoids a pack read per actor at world startup. Homebrew ancestries keep an empty pointer — the sheet's new picker is the path for those.

For #8 the step is now always open once an ancestry is picked, but `getCurrentStage` still only gates when a `defaultBonus` exists. That keeps the existing "skips ANCESTRY_BONUS" test true while giving homebrew an escape hatch, rather than blocking creation on config a homebrew ancestry may never have.

Promoting `DocumentPicker` moved its props interface to `types/components/` and added it to the Shared Code Inventory, so the shared component no longer depends on `rulesBuilder` types.

Pushed as 8d6be41.

Fronix (changes requested):
## Fix before merge

### 1. Don't check in the migration test

`src/migration/migrations/ancestryBonusSplit.test.ts` (393 lines) should be deleted. Migration tests serve no purpose once the migration has run — they're dead weight in the suite from that point on. `dev` has zero of them across 40 files under `src/migration/`, so this matches existing convention.

Run it locally to validate the logic, then drop it from the commit.

### 2. Wizard dead-ends when the bonus pack is unavailable

`characterCreation/state.svelte.ts` (`getCurrentStage`)

The stage holds while `defaultBonus` is truthy and the bonus is unconfirmed. If `nimble-ancestry-bonuses` is missing or disabled, `ancestryBonuses` is empty, and there is no way to advance — character creation cannot be completed.

Fix: allow advancing when `ancestryBonuses.length === 0`.

### 3. Save-choice gate doesn't match the stamp

- `ancestryBonusRequiresSaveChoice.ts:20` gates on any `requiresChoice` rule, regardless of `target`
- `CharacterCreationDialog.svelte.ts:262` only writes `rule.selectedSave` when `target === 'neutral'`

A homebrew bonus with `requiresChoice: true, target: 'strength'` forces a save choice, applies it to the actor, but leaves `selectedSave` unset on the embedded rule — so a later re-prepare resolves differently.

Fix: drop the `rule.target === 'neutral'` condition in `processOriginSource` so the stamp covers everything the gate covers.

### 4. Nested create during character creation

`ancestry.ts:49`

`submitCharacterCreation` creates `[background, class, ancestry, ancestryBonus]` in one batch. The ancestry's `_preCreate` nests a `createEmbeddedDocuments` for its default bonus; the chosen bonus's `_preCreate` then deletes it. The default is created → deleted → recreated even when unchanged, and any `grantItem` / active effect on it runs against a document that's immediately removed.

Fix: skip `replaceAncestryBonus` when the operation already carries an `ancestryBonus` — either via a flag on `options` set by `submitCharacterCreation`, or by inspecting the create payload.

### 5. Incorrect comment

`Migration035AncestryBonusSplit.ts:6-9`

The comment claims `dev-rebrand.mjs` rewrites the `Compendium.nimble.*` ids in this file. It doesn't — the script's only `rewriteFiles` call globs `packs/**/*.json`. On the dev build the map keys never match, so the `defaultBonus` backfill and `_stats` stamping silently no-op. (The split itself doesn't use the map, so it still works. Same pattern as Migration003–034, not a regression.)

Fix: correct the comment. Optionally normalize the prefix before lookup:

```ts
packSource?.replace(/^Compendium\.nimble(-dev)?\./, 'Compendium.nimble.')
```

---

## Decide

### 6. Dropping a `defaultBonus`-less ancestry deletes the player's bonus

`ancestry.ts:29`

`dropExistingBonus()` runs when the incoming ancestry declares no default — i.e. the homebrew/module case the UI treats as "optional, choose any bonus." Dragging one onto an existing character removes a deliberate choice with no prompt. Since the bonus is designed to be independent of the ancestry, leaving it in place is arguably correct. Product call.

### 7. Scope

Four unrelated changes ride along: `CharacterMovementConfigDialog` extraction, `DocumentPicker` move, `ActorSavingThrowConfigDialog` `sourceId` fix, `_button.scss` `secondary` variant.

The `sourceId` fix is a real standalone bug — `item.flags?.core?.sourceId` was always undefined (`NimbleBaseItem.sourceId`, `base.svelte.ts:88`, reads `_stats.compendiumSource`), so that repair path always fell through to the destructive strip. Worth splitting out for visibility, or at minimum calling out in the PR description.

---

## Minor

- `Migration035` implements only `updateActor` — world-level ancestry Items keep their baked trait and get no `defaultBonus`.
- `ancestry.ts:45` assumes `toObject()` returns `_stats`; true for real documents, but the cast hides the assumption.
- `NimbleAncestryItem.prepareChatCardData` still hardcodes `'No description available.'` while the new bonus version localizes it — worth backporting.

trevlar:
## Fixed

- **Checked-in migration test** — deleted `src/migration/migrations/ancestryBonusSplit.test.ts`. Ran it locally first against the changes below (32/32) before dropping it.
- **Wizard dead-ends without the bonus pack** — the stage now also gates on `hasAncestryBonuses`, resolved from `ancestryBonusOptions` the same way `hasAncestries` is; with no bonuses to pick it falls straight through to background (`state.svelte.ts:255`).
- **Save-choice stamp narrower than the gate** — dropped the `target === 'neutral'` condition, so every `requiresChoice` `savingThrowRollMode` rule gets `selectedSave` written (`CharacterCreationDialog.svelte.ts:265`).
- **Default bonus created then deleted during creation** — `submitCharacterCreation` passes `nimbleAncestryBonusInBatch` on the create operation; the ancestry's `_preCreate` skips `replaceAncestryBonus` when it's set, so nothing is granted only to be removed (`ancestry.ts:96`, `CharacterCreationDialog.svelte.ts:350`).
- **Wrong `dev-rebrand` comment** — corrected, and took the optional normalization: `toSnapshotId`/`toInstalledId` fold ids between the snapshot namespace and the running install, so the `defaultBonus` backfill and `_stats` stamping work on `nimble-dev` and every uuid written is one `fromUuid` can resolve (`Migration035AncestryBonusSplit.ts:139`). Also fixed the same claim in AGENTS.md, which is where the comment came from.
- **Bonus deleted when a default-less ancestry is dropped** — it stays put now. An ancestry with no default states no opinion, so there's nothing to swap to (`ancestry.ts:38`).
- **`Migration035` only handled actors** — added `updateItem` for world-level ancestry Items: backfills `defaultBonus`, and strips the trait only when it's demonstrably the pack bonus's (name match). Homebrew traits stay — an Item can't own the extracted bonus, so stripping them would delete rules with no replacement (`Migration035AncestryBonusSplit.ts:381`).
- **`toObject()` `_stats` assumption** — seeds the object instead of assuming it's there (`ancestry.ts:59`).
- **Unlocalized ancestry chat card fallback** — now uses `NIMBLE.ancestryBonusSelection.noDescriptionAvailable`, matching the bonus version (`ancestry.ts:78`).
- **Scope not called out** — PR description now has a "Drive-by changes" section, leading with the `sourceId` bug and why `flags.core.sourceId` was always undefined. Also corrected the stale `Migration026` / version 26 / 1453-test references in the body.

## Not fixed

- Nothing.

## Approach

Kept the four drive-by changes in this PR rather than splitting them — Trevor's call. The `CharacterMovementConfigDialog` extraction has tests wired into this branch, and its speed-bonus fix exists because speed rules moved onto the bonus, so it isn't cleanly separable anyway.

Went with a create-operation flag over inspecting the payload for #4: `_preCreate` only sees its own source, so payload inspection would have meant a static `_preCreateOperation` override for one boolean.

New coverage for the two `submitCharacterCreation` fixes went into `CharacterCreationDialog.test.ts`, which had none for the bonus path — that's where both bugs lived.

Pushed as `69c0ef9`. Full gates green: type-check clean, 3274/3274 tests, lint + circular-deps clean.

Fronix (approved):
LGTM

PR #838, opened 2026-06-30
fix(combat): restore per-turn actions for characters and solo monsters

### LINKED ISSUE #1: PC sheets display two "Prototype token" entries and none of them is working
Foundry v13.345 / Nimble 0.1.0 / No module
Nimble PC sheets show two "Prototype token" entries and none of them is working (clicking on any of the two triggers a js error).

Image

Image

### LINKED ISSUE #4: Items not stacking and not taking inventory slots
When you define an item as stackable, its not having any effect, duplicates are not being stored together, and inventory slots are not being counted at all:

Image

trevlar:
## Summary

Fixes combat action-economy bugs seen in solo-monster encounters, where action points were not restored correctly at the start/end of turns.

### Action model (for reference)
- A hero's **max actions is always 3**.
- The **initiative roll only seeds starting actions for round 1** (≥20 → 3, ≥10 → 2, <10 → 1).
- **End of turn** → a hero's actions restore to 3.
- **Dying** → restore to 1 only.
- **Dazed** (which applies Hampered) → one fewer action: reduces *current* by 1, max stays 3.

### 1. Refill the correct hero on turn-end (expanded solo-monster turns)
Foundry derives the combatant passed to `_onEndTurn` from its own `previous` turn snapshot. Our direct turn-index resync (`#syncTurnIndexWithAliveTurns`) can leave that snapshot stale when the turn list contains **duplicate solo-monster entries**, so the turn-end refill could target the wrong combatant — a hero might not regain actions on their own turn-end, or a hero who hadn't acted yet could be refilled early (matching the report).

**Fix:** capture the active combatant immediately before advancing (`nextTurn`/`nextRound`) and have `_onEndTurn` refill *that* captured combatant. Falls back to Foundry's value when there's no capture (drag-drop, direct calls), so it's a no-op except in the buggy case. The captured id is cleared in a `finally` so it never leaks into a later turn-end.

### 2. Solo monsters refill on each interleaved turn
Solo monsters take a turn after each character (`expandLegendaryTurns`), but their actions were only reset at `_onEndRound`. `nextTurn`/`nextRound` never restored the incoming non-character, while `previousTurn`/`previousRound` already did — an asymmetry that left a solo monster stuck at 0 actions for the rest of the round.

**Fix:** `nextTurn`/`nextRound` now restore the incoming non-character via the existing `#restoreNonCharacterTurnState()`. No-op for ordinary single-turn NPCs.

### 3. Keep hero max at 3 (initiative seeds round 1 only)
An early pass on this branch tied a hero's `system.actions.base.max` to their initiative roll — wrong, since max must always stay 3 (restoration refills to max, lowered only by Dying). Reverted; `applyCharacterInitiativeActionUpdate` now seeds only `current` and never touches `max`.

### 4. Backstop when Foundry skips the outgoing hero's turn-end entirely
The `#endingTurnCombatantId` capture (fix #1) corrects *which* combatant `_onEndTurn` refills — but only when `_onEndTurn` actually fires. Live logs from a 3-hero + solo-monster encounter showed that the **last hero before the solo monster's trailing turn never received an `_onEndTurn` call at all**, so they were left at 0 actions after clicking End Turn.

**Root cause:** Foundry's turn-event dispatcher (`#triggerTurnEvents`) decides which `_onEndTurn`s to fire by comparing its **cached** `current`/`previous` turn *indices*. Because our resync mutates `this.turn` directly (outside any `update()`), that cached history desyncs from the real position under the expanded (duplicate solo-monster) turn order, and Foundry computes an empty interval for the last hero — skipping their end-of-turn.

**Fix:** don't depend solely on Foundry firing `_onEndTurn`. The refill is extracted into `#applyCharacterTurnEndRefill`, and `nextTurn`/`nextRound` call `#ensureOutgoingCharacterRefilled` after advancing as a backstop. It refills the captured outgoing hero when `_onEndTurn` never ran, deduped by id so a single advance never refills — or fires the `nimbleCombatTurnEnd` hook for — the same hero twice. No-op in the cases where `_onEndTurn` already handled it.

## Tests
- `combatInitiative.test.ts` — initiative seeds `current` (all roll bands + the 20 boundary), never sets `max`, no-op for non-characters.
- `combat.svelte.test.ts`:
  - turn-end refills the captured outgoing hero even when Foundry reports the wrong combatant;
  - **turn-end refills the outgoing hero even when Foundry never fires `_onEndTurn`** (fix #4);
  - **does not double-refill when both `_onEndTurn` and the backstop run** (fix #4);
  - advancing forward into a solo monster's turn restores its actions to max;
  - two existing `nextTurn` tests updated to mock `updateEmbeddedDocuments`, now exercised by the forward restore path.

Full suite: 1660 tests pass; typecheck and Biome clean.

### DIFF
--- a/src/documents/combat/combat.svelte.ts
+++ b/src/documents/combat/combat.svelte.ts
@@ -118,11 +118,13 @@ class NimbleCombat extends Combat {
 	// `_onEndTurn` can refill the correct combatant even when Foundry's own `previous`
 	// turn snapshot is stale under the expanded solo-monster turn order.
 	#endingTurnCombatantId: string | null = null;
-	// The character combatant id already refilled by `_onEndTurn` during the current
-	// forward advance. Reset at the start of each `nextTurn`/`nextRound` so the post-advance
-	// fallback (`#ensureOutgoingCharacterRefilled`) can tell whether Foundry actually fired
-	// the turn-end event for the outgoing hero — Foundry skips it when its cached turn history
-	// desyncs from our expanded (duplicate solo-monster) turn order.
+	// The character combatant id whose turn-end refill has been claimed during the current
+	// forward advance. Foundry dispatches turn events fire-and-forget, so `_onEndTurn` and the
+	// post-advance backstop (`#ensureOutgoingCharacterRefilled`) can interleave; whichever runs
+	// first claims the id synchronously — before any await — and the other becomes a no-op.
+	// Reset at the start of every turn-changing entry point (`nextTurn`, `nextRound`,
+	// `previousTurn`, `previousRound`, `setActiveTurnToCombatant`) so a stale claim never
+	// suppresses a later legitimate refill.
 	#endingTurnRefilledCharacterId: string | null = null;
 	#initiativeRollRequests = new Map<string, Promise<LockedInitiativeRollOutcome | null>>();
 
@@ -711,10 +713,23 @@ class NimbleCombat extends Combat {
 			: null;
 		const endingCombatant = capturedCombatant ?? combatant;
 
+		// Claim the refill synchronously — before any await — because Foundry dispatches turn
+		// events fire-and-forget, so `nextTurn`'s backstop can run interleaved with this method.
+		// Whoever claims the id first performs the refill; the other becomes a no-op.
+		const endingCombatantId = endingCombatant.id ?? null;
+		let shouldRefillCharacter = false;
+		if (endingCombatant.type === 'character') {
+			shouldRefillCharacter =
+				endingCombatantId === null || this.#endingTurnRefilledCharacterId !== endingCombatantId;
+			if (shouldRefillCharacter && endingCombatantId) {
+				this.#endingTurnRefilledCharacterId = endingCombatantId;
+			}
+		}
+
 		await super._onEndTurn(combatant, context);
 
 		if (endingCombatant.type === 'character') {
-			await this.#applyCharacterTurnEndRefill(endingCombatant);
+			if (shouldRefillCharacter) await this.#applyCharacterTurnEndRefill(endingCombatant);
 		} else {
 			// @ts-expect-error Custom hook
 			Hooks.call('nimbleCombatTurnEnd', endingCombatant);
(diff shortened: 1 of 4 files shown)


src/documents/combat/combat.svelte.ts
```diff
@@ -690,20 +700,65 @@ class NimbleCombat extends Combat {
...
+		// @ts-expect-error Custom hook
+		Hooks.call('nimbleCombatTurnEnd', combatant);
+
+		await combatant.update({
+			'system.actions.base.current': getCombatantResetActions(combatant),
+			'system.actions.base.additional': 0,
+			...this.#buildHeroicReactionAvailabilityUpdate(true),
+		} as Record<string, unknown>);
+
+		if (combatantId) this.#endingTurnRefilledCharacterId = combatantId;
```
Fronix: The dedup flag is set only after two awaits (`super._onEndTurn` and the combatant update), but Foundry fires turn events without awaiting them (`_manageTurnEvents` is fire-and-forget from `_onUpdate`), so `nextTurn`'s backstop can check `#endingTurnRefilledCharacterId` before this line runs. Both paths then refill and `nimbleCombatTurnEnd` (charge recovery) fires twice. Claim the id synchronously before any await, e.g. in `_onEndTurn`'s character branch before `super._onEndTurn`: skip if it already matches, otherwise set it there.

Fronix (dismissed):
One timing gap worth a look in the turn-end dedup in combat.svelte.ts.

The turn-end dedup flag is set too late to guarantee the hook fires only once per advance when Foundry dispatches the turn-end event asynchronously.

src/documents/combat/combat.svelte.ts
trevlar: Done in 05e1d73e5 — the refill is now claimed synchronously before any await. `_onEndTurn` check-and-sets `#endingTurnRefilledCharacterId` in its character branch before `super._onEndTurn` (skipping its refill if the backstop already claimed it), and `#ensureOutgoingCharacterRefilled` claims the id before awaiting the refill, so whichever path runs first wins and the other is a no-op — the hook can no longer fire twice. Since a match now suppresses the refill, the claim is also cleared at the start of `previousTurn`/`previousRound`/`setActiveTurnToCombatant` so a stale claim from a prior advance never swallows a legitimate later turn-end.

Fronix (approved):

PR #847, opened 2026-07-11
fix(combat): apply banked dice-pool bonuses through monster armor

Fronix:
## Summary

Fixes Apply Damage dropping banked dice-pool bonuses (Berserker Fury Dice, Oathsworn Judgment Dice) against armored monsters, and adds a per-target damage preview so the GM can see exactly how much each target will take before applying.

image

## Changes

- Count banked dice-pool contributions as dice for monster armor. These bonuses ship on the damage roll as flavored numeric terms, so `calculateArmorAdjustedDamage` was treating them as ignorable modifiers and stripping them for medium/heavy armor targets. Per the rules ("your Fury Dice are dice when calculating damage for monster armor") they now count toward dice damage: full against medium armor, halved against heavy. Flat static modifiers stay stripped by armor as before.
- The fix is general across every banked-dice mechanic, not just Fury Dice: Fury Dice (auto and manual), Judgment Dice, and charge-die pools (Combat/Mana Dice) all resolve correctly. Charge dice were already counted; flat damage bonuses remain armor-stripped.
- Add a per-target damage preview next to each name in the chat card target list (GM only), computed through the same armor-aware path as Apply Damage so the number always matches what gets applied. Shows nothing on cards with no applicable damage (healing, conditions, save-gated damage).

## Test Plan

1. As a Berserker with banked Fury Dice, attack a monster that has Heavy Armor (e.g. Basilisk) with a melee weapon and hit.
2. Confirm the target row shows a damage preview in parentheses next to the name, and that it equals half the combined weapon dice plus Fury Dice rounded up, with flat modifiers excluded.
3. Press Apply Damage and confirm the target's HP drops by that same previewed amount (previously it dropped only by the base weapon die).
4. Repeat against a Medium Armor target and confirm the full dice total (weapon dice plus Fury Dice) is applied, with flat modifiers excluded.
5. Repeat against an unarmored target and confirm the full card total is applied and previewed.
6. Confirm a Berserker manual Fury Dice spend and an Oathsworn Judgment Dice spend behave the same way.

## Checklist

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -1820,7 +1820,8 @@
 			"noTargetsSelected": "No targets selected",
 			"addSelectedTokensAsTargets": "Add Selected Tokens as Targets",
 			"addTargetedTokensAsTargets": "Add Targeted Tokens as Targets",
-			"removeTarget": "Remove Target"
+			"removeTarget": "Remove Target",
+			"damagePreview": "Damage this target will take (after armor)"
 		},
 		"chatEffects": {
 			"noEffect": "No Effect"
--- a/src/documents/chatMessage.test.ts
+++ b/src/documents/chatMessage.test.ts
@@ -1225,14 +1225,52 @@ describe('NimbleChatMessage.getDamagePreviewForTarget', () => {
 		expect(message.getDamagePreviewForTarget('Scene.scene.Token.token')).toBe(42);
 	});
 
-	it('returns 0 for a miss (no damage is applied)', () => {
+	it('returns null for a miss so the target list shows no preview', () => {
 		globals().fromUuidSync.mockReturnValue({
 			actor: { system: { attributes: { armor: 'none' } } },
 		});
 
 		const message = createDamageMessage({ roll: battleaxeRoll(), isMiss: true });
 
-		expect(message.getDamagePreviewForTarget('Scene.scene.Token.token')).toBe(0);
+		expect(message.getDamagePreviewForTarget('Scene.scene.Token.token')).toBeNull();
+	});
+
+	it('counts a disposition-targeted damage node once when its outcome child is also surfaced', () => {
+		globals().fromUuidSync.mockReturnValue({
+			actor: { system: { attributes: { armor: 'none' } } },
+		});
+
+		const message = new NimbleChatMessage({
+			type: 'spell',
+			system: {
+				targets: ['Scene.scene.Token.token'],
+				isCritical: false,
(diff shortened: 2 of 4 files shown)


trevlar (approved):
**✅ Approved**

This correctly fixes Apply Damage stripping banked dice-pool bonuses (Fury/Judgment Dice) against armored monsters, and the per-target preview is a nice touch. The reasoning holds up under tracing — I confirmed the central assumption end to end: in a damage roll, only pool contributions carry flavor on numeric terms (`+${face}[${label}]` for manual/auto pool spends in `itemActivationConfigDialogState`/`buildAutoBonusFormula`), while situational mods (`+${situationalModifiers}`) and flat feat bonuses (`+ ${numericBonus}` in `ItemActivationManager`) are appended unflavored and stay armor-stripped. Charge/dice bonuses are real dice and were already counted. So `getTermFlavor`-based classification is sound.

- The preview genuinely mirrors the apply path: outcome derivation matches `DamageNode`/`DamageOutcomeNode` (`isMiss ? 'noDamage' : 'fullDamage'` and `node.outcome`), the half-damage multiplier matches `DamageRoll.svelte`, and it reuses `calculateArmorAdjustedDamage`, so the number always equals what gets applied.
- Test coverage is thorough — Fury (auto + manual), Judgment, medium/heavy/none armor, crit, `ignoreArmor`, miss, and the null/no-actor preview cases are all exercised.

**Worth confirming (non-blocking)**

- **src/view/chat/components/Targets.svelte:112** — on a *miss*, `getDamagePreviewForTarget` returns `0` (not `null`, since the card still has a damage roll), so each target renders a red `(0)`. Your PR description says the preview "shows nothing on cards with no applicable damage" — a miss is a borderline case. If a red `(0)` on every missed attack is undesired visual noise, return `null` when the summed total is `0` on a miss, or gate the render on `damagePreview > 0`.
- **src/documents/chatMessage.ts:588** — latent double-count: `findNodesByContexts` surfaces a base `damage` node *and* its `on.hit` children only when the base node has `targetDisposition`. No current content hits this (Searing Light's disposition damage has no `on.hit`), so it's theoretical — but if such a node ever exists, `#collectApplicableDamageRolls` would count its roll twice. A one-line guard (skip a `damage` node when it has surfaced outcome children) would harden it.
- **src/documents/chatMessage.ts:577** — small doc-comment grammar nit: "Save-gated damage is intentionally excluded its per-target outcome is unknown" is missing a "—" or "because".

PR #851, opened 2026-07-15
feat(rules): damageReduction rule and banked Fury Dice reduction

### LINKED ISSUE #580: [Foundation] Rule type: damageReduction
Reduce incoming damage by a flat number or formula, optionally scoped by damage type, gated by the inherited `predicate` field.

```jsonc
{
  "type": "damageReduction",
  "label": "Armor of Shadows",
  "value": "@abilities.int.mod",
  "damageTypes": ["any"],
  "predicate": { "self": "hasShadowMinion" }
}
```

Don't add a bespoke `condition` field. Predicate is already on every rule via the base class and uses the #579 vocabulary.

## Shape to mirror

Follow `damageBonus`:

- `afterPrepareData()` pushes entries into `actor.system.damageReductions[]`
- An activation-site filter consumes the array, filters by `damageTypes` and `predicate`, sums matching entries, applies before HP deduction

The reduction's application point in the incoming-damage pipeline is worth tracing before starting. Find where damage hits HP today and put the reduction one step before that.

## Stacking (open)

Original issue flagged "stacking rules defined" as undecided. Still open. Default suggestion: multiple reductions sum unless something later needs a non-stacking mode. Worth a quick rulebook check first; Nimble may have a uniform position on reduction stacking.

## Features this unblocks

- Berserker: "That all you got?!" (`predicate: { self: 'raging' }`)
- Shadowmancer: Armor of Shadows
- Fey Touched background, half magic damage (`damageTypes`)
- Various resistance / vulnerability features

---

Needs #579 first (predicate vocabulary).

Fronix:
## Summary

Adds the `damageReduction` foundation rule (#580) and completes the Berserker Rage stack end-to-end: "That all you got?!" now spends Fury Dice into a visible, banked one-shot reduction that is applied automatically on the next damage taken. Along the way this establishes a system-wide policy that Active Effects are never hidden. A multi-agent review pass hardened the edge cases (full absorption, automation off, sheet closed, multi-roll/multi-token cards).

## Changes

- **New rule type `damageReduction`**: reduce incoming damage by a flat/formula amount, optionally scoped by damage types (empty = all), gated by the standard predicate. Entries accumulate on `actor.system.damageReductions` and are consumed at damage application, after outcome/armor halving and before temp HP. Multiple reductions sum (rulebooks define no stacking rule).
- **Damage pipeline**: `DamageApplyOptions` now carries `damageType` (was silently dropped); both Apply Damage and the per-target damage preview go through the same reduction-aware calculation. Typed reductions require a known damage type, so the minion-group card (no roll metadata) only receives untyped reductions.
- **Banked one-shot reductions**: `diceConsumer` gains `effectType` (`generic` | `damageReduction`). A manual spend with `damageReduction` banks the rolled total as an Active Effect on the actor ("Damage Reduction (8)", source feature's icon, tooltip description). Repeated spends accumulate; the next damage application consumes and deletes it, even when it absorbs the hit entirely (Apply Damage stays clickable for that case); deleting drops it, disabling suspends it. The bank is applied once per actor per card: the first application entry only, so multi-roll previews and same-actor-via-two-tokens targeting match what actually lands.
- **Spend flow discoverability**: activating an item with a manual dice consumer opens the pool's spend panel pre-selected to that feature (client-local hook, `dicePool.requestSpend`). If the character sheet is closed (e.g. hotbar macro), a ready-hook router opens it first and the tracker consumes the parked request on mount. Empty pools notify instead.
- **One message per use**: rules can declare `suppressesActivationCard()`; the manual-spend flow suppresses the redundant descriptive activation card. Suppression only happens when the activation produced no rolls/effect nodes **and** "Auto-Apply Conditions from Rules" is enabled — that setting gates the rule dispatch that opens the spend panel, so with automation off the normal card posts instead of a silent no-op. The `useItem` hook still fires so charges and triggers keep working.
- **No hidden Active Effects**: every enabled AE renders on the token (`temporaryEffects` override) **and** in the canvas conditions panel and sheet effects lists — both sourced from `allApplicableEffects()`, so item-granted effects (which stay embedded in their item on v13) show up too, bucketed as passive; ad-hoc effects bucket as transient. Both buckets are deletable: deleting an item-granted effect asks for confirmation (not recoverable without re-adding the item) and routes through the effect's owning document; ad-hoc effects remove instantly like conditions.
- **Migration 029**: existing world copies of "That all you got?!" get `effectType: "damageReduction"` stamped on their manual Fury consumer and canonical descriptions upgraded to the click-to-use wording (matched by compendium source id, falling back to berserker class + name; customized rules/descriptions are left alone).
- **Chat polish**: dice sound only plays when an activation actually rolled dice.
- Pack: "That all you got?!" wired with `effectType: "damageReduction"`, description shortened to click-to-use.
- Internal: the activation-card creation tail is shared between item types, and dice-formula detection / actor-array accumulation are shared `NimbleBaseRule` helpers (`isDiceExpression`, `pushToActorSystemArray`) used by `damageBonus` and `damageReduction`.

## Test Plan

1. Enable the world setting **"Auto-Apply Conditions from Rules"** and reload (the spend-panel flow is gated on it).
2. Create a Berserker with Rage and "That all you got?!". Activate Rage, attack something so Fury Dice accumulate.
3. Have a monster deal damage to the Berserker. Click "That all you got?!": no chat card posts; the Fury Dice spend panel opens with TAYG pre-selected. Pick dice and Spend: one chat card posts with the rolled reduction, and a "Damage Reduction (N)" effect (TAYG icon) appears on the token, the canvas panel, and the sheet's Temporary Effects.
4. GM clicks Apply Damage on the monster's damage card: damage lands reduced by N (armor halving first, then reduction, floor 0; temp HP absorbs the remainder), and the effect disappears. The per-target damage preview in the card's target list matches the applied number.
5. Full absorption: bank more reduction than the incoming hit. Apply Damage stays clickable; clicking it deals 0, notifies, and deletes the bank (it must not survive to the next attack).
6. Automation off: disable the setting, reload, click TAYG — a normal activation card posts (no silent no-op). Re-enable afterwards.
7. Sheet closed: drag TAYG to the macro hotbar, close the sheet, trigger the macro — the sheet opens with the spend panel pre-selected.
8. Typed check: add a `damageReduction` rule with `damageTypes: ["lightning"]` via the Rules Builder to any feature; lightning damage is reduced, fire is not.
9. Rage gating: a `damageReduction` rule with predicate `{ "self": "raging" }` applies only while Rage is toggled on.
10. Effect deletion: right-click the banked reduction icon on the canvas panel — instant removal. Add an item carrying a transferred Active Effect: it appears under Passive Effects on the sheet and on the canvas panel; deleting it (canvas right-click or sheet trash button) shows a confirmation dialog first, and confirming removes the effect from the owning item.
11. Roll-less features/spells no longer play the dice sound; damage-dealing activations still do.
12. Migration: open a world last saved before this PR where a Berserker owns "That all you got?!" — after the migration runs, the feature's Fury consumer rule carries `effectType: "damageReduction"` and the description ends with "Click this feature to use it."

(Multi-roll preview math and same-actor-via-two-tokens application are covered by unit tests in `chatMessage.test.ts`.)

## Checklist

## Related Issues

Closes #580

Deferred follow-ups:
- Builder-editable `suppressActivationCard` envelope field on all rules (spec: `_bmad-output/planning-artifacts/suppress-activation-card-envelope-field.md`)
- Armor of Shadows pack wiring (blocked on a minion-count roll-data key)
- Half-damage / resistance mode (Fey Touched, monster `damageResistances`)
- Banked reduction expiry (currently persists until consumed or manually removed)
- `effectType` registry: the `damageReduction` spend side-effect is currently an `if` branch in `DicePoolPanel.svelte.ts`; worth generalizing when a second effect type (temp HP, healing) exists


trevlar (approved):
**✅ Approved**

This cleanly lands the `damageReduction` rule and the banked one-shot Fury Dice reduction end-to-end, and the edge cases you called out in the summary genuinely hold up under reading — full absorption still spends the bank (`applyDamage` consumes `bankedReductionActors` before the zero-damage exit), Apply Damage stays live for the absorb-everything case, and the once-per-actor-per-card guard handles multi-token/multi-roll correctly.

- Nice consistency work: `calculateAdjustedDamage` is shared by both the apply path and the per-target preview, so the number in the card's target list can't drift from what actually lands.
- Conventions are tight throughout — `SYSTEM_ID`/`systemHookName`, computed flag keys, complete `en.json` coverage, and Migration 029 is idempotent (leaves non-`generic` effectTypes and customized descriptions alone) with `LATEST_SCHEMA_VERSION` bumped to match.

PR #853, opened 2026-07-19
fix(combat): let players roll initiative from their character sheet

trevlar:
## Problem

Players got a *"lacks permission to update Combat"* error when rolling initiative **from their character sheet** (both the initiative-value button and the ActionTracker button). The d20 icons on/around the combat tracker worked fine.

Reported on Discord by a GM whose players couldn't roll into initiative from their sheets.

## Root cause

In Nimble, initiative sets a combatant's **starting number of actions** — it does **not** determine turn order (turn order follows the manual combatant sort in `sortCombatants`: dead state → `system.sort` → type priority → name).

`Combat#rollInitiative`, however, still carried Foundry core's initiative-based turn reconciliation. After rolling, it ran a turn update on the **Combat document**:

```ts
if (updateTurn && currentId && updates.length > 0) {
    await this.update({ turn: this.turns.findIndex((t) => t.id === currentId) });
}
```

`NimbleCombat#update` augments any `turn`/`round` update with a `flags.nimble.expandedTurnIdentity` write. Foundry lets a non-GM update a Combat's **turn-only** fields, but the added flag write makes it a disallowed update. A player may set their own initiative, but cannot write to the GM-owned Combat document — so they hit the permission error.

Because turn order in Nimble is decoupled from initiative, this reconciliation was a no-op that only ever caused harm: rolling initiative never reorders the turn list, so there is no active-turn index to correct.

## Fix

Remove the turn reconciliation from `rollInitiative` entirely (rather than gating it on GM), since it only made sense under D&D-style initiative-as-turn-order. This fixes the player permission error and drops logic that doesn't apply to Nimble. GMs and players alike simply set initiative; turn order is untouched.

The existing "already rolled" guard is unchanged — if a combatant's initiative is already set, it still can't be re-rolled.

## Tests

- New regression tests in `combat.svelte.test.ts`: a non-GM can roll initiative with **no** write to the Combat document, and rolling initiative never updates `Combat#turn`.
- Full suite green: **2361 tests pass**; typecheck, Biome lint/format, ESLint (Svelte), and Prettier all clean.

### DIFF
--- a/src/documents/combat/combat.svelte.test.ts
+++ b/src/documents/combat/combat.svelte.test.ts
@@ -1883,6 +1883,155 @@ describe('NimbleCombat', () => {
 		]);
 	});
 
+	it('lets a non-GM roll initiative without writing to the Combat document', async () => {
+		// Regression: a non-GM may set their own initiative but cannot update the (GM-owned) Combat
+		// document. Because initiative doesn't affect turn order in Nimble, rolling it must not touch
+		// `Combat#turn` — otherwise a player rolling from their sheet hits a "lacks permission to
+		// update Combat" error.
+		globals().game.user = { isGM: false, role: 1 };
+
+		const combatId = 'combat-initiative-non-gm-turn-guard';
+		const actor = createCombatActorFixture({
+			id: 'actor-initiative-non-gm',
+			type: 'character',
+			isOwner: true,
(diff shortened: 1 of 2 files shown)


Fronix (approved):
lgtm!

PR #854, opened 2026-07-20
feat(level-up): show duplicate class features with source labels and selection

### LINKED ISSUE #707: Level Up: Show duplicate class features from items and compendium, with source labels and selection
### Problem or Motivation

When a player levels up, class features can sometimes exist in both the world items list and the compendium. Right now, this can create ambiguity or hide useful choices if duplicate features are treated as a single option.

This is a problem because the item a GM or player wants to use may differ depending on where it comes from. For example, a world item may contain custom changes, while the compendium version may reflect the default or latest packaged version. During level up, players should be able to clearly see all valid matching options and choose the one they want.

### Proposed Solution

Update the level-up dialog so that when matching class features are found in both the items list and the compendium, both options are shown instead of collapsing them into a single entry.

Expected behavior:

* Display all matching class feature candidates found from both sources.
* If multiple candidates represent the same class feature, allow the user to explicitly choose which one to take.
* Show a visible source tag for each option, such as `Item` or `Compendium`.
* Preserve the current level-up flow when only a single matching feature exists.
* Make the distinction clear enough that users can tell whether they are selecting a customized world item or a compendium-backed entry.

This would give users more control and make the selection process more transparent.

### Alternatives Considered

One alternative would be to always prioritize one source over the other, such as preferring compendium over item entries or vice versa. That would be simpler, but it removes user choice and could lead to the wrong version being added.

Another option would be to merge duplicates into a single row and reveal source details only in a tooltip or secondary view. That may reduce visual clutter, but it makes the behavior less obvious and adds friction when users need to distinguish between sources.

### Additional Context

This is especially useful in setups where:

* GMs copy compendium content into world items and then customize it
* packaged content and local overrides coexist
* users need confidence about which version of a class feature is being added during level up

trevlar:
## Summary

Closes #707.

When the same class feature exists in more than one place — a GM's customized world item and the packaged compendium original — level-up and character creation used to grant every copy silently. This adds a picker that shows each copy, says how they differ, and lets the player take one or all of them.

## What the player sees

A duplicate no longer renders as two identical cards. Each copy gets a row identified by **where it lives** (its world folder, or the compendium's own label) and **where it came from** (derived from a pack, or created in this world), because the World/Pack badge says what *kind* of thing a copy is but not *which* copy it is — with two world items, a badge alone distinguishes nothing.

Every row expands to its description, with **Compare all** to open them at once. Each row also carries a difference note — `identical to Nimble Core`, or `differs: duration, healing` — so a copy that is byte-identical announces itself and never needs opening. When a row is expanded, the words that actually changed are highlighted.

The player takes exactly one copy or all of them. **Keep all N** toggles to **Keep recommended**, so taking everything is reversible without hunting for the right radio. The world copy is recommended by default; between two world copies, the most recently modified wins.

## Behaviour decisions

Four questions came out of design review and were settled deliberately:

- **One or all.** `selectionCount: 1`, `selectionMax: N`. Arbitrary subsets ("keep 2 of 3") are not offered — two clean outcomes are easier to describe and match the filename-collision convention players already know from Finder and Explorer.
- **Recommended copy.** The world copy: if a GM edited a feature into their world, that edit is the thing they meant to use. Ties break on most-recently-modified.
- **Full field-level diff**, rather than a bare identical/differs flag.
- **Already-owned copies stay in the cluster.** Previously an owned copy was filtered out *before* clustering, which let its sibling shrink the cluster to a singleton and be auto-granted — handing the player a second copy of a feature they already had on a re-level. The owned copy now remains as the comparison baseline, taking nothing is a valid outcome (the floor drops to 0), an owned copy is never re-granted, and a cluster where every copy is owned disappears entirely.

## How it works

**Resolver** (`src/utils/getClassFeatures.ts`) — duplicate detection clusters candidates by compendium-source link, falling back to name, so a renamed world copy still matches its original. Promotion is opt-in via `promoteDuplicateSources`: the two dialogs opt in, the class sheet's progression view does not. Owned copies are retained only for auto-grant groups when promotion is on, so the named-group and class-sheet paths keep their existing ownership filtering and selection counts unchanged.

**Diff** (`src/view/dialogs/featureDiff.ts`) — compares the structured schema (`activation.cost` / `duration` / `targets` / `template` / `effects`, `rules`, `levelUpOptions`, `description`, `macro`) and returns stable field keys plus the `type` of any rule that changed, so a row can say `healing` rather than the blunter `effects`. Object key order is ignored; array order is not, since rule order is meaningful.

Descriptions are reduced to plain text before the word-level LCS diff. A word boundary is not a tag boundary, so diffing raw HTML would split tags across a highlight and emit broken nesting; plain text also means segments render as text rather than as trusted markup from a world item. Formatting is preserved everywhere else — this only affects the comparison view.

**UI** (`DuplicateSourceGroup.svelte`) — both dialogs route range groups here and keep `FeatureGroupSelection` for ordinary class choices. The row is the expander and the radio is the only thing that selects, matching `FeatureCard`'s existing behaviour.

## Fixes made along the way

- The synthetic group key was reaching the class sheet through `getClassProgressionData`, rendering as a heading and getting written into a real `system.group` by the add-feature button. Promotion is now opt-in, and the key carries no system id.
- Extracted `SourceTag` (replacing a third copy of the World/Pack badge markup and four duplicated strings), plus `getItemSource` and `selectionGroupRules`. All call sites migrated.
- A cap test could not fail: `waitFor` checks its callback once synchronously before Svelte flushes, so asserting "nothing changed" always passed. Fixed and mutation-verified.
- The picker had hand-rolled a copy of `SelectionIndicator` that set `width`/`min-width`/`height` but not `max-width`/`min-height`/`max-height`; Foundry's core button styles stretched it into an oval. It now uses the component, which gained optional `role`/`ariaChecked` (defaulting to undefined, so existing call sites are untouched) to keep radiogroup semantics.
- The identity line clips with an ellipsis instead of wrapping, so rows stay one line tall and compare down the column. `tooltipWhenClipped` restores the full text on hover, but only while the text is actually cut off.

## Testing

`pnpm check` passes: format, lint, circular-deps, type-check, and **3302 tests**.

New coverage includes the field diff (tag-splitting, script-stripping, entity decoding, rule-type naming, stable field order), the word diff, the comparison list, owned-copy handling, the recommended pick, the selection cap in both dialog factories, and the clipping action.

## ⚠️ Localization review needed

31 new English strings under `NIMBLE.classFeatureSelection` (the `duplicate*` keys and the `diffField.*` block). Per AGENTS.md these need human review; `es.json` and `fr.json` are untouched.

Note that `SourceTag` reads the existing `NIMBLE.classSheet.progressionSourceTag*` keys, so a shared component is pulling strings from the class sheet's namespace. Moving them to a neutral namespace would require matching edits to `es.json` and `fr.json`, which agents must not make — flagging it for a human pass rather than doing it here.

## Test plan

- [x] `pnpm check` — format, lint, circular-deps, type-check, full suite (3302 tests)
- [ ] Manual in-Foundry verification of the level-up dialog with a duplicated feature — **still outstanding**. Two visual defects on this branch (the oval radios, the overflowing identity line) passed the entire test suite, so the rendered dialog is the remaining risk, not the logic.


Fronix (dismissed):
## Overview

When the same class feature exists as both a GM-customized world item and its compendium original, level-up/creation used to grant every copy. This adds a duplicate-source picker: `getClassFeatures` clusters candidates (by `sourceId` link, falling back to name) behind an opt-in `promoteDuplicateSources` flag, a new `featureDiff` module reports what actually differs, and `DuplicateSourceGroup.svelte` renders the choice. It also extracts `SourceTag`, `getItemSource`, and `selectionGroupRules` from three copies of the same markup/logic.

The resolver design is sound and the extractions are clean. My concerns are concentrated in the picker component.

## Verified as correct

- **The synthetic group key never persists.** `grantLevelUpFeatures` (`src/documents/actor/character.ts:1601`) iterates `[, features]` and only reads values, so the `duplicate-source:` key cannot reach a real `system.group`. `getClassProgressionData.ts:33` passes `{}`, so the class sheet genuinely stays on the old path.
- **The `'pack'` -> `'compendium'` vocabulary migration is complete** — no stale `data-source='pack'` selectors or comparisons remain anywhere.
- **No XSS.** Diff segments render as `{segment.text}` (escaped) and `toPlainText` strips `<script>`/`<style>` first.
- **Name-fallback clustering cannot misfire on shipped content.** I scanned all 182 class/level auto-grant groups in `packs/classFeatures`: zero duplicate names. Promotion only triggers on real world/pack collisions.
- **`selectionCount: 0` flows correctly** through `classFeaturesComplete` in both dialogs, and `filteredAutoGrant` still strips owned copies from auto-grant while selection groups intentionally keep them.
- **Badge contrast holds in both themes** — the alternate theme only drops the backgrounds to L=30%, so the `--nimble-badge-text-color: white` comment is accurate.

## Issues

### 1. The outcome line is wrong on first render — `DuplicateSourceGroup.svelte.ts:196`

```ts
if (selectedFeatures.length === 0) {
    return localize('NIMBLE.classFeatureSelection.duplicateOutcomeNone');
}
```

`duplicateOutcomeNone` is *"Keeps only the copy you already have."* — but zero selections is the **default state of every fresh duplicate cluster**, where the player owns no copy at all. The most common first thing a player reads about this feature is false.

The intended guard is already written and unused: `hasOwnedCopy` (line 179) is never referenced by the component. Gate on it and add a "nothing selected yet" string, or preselect the recommended copy (see #6). No test covers the no-owned, zero-selection case.

### 2. No way back to "take nothing" — `DuplicateSourceGroup.svelte:159` / `:149`

The radio always sets, never toggles:

```svelte
onclick={() => onSetSelection([candidate.feature])}
```

In the owned case (`selectionCount: 0`, one offerable copy) the Keep-all / Keep-recommended button is hidden by `{#if groupState.offerable.length > 1}`. So once a player clicks the radio to look around, **no control returns them to zero selections** — which is exactly the outcome the resolver went out of its way to make legal ("the floor drops to 0"). The existing toggle-off test drives `handleFeatureSelect` through the harness rather than this picker, so it does not cover this.

Suggest making the radio toggle when `selectionCount === 0`, or adding an explicit "Keep only what I have" control.

### 3. Diffs recompute on every click — `DuplicateSourceGroup.svelte.ts:33-63`

`get candidates()` is a plain getter, and `getProps()` reads *both* `group` and `selectedFeatures`. Every radio click therefore invalidates it and re-runs `compareFeatures` plus the LCS `diffDescription` for every candidate — even though `buildCandidates` only destructures `group`. Descriptions are also diffed eagerly for rows that are still collapsed.

Wrap in `$derived` keyed on `group`, or split the props getter so the candidate build does not depend on the selection.

### 4. Raw schema identifiers leak into localized prose — `DuplicateSourceGroup.svelte.ts:1988-1993`

`changedRuleTypes` are the rules' `type` discriminators verbatim. The PR description's example (`healing`) reads fine, but real values include `damageBonus`, `diceConsumer`, `abilityBonus`, which render as `differs: damageBonus` sitting next to properly localized labels like `action cost`. This needs a label map, falling back to the localized `effects` / `rules` for unrecognized types.

### 5. Range-group support in `FeatureGroupSelection` is unreachable

Both dialogs route `isRangeGroup(group)` to `DuplicateSourceGroup`, and `LevelUpFeatureOptionPicker` never sets `selectionMax`. So the range branches in `buildHintText` / `buildProgressText` never run in production, and **2 of the 31 strings queued for localization review (`duplicateSourceHint`, `nOfMKept`) are dead** — worth telling the localizers before they get translated. The same applies to the range-cap tests in `ClassFeatureSelection.svelte.test.ts`. Either delete the branches or document them as a deliberate fallback.

### 6. `recommendedUuid` doc comment is wrong — `types/components/ClassFeatureSelection.d.ts`

*"Candidate to preselect and mark as recommended"* — nothing preselects it. Actually preselecting would also fix the bad default in #1 and remove a required click in the common case.

## Smaller notes

- **`getDocumentSourceLabel` returns a hardcoded English `'World'`.** Pre-existing, but `describeLineage` now feeds it into `duplicateDerivedFrom` ("from World") when a world item's `sourceId` points at another world item — newly visible inside an otherwise-localized dialog.
- **Accessibility, `DuplicateSourceGroup.svelte:57-77`:** `<ul role="radiogroup">` with `<li>` children (listitem is not a valid radiogroup child); each row is `role="button"` and *contains* the radio `<button>` (nested interactive content); there is no roving tabindex or arrow-key navigation between radios; and the collapsed body uses `0fr` + `overflow: hidden`, so it stays in the accessibility tree while `aria-expanded="false"`.
- **`SelectionIndicator.svelte:21`:** `aria-checked={role ? ariaChecked : undefined}` emits nothing if a role is set without `ariaChecked`, producing an invalid radio. Default it to `false`.
- **`toPlainText` decodes `&amp;` before `&lt;`/`&gt;`,** so `&amp;lt;` double-decodes to `<`. Harmless since the result renders as text, but wrong; decode in a single pass, and consider numeric entities beyond `&#39;`.
- **`describeOrigin` uses `folder?.name` (leaf only).** Two world copies under different parents that share a leaf folder name remain indistinguishable — the exact case this design targets. Folder ancestry path would close it.
- **Badge font-size** moved from a hardcoded `0.5625rem` to `--nimble-xxs-text` (`0.579rem`), a small unannounced visual change on the class sheet.
- **`showSourceLabel` on a named group badges every candidate,** including ones that are not duplicated (the test asserts 3 badges where only 2 collide). Minor noise.
- **Test naming:** "flags a copy that is byte-identical" — `compareFeatures` normalizes whitespace and key order, so it is not byte-identical.

## Test coverage

Strong on the resolver (owned copies, cluster merging, blank names and sources, opt-in gating) and on `featureDiff`. The gap is the picker's own state machine: both blocking bugs above live in paths the component tests do not exercise, since the toggle-off test targets the harness factory instead. The PR's own note that manual in-Foundry verification is still outstanding is well placed.

Requesting changes for #1 and #2; #3-#6 are worth doing but would not block on their own.

trevlar:
## Fixed

- **Outcome line lied on a fresh cluster** — zero selections now splits on `hasOwnedCopy`: with a copy on the sheet it still reads "keeps only the copy you already have", without one it reads a new `duplicateOutcomePending` string, and clusters that must keep a copy open on their recommended one instead of on nothing (`DuplicateSourceGroup.svelte.ts:249`).
- **No route back to keeping nothing** — the radio toggles off when the group's floor is zero, so clicking the chosen copy again hands the selection back (`DuplicateSourceGroup.svelte:56`).
- **Diffs recomputed on every click** — the state factory takes the cluster and the selection as two separate getters, and the candidate build is `$derived.by` keyed on the cluster alone, so a click no longer re-runs an LCS per candidate (`DuplicateSourceGroup.svelte.ts:181`).
- **Raw schema identifiers in localized prose** — `damageBonus` and friends resolve through `CONFIG.NIMBLE.ruleTypes` / `effectTypes`, the same labels the Rules Builder shows, falling back to the localized field name when a type has none (`DuplicateSourceGroup.svelte.ts:107`).
- **Unreachable range-group branches** — the range paths in `buildHintText` / `buildProgressText` are gone, along with `duplicateSourceHint` and `nOfMKept`, so nothing dead reaches the localizers (`FeatureGroupSelection.svelte.ts`).
- **`recommendedUuid` doc comment** — now says what it does: marks the recommended copy, and starts the group there when a copy has to be kept (`ClassFeatureSelection.d.ts:35`).
- **Hardcoded English `'World'`** — `getDocumentSourceLabel` reads the source badge's own localized labels for both branches (`getDocumentSourceLabel.ts:6`).
- **Radiogroup structure and keyboard access** — rows are `<div>`s under the radiogroup rather than `<li>`s, the expander and the radio are siblings instead of nested controls, the group holds one tab stop with arrow keys carrying the selection, and a collapsed body is `inert` so `aria-expanded="false"` stops lying (`DuplicateSourceGroup.svelte:117`).
- **`aria-checked` absent on a role'd indicator** — defaults to `false` (`SelectionIndicator.svelte:12`).
- **`&amp;lt;` double-decoded** — one regex pass over named and numeric references, so a decoded `&` is never re-read as the start of an entity; numeric and hex forms included, unrecognized ones left alone (`featureDiff.ts:161`).
- **Leaf-only folder name** — world origins carry their folder path, outermost first, capped at four levels (`DuplicateSourceGroup.svelte.ts:44`).
- **Badge on every candidate of a named group** — `showSourceLabel` became `duplicatedSourceUuids`, so only candidates that actually collide get badged (`getClassFeatures.ts:484`).
- **"byte-identical" test name** — renamed to what `compareFeatures` actually checks, every compared field matching (`DuplicateSourceGroup.svelte.test.ts:221`).
- **Badge font-size** — the token stands, with a comment recording that the three copies it replaces predated the scale variable (`SourceTag.svelte:30`).

## Not fixed

- Nothing.

## Approach

The two blocking bugs share one root: the component had no notion of whether keeping nothing was a legal outcome, so it could neither describe zero selections honestly nor return to it. `canKeepNone` (derived from `selectionCount === 0`) now drives both the outcome line and the radio's toggle, which is why preselecting the recommended copy is safe — it only happens when the floor is above zero, leaving the owned case genuinely empty rather than choosing for the player.

On #5 I deleted the range branches rather than documenting them. `LevelUpFeatureOptionPicker` never sets `selectionMax` and both dialogs route range groups to the duplicate picker, so a documented fallback would have been a comment guarding code with no path to it.

Test coverage moved onto the picker itself: the no-owned zero-selection case and the toggle-back-to-zero case are now driven through the component (`DuplicateSourceGroup.svelte.test.ts:211`, `:260`), alongside the arrow-key and single-tab-stop behaviour. Manual in-Foundry verification is still outstanding.

Pushed as `a0b556b`. Thanks for the review — the `hasOwnedCopy`-is-written-but-unused catch was the useful thread to pull.

Fronix (approved):
LGTM

PR #855, opened 2026-07-20
feat(rules): damage resistance mode, monster defenses, and banked reduction expiry

Fronix:
## Summary

Builds the two remaining actionable follow-ups from #851: a half-damage (resistance) mode for the damage-reduction pipeline, and expiry for banked one-shot reductions. Monster damage resistances/immunities go from inert schema fields to fully wired and editable, with visibility on the monster sheet and a per-target "Reduced Damage" breakdown on attack cards so the GM can always tell why a target takes less than the roll.

## Changes

- `damageReduction` rule gains a **Mode** field: `flat` (default, unchanged) subtracts the value; `half` halves incoming damage instead (Core Rules glossary resistance). The value field hides when `half` is selected.
- Monster `attributes.damageResistances` (half damage) and `damageImmunities` (no damage) are now honored by the damage pipeline and editable as damage-type tag groups in the monster metadata config dialog. `damageVulnerabilities` is editable but flagged not-yet-automated.
- Application order: outcome/armor halving → immunity (zero) → resistance halving (applies once regardless of how many sources match, rounds up like heavy armor) → flat reductions + banked reduction → temp HP. The Apply Damage preview reflects all of it.
- Banked one-shot reductions expire at combat end (active-GM-gated, dual-signal dedup like the encounter-end dice trigger) and are no longer consumed when immunity or armor already zeroes the hit.
- Attack cards: GM-only **Reduced Damage** block under the target list, listing per target every applicable modifier (immunity, resistance, labeled flat reductions, banked amount). Reduction entries now carry their rule's label for attribution.
- Monster sheets: damage defenses shown in the Core tab's defenses section under saves/armor, one wrapping row each for Resistant / Immune / Vulnerable; clicking opens the config dialog. Fixes included: override Foundry core's fixed button height (was crushing multi-row content) and stop tab sections from flex-shrinking (body scrolls instead of compressing).
- Docs: system reference (`docs/system/rules.md`) and user docs (monster pages) updated.

## Test Plan

1. Open a monster's sheet, click the edit button next to the metadata line, and set some damage resistances/immunities/vulnerabilities. Confirm they appear as rows under Saves on the Core tab and that clicking those rows reopens the dialog.
2. Attack that monster with a typed damage roll (e.g. fire) matching a resistance. Confirm the target list's damage preview shows the halved number, the "Reduced Damage" block explains it, and Apply Damage applies the halved amount.
3. Repeat with a matching immunity: preview shows 0 and nothing is applied.
4. Add a Damage Reduction rule with Mode "Half" to an item on a character and confirm the same halving behavior, with the rule's label shown in the card breakdown.
5. Berserker "That all you got?!": spend Fury Dice to bank a reduction, then end the combat without applying damage. Confirm the "Damage Reduction (N)" effect is removed at combat end. Bank again, attack with a type the actor is immune to, apply: the bank must survive.
6. Check the monster sheet in both light and dark mode at the default sheet size with all 13 damage types set as resistances (rows wrap, nothing overlaps).
7. Check the card's "Reduced Damage" block attributes every line: a reduction rule with a label shows the label, one without shows the owning item's name, and a Fury Dice spend shows the feature's name (no "banked" wording anywhere). Reopen the sheet or re-target repeatedly and confirm no line ever appears twice.

## Checklist

- [x] Tested in Foundry with no console errors (sheet UI, config dialog, and layout live-verified against Foundry v14 via Playwright; damage application and banked expiry are unit-tested - please run the test plan)

## Related Issues

Follow-ups from #851. Armor of Shadows wiring stays deferred pending summoning logic; the `effectType` registry stays deferred until a second effect type exists.


Fronix:
Added `d0df9196` after live testing:

- **Reduced Damage attribution**: every line in the card's breakdown now names its source. Reduction rules with no label fall back to the owning item's name, and dice-spend banked reductions record the banking feature's name ("That all you got?! (-8)"). The word "banked" no longer appears anywhere user-facing.
- **Accumulator duplication fix**: rule accumulator arrays (`damageReductions`, `damageBonuses`) duplicated entries whenever Foundry called `prepareData()` directly without re-initializing the document (same `system` object reused). Rules now register accumulator paths in a shared registry and the actor resets them at the start of each prepare cycle. Note for reviewers: the `damageBonuses` half of this predates this branch — damage bonuses have been silently doubling on those code paths; it only became visible once the breakdown UI listed the entries.

trevlar (changes requested):
**🔧 Push back — one hardcoded tooltip breaks the i18n convention you're following everywhere else**

This is a strong follow-up to #851 — the damage pipeline changes are clean, the ordering is well-reasoned and documented, and the test coverage is genuinely good (immunity, resistance stacking, bank-survives-immunity, expiry dedup across both combat-end signals). Just one thing to fix before merge.

**Must fix before merge**
- **src/view/sheets/pages/NPCCoreTab.svelte:519** — `data-tooltip="Edit"` is a hardcoded user-facing string, sitting directly between tooltips that use `game.i18n.localize(...)` (lines 495, 507). It also contradicts your checklist ("No hardcoded user-facing strings"), and with Babele in play an untranslated literal leaks to every locale. Swap it for a localized key — `NIMBLE.skillsConfig.edit` already resolves to "Edit", or add a dedicated `NIMBLE.npcSheet.editMetadata`.

**What's good**
- The damage-application order (immunity zeroes → resistance halves → flat/banked subtract) is sound, and halving-once-regardless-of-source-count is correctly enforced and tested. `Math.ceil` matches the existing heavy-armor halving convention.
- Bank consumption reworked so immunity/armor-zeroed hits leave the bank intact — verified by `does not consume a banked reduction when immunity already zeroes the hit`. The expiry hook mirrors `encounterEndTrigger`'s dual-signal dedup pattern and adds appropriate active-GM gating.

- **src/documents/chatMessage.ts:280** — `calculateAdjustedDamage` still accepts a `bankedReduction?` param, but `buildDamageApplicationPlan` no longer passes it (banked is now subtracted afterward). The param and the `+ (params.bankedReduction ?? 0)` line are dead — dropping them would remove a misleading second banked-subtraction path.
- **src/documents/actor/base.svelte.ts:276** — the accumulator-reset in `prepareData()` is the load-bearing fix for duplication, but it has no direct actor-level test (only path registration is tested in `damageReduction.test.ts`). A small regression test asserting a double `prepareData()` doesn't duplicate `damageReductions` would lock the behavior in.

---

1. Localize `data-tooltip="Edit"` in NPCCoreTab
2. Drop dead `bankedReduction` param from `calculateAdjustedDamage`
3. Add a test for the prepareData accumulator reset

PR #856, opened 2026-07-20
feat(rules): builder-editable suppressActivationCard envelope field

Fronix:
## Summary

Makes activation-card suppression a rule-agnostic, builder-editable setting instead of a code-level hook, per the deferred follow-up from #851. Any rule on any item can now declare, from the Rules Builder's Advanced section, whether using the item should skip its descriptive chat card.

## Changes

- New tri-state `suppressActivationCard` envelope field (`auto` / `always` / `never`, initial `auto`) on the base rule schema, so every rule type gets it with no migration needed.
- `suppressesActivationCard()` resolves the field first; `auto` defers to a new protected `_autoSuppressesActivationCard()` hook. `diceConsumer`'s automatic spend-flow suppression moved into that hook, so its behavior is unchanged but can now be opted out per rule.
- Semantics: each rule resolves independently and the card is suppressed when any enabled rule resolves true. `Never` only stops that rule from suppressing; it does not veto other rules. The item-side guard is unchanged: a card carrying rolls or effect nodes is never suppressed.
- Rules Builder: localized "Suppress activation card" select in the rule card's Advanced section, next to "Applies when". The field is added to the envelope list in all three synced locations (RuleCard `FIXED_FIELDS`, docs generator `BASE_RULE_FIELDS`, `docs/system/rules.md`).
- Docs: envelope-field section and pool-consumption section in `docs/system/rules.md`; user docs in Rules Builder Basics and the generated reference's shared-fields tip.

## Test Plan

1. Open any item's Rules Builder, expand a rule's **Advanced** section, and confirm the "Suppress activation card" select shows Automatic / Always / Never with Automatic selected by default.
2. On an item with only a passive rule (e.g. a Stat Bonus) and no rolls/effects, set the field to **Always** and activate the item: no chat card should post. Set it back to Automatic: the card posts again.
3. On a Berserker with "That all you got?!" (manual dice consumer), activate the feature: the activation card is suppressed automatically and the spend panel opens (unchanged behavior). Set the consumer rule's field to **Never** and activate again: the descriptive card posts alongside the spend flow.
4. Give an item a damage roll and set a rule to Always: the card must still post (cards with rolls or effects are never suppressed).
5. Confirm existing items without the field saved behave exactly as before (defaults to Automatic).

## Checklist

- [x] Tested in Foundry with no console errors (unit-tested only; please run the test plan)

## Related Issues

Deferred follow-up from #851


trevlar (changes requested):
**🔧 Push back — the PR description doesn't match the implementation for `Always`**

The refactor is clean — tri-state envelope field, `auto` deferring to a protected `_autoSuppressesActivationCard()` hook, `diceConsumer` moved onto that hook with its behavior preserved, and the three-location field sync all updated. Nice, well-documented work. One reconciliation needed before merge, plus a test gap.

**Must fix before merge**

- **src/documents/item/base.svelte.ts:173** — the docs and test plan promise that **Always** silences a card-less activation unconditionally, but the code doesn't do that under default settings. `_shouldSuppressActivationCard` returns early on `if (!isAutoApplyEnabled()) return false` before it ever consults the rules, and that setting (`automation.autoApplyConditions`) defaults to **`false`** (`src/settings/index.ts:97`). So on a fresh world, a rule set to `Always` still posts its card — test-plan step 2 fails unless the GM first enables auto-apply, which the step doesn't mention.

  Note this gate is *correct* for the `auto` path: a `diceConsumer` only suppresses because its spend flow posts a replacement card, and that only fires when automation is on — so with automation off you genuinely want the normal card. `always` is different: it has no replacement flow, so sharing the gate is what breaks it. Reconcile one of two ways:
  - **Code fix (matches the docs):** let an enabled `always` rule suppress regardless of automation; keep the `isAutoApplyEnabled()` gate only for the auto-resolved path.
  - **Docs fix (if suppression is meant to be automation-gated across the board):** update `docs/system/rules.md` ("silence a card-less activation outright"), the Rules Builder Basics entry, and test-plan step 2 to state that `Always` requires "Auto-Apply Conditions from Rules" enabled.

- **src/documents/item/base.svelte.ts:164** — the item-level resolution in `_shouldSuppressActivationCard` (the automation gate + the `!rule.disabled && …some()` fold) has no test. The new unit tests all exercise `suppressesActivationCard()` on the rule in isolation, where the behavior is correct — the discrepancy above lives one level up, untested. Add item-level tests: automation off + `always`, `always` on a rule with rolls present (must still post), and an enabled `always` rule alongside a disabled one.

**What's good**
- The `auto` → protected `_autoSuppressesActivationCard()` split is the right shape, and moving `diceConsumer`'s override onto it keeps behavior identical while making it opt-out-able. Rule-level tests cover always/never/auto for both the base and the consumer.
- The envelope-field sync (`FIXED_FIELDS`, `BASE_RULE_FIELDS`, `docs/system/rules.md`) hits all three locations plus a note telling future editors they must stay in sync.

CI is green.

---

1. Reconcile `Always` behavior with the docs/test-plan (code or docs)
2. Add item-level `_shouldSuppressActivationCard` tests

PR #857, opened 2026-07-22
feat(rules): add skillRollMode rule type for skill check advantage

### LINKED ISSUE #844: [Foundation] Rule type: skillRollMode
Set the roll mode (advantage / disadvantage) on one or more skill checks, gated by the inherited `predicate` field. This lets features like "you have advantage on Stealth checks" be authored in the Rules Builder **without hand-writing dice syntax** — advantage in Nimble is a *level*, not a fixed `kh`/`kl` expression, and can stack past 1 (CoreRules-2 p.256: "If you have multiple instances of advantage or disadvantage, for each one, roll an extra die and remove the lowest").

Today there is no way to express this. `skillBonus` only adds a numeric/dice value to `skills.<skill>.bonus` — it cannot touch the roll mode. Roll-mode rules exist for saving throws (`savingThrowRollMode`) and initiative (`initiativeRollMode`), but there is **no `skillRollMode`**.

```jsonc
{
  "type": "skillRollMode",
  "label": "Professional Skulker",
  "skills": ["stealth"],   // one or more skills, or "all"
  "value": 1,              // signed integer: +N advantage, -N disadvantage
  "mode": "adjust"         // "adjust" (stack with other sources) or "set"
}
```

## Shape to mirror

Follow `savingThrowRollMode` (`src/models/rules/savingThrowRollMode.ts`) and reuse `skillBonus`'s skill-selection field:

- `skills`: `ArrayField<StringField>` with `choices: [...Object.keys(CONFIG.NIMBLE.skills), 'all']`, exactly like `skillBonus` (`src/models/rules/skillBonus.ts:19-34`). `"all"` is the runtime sentinel meaning every skill.
- `value`: signed `NumberField` — positive = advantage levels, negative = disadvantage levels. This maps directly onto the roll engine, which already models roll mode as a signed integer (`src/dice/constructD20RollFormula.ts:16` — *"positive for advantage, negative for disadvantage, 0 for normal"*). **No dice syntax anywhere.**
- `mode`: `"adjust"` (default, sums with the character's existing skill roll mode and other rules — so two sources of advantage give advantage 2) or `"set"` (overrides).
- Register in both `ruleTypes` and `ruleDataModels` in `src/config/registerRulesConfig.ts`, add `skillRollMode` to `RULE_ICONS` in `RuleTypePicker.svelte.ts`, add `NIMBLE.ruleTypes.skillRollMode` + `NIMBLE.ruleDescriptions.skillRollMode` (+ field labels/hints) to `en.json`.
- `static group = 'bonuses'`.

## Where it applies

`savingThrowRollMode` deliberately does **not** mutate during data prep — the default roll mode is a persisted, user-editable value, and rule contributions are recomputed on demand (see its `prePrepareData` note). Decide up front whether `skillRollMode` should:

1. follow that pattern (surface a "Reset to class defaults" contribution the skill config dialog consumes), or
2. contribute live during data prep to a `skills.<skill>.rollMode` accumulator that the skill-check activation site reads when building the d20 formula.

Option 2 is simpler for "passive, always-on" features (which is what Stealth advantage usually is) and needs a `skills.<skill>.rollMode` field on the character skill schema. Trace where a skill check builds its d20 formula and feed the accumulated roll mode into `constructD20RollFormula`'s `rollMode` arg. Worth a quick look at how saving throws thread their roll mode through before committing.

## Acceptance criteria

- A rule `{ type: "skillRollMode", skills: ["stealth"], value: 1 }` on any embedded item causes Stealth checks to roll with advantage 1.
- `value: 2` (or two separate advantage-1 rules with `mode: "adjust"`) rolls with advantage 2 — an extra die each, per the stacking rule.
- Negative `value` applies disadvantage; advantage and disadvantage from different sources cancel per instance (p.256).
- `skills: ["all"]` applies to every skill; `predicate` gates the rule like every other rule.
- Authorable end-to-end in the Rules Builder with a skill multi-select and a plain signed-number field — no formula/dice string required.

## Features this unblocks

- The Cheat: Professional Skulker, Leave No Trace (Stealth advantage)
- Hunter: Keen Sight / Skilled Tracker (Perception / tracking advantage)
- Backgrounds & ancestries granting advantage on a specific skill
- Any homebrew "advantage on <skill>" feature

---

Parent: #681

Fronix:
## Summary

Adds the `skillRollMode` rule type (#844): features can now grant stacking advantage/disadvantage levels on skill checks via the Rules Builder, with no dice syntax. Along the way, fixes a latent bug where seven existing rule types silently ignored their `predicate` field.

skillRollMode-roll-dialog
skillRollMode-rules-builder

## Changes
- New `skillRollMode` rule: `skills` multi-select (with `"all"` sentinel), signed integer `value` (+N advantage / −N disadvantage), and `mode` (`adjust` stacks with other sources, `set` overrides). Applies live during data prep, mirroring `initiativeRollMode`'s reset-then-apply pattern on `skills.<skill>.defaultRollMode`.
- Ships **Leave No Trace** (The Cheat / Tools of the Silent Blade) as the first consumer: Stealth advantage gated on the `self:fullHp` tag.
- Fixes predicate gating in seven rules that mutated actor data without calling `this.test()`: `initiativeRollMode`, `skillBonus`, `abilityBonus`, `healingPotionBonus`, `maxHitDice`, `maxWounds`, `savingThrowBonus`. No shipped pack content carries predicates on these types, so existing compendium behavior is unchanged.
- Known limitation (documented in the sweep commit): these seven apply in `prePrepareData`, before late tags like `self:fullHp` are populated, so predicates on late tags won't match for them. Follow-up branch `feature/predicate-late-tags` addresses this.

## Test Plan

1. Create a character and give it any feature item. Open the item's Rules tab, then Rules Builder, and add a **Skill Roll Mode** rule: Apply to `stealth`, Roll mode value `1`, How to apply `adjust`.
2. Roll a Stealth check from the character sheet. The roll dialog slider should seed at **Advantage × 1** and the formula should read `2d20khn`.
3. Add a second identical rule (or set value to `2`). The dialog should seed at **Advantage × 2** (`3d20khn`).
4. Set a negative value. The slider should seed on the disadvantage side. Disable or delete the item, and Stealth should return to a straight roll.
5. Drag **Leave No Trace** from the class features compendium onto a character with full HP. Stealth seeds at Advantage × 1. Damage the character and the advantage disappears. Heal to full and it returns.
6. Predicate regression: add a rule of any fixed type (e.g. Skill Bonus) with an impossible predicate (key `self`, value `nonexistent`). The bonus must **not** apply. Clear the predicate and it applies.

## Checklist

## Related Issues

Closes #844

### DIFF
--- a/docs/documentation/reference/_partials/skillRollMode.md
+++ b/docs/documentation/reference/_partials/skillRollMode.md
@@ -6,3 +6,5 @@
 - **How to apply** → `adjust`
 
 Advantage stacks: two separate advantage rules (or a value of `2`) roll two extra dice. Negative values apply disadvantage instead, and advantage and disadvantage from different sources cancel each other one for one. Use `set` to override other sources entirely.
+
+When mixing `set` and `adjust` rules on the same skill, order matters. Rules run in priority order, so a `set` rule that runs after an `adjust` rule discards that adjustment, while a `set` that runs before it composes normally. Use the rule priority to control which one wins.
--- a/docs/documentation/reference/_partials/skillRollMode.md
+++ b/docs/documentation/reference/_partials/skillRollMode.md
@@ -1,0 +1,8 @@
+**Example: Professional Skulker.** A feature that grants advantage on Stealth checks.
+
+- **Label** → `Professional Skulker`
+- **Roll mode value** → `1`
+- **Apply to** → `Stealth`
+- **How to apply** → `adjust`
+
+Advantage stacks: two separate advantage rules (or a value of `2`) roll two extra dice. Negative values apply disadvantage instead, and advantage and disadvantage from different sources cancel each other one for one. Use `set` to override other sources entirely.
--- a/packs/classFeatures/core/the-cheat/the-cheat-subclasses/tools-of-the-silent-blade/leave-no-trace.json
+++ b/packs/classFeatures/core/the-cheat/the-cheat-subclasses/tools-of-the-silent-blade/leave-no-trace.json
@@ -6,7 +6,19 @@
 	"system": {
 		"macro": "",
 		"identifier": "",
-		"rules": [],
+		"rules": [
+			{
+				"type": "skillRollMode",
+				"label": "Leave No Trace",
+				"skills": ["stealth"],
+				"value": 1,
+				"mode": "adjust",
+				"predicate": {
+					"self": "fullHp"
+				},
+				"id": "lyHWwQGcbooGei1a"
+			}
+		],
 		"activation": {
 			"acquireTargetsFromTemplate": false,
 			"cost": {
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -662,6 +662,7 @@
 			"savingThrowBonus": "Save Bonus",
 			"savingThrowRollMode": "Save Roll Mode",
 			"skillBonus": "Skill Bonus",
+			"skillRollMode": "Skill Roll Mode",
 			"speedBonus": "Speed Bonus",
 			"toggleEffect": "Toggle Effect",
 			"unarmedDamage": "Unarmed Damage"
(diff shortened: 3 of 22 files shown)


trevlar (approved):
**✅ Approved**

This adds the `skillRollMode` rule type cleanly — it mirrors `initiativeRollMode`'s reset-then-apply pattern in `afterPrepareData`, and I traced the ordering to confirm it's correct: the `skill.defaultRollMode = 0` reset and the `self:fullHp` tag are both set in `character.prepareDerivedData` before `afterPrepareData` hooks fire, so the rule reads a clean baseline and the Leave No Trace predicate resolves. `rollSkillCheck` consumes `defaultRollMode`, so it surfaces in the dialog as intended.

- Strong test coverage — stacking, set-vs-adjust, negative values, the `"all"` sentinel, disabled, failing predicate, and the no-skills actor edge case are all covered.
- The predicate fix is genuinely worth having, and the regression claim holds: I scanned every pack and confirmed **no** shipped content carries a predicate on any of the seven affected rule types, so honoring predicates changes no existing compendium behavior. The `set`/`adjust` ordering caveat is documented in both the lang hint and the docs partial, and the late-tag limitation for the `prePrepareData` rules is called out honestly.

PR #858, opened 2026-07-22
fix(rules): make self:fullHp an early tag and guard late-tag predicates

Fronix:
> [!IMPORTANT]
> Dont merge before #857 is merged to dev (it will auto switch to dev)

## Summary

Eliminates the "authorable but silently broken" predicate trap: tags populated late in data prep were invisible to rules applying in `prePrepareData`, so a correctly-configured predicate could preview green in the Rules Builder yet never match at runtime. `self:fullHp` is now a genuinely early tag that works with every rule type, and the one remaining inherently-late family (`<ability>:<mod>`) is loudly guarded at both authoring time and runtime.

lateTag-builder-warning

## Changes

- **`self:fullHp` is now an early tag.** Character `hp.max` is computed in a new `_prepareEarlyDerivedData()` hook before tag population, so the tag is added uniformly in the base actor class for all four actor types and is visible to `prePrepareData` rule predicates (this is what Leave No Trace-style features gate on). The four duplicated per-subclass tag blocks are removed. Side effect: fixes a latent staleness bug where the bloodied HP-threshold failsafe read the previous cycle's `hp.max` for characters.
- **Guardrails for `<ability>:<mod>` predicates on early rules** (these cannot be made early — `abilityBonus` rules feed the mods during `prePrepareData`):
  - `CONFIG.NIMBLE.LATE_PREDICATE_KEYS` registry derived from the ability score config.
  - `NimbleBaseRule.appliesInPrePrepareData` introspects each rule class for a `prePrepareData` implementation — no per-rule metadata to maintain. `savingThrowRollMode`'s comment-only no-op hook is removed so it reports correctly.
  - `Predicate.extractReferencedKeys()` walks a raw predicate (including `$and`/`$or` atoms and nested sub-predicates).
  - The Rules Builder predicate editor shows an amber warning banner **instead of** a false green "matches" preview when an early-phase rule references a late key — including on unowned/compendium items where no preview domain exists.
  - Predicate evaluation emits a once-per-rule-content `console.warn` for the same misconfiguration (checked at `test()` time so transient pre-creation instances stay silent).
- Docs updated (`docs/system/rules.md` tag-timing section, Rules Builder predicates page), including removing a bogus `constitution:2` example.

## Test Plan

1. Create a character, give it a class, and set HP to max. Add a feature item with a **Skill Bonus** rule (Bonus `2`, Apply to `stealth`) and condition `self` is `fullHp`. The Stealth modifier on the sheet should include the +2. Damage the character: the bonus disappears. Heal to full: it returns. (Before this PR the bonus never applied.)
2. Drag **Leave No Trace** onto a full-HP character: Stealth advantage applies as before (regression check).
3. Set a character's HP to exactly half without toggling any condition: the bloodied state should be reflected immediately.
4. Open the Rules Builder on an item with an **Ability Bonus** rule and add condition `strength` is at least `2`: an amber warning banner appears instead of the green/red preview, including on items not owned by any actor. The same condition on a **Damage Bonus** rule shows the normal preview.
5. With that misconfigured rule on a character, check the console: exactly one warning naming the rule and the offending key, with no repeats as the actor updates.
6. Drop the **Celestial** ancestry on a character: no console warnings (its ability-score predicate sits on an after-phase rule), and it still grants the Celestial language.

## Checklist

## Related Issues

Follow-up to #857 (stacked on `feature/844-skillRollMode`; retarget to `dev` once that merges).

### DIFF
--- a/docs/documentation/reference/_partials/skillRollMode.md
+++ b/docs/documentation/reference/_partials/skillRollMode.md
@@ -6,3 +6,5 @@
 - **How to apply** → `adjust`
 
 Advantage stacks: two separate advantage rules (or a value of `2`) roll two extra dice. Negative values apply disadvantage instead, and advantage and disadvantage from different sources cancel each other one for one. Use `set` to override other sources entirely.
+
+When mixing `set` and `adjust` rules on the same skill, order matters. Rules run in priority order, so a `set` rule that runs after an `adjust` rule discards that adjustment, while a `set` that runs before it composes normally. Use the rule priority to control which one wins.
--- a/docs/documentation/reference/_partials/skillRollMode.md
+++ b/docs/documentation/reference/_partials/skillRollMode.md
@@ -1,0 +1,8 @@
+**Example: Professional Skulker.** A feature that grants advantage on Stealth checks.
+
+- **Label** → `Professional Skulker`
+- **Roll mode value** → `1`
+- **Apply to** → `Stealth`
+- **How to apply** → `adjust`
+
+Advantage stacks: two separate advantage rules (or a value of `2`) roll two extra dice. Negative values apply disadvantage instead, and advantage and disadvantage from different sources cancel each other one for one. Use `set` to override other sources entirely.
--- a/docs/documentation/rules-builder/predicates.md
+++ b/docs/documentation/rules-builder/predicates.md
@@ -26,7 +26,7 @@ These are the tag families that exist:
 | `background:soldier` | The character's background, by identifier. |
 | `armor:equipped` / `armor:unarmored` | Whether the character is wearing anything that actually provides armor. Exactly one of these is always present. |
 | `self:shield` / `self:noShield` | Whether the character carries a shield. |
-| `strength:3`, `dexterity:1`, `constitution:2`, `intelligence:0`, `will:4` | The character's current stat modifiers. |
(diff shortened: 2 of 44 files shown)


trevlar (approved):
**✅ Approved**

Solid fix for the "authorable but silently broken" predicate trap. Making `self:fullHp` a genuinely early tag (via `_prepareEarlyDerivedData()` before tag population) and loudly guarding the inherently-late `<ability>:<mod>` family at both authoring and runtime is exactly the right split. I traced the actor-prep reorder and couldn't find a regression.

- The `_prepareHitPoints` move is safe: it reads only class `maxHp` and the stored `hp.bonus` (written by `maxHpBonus` in persistence hooks, not during data-prep), so computing `hp.max` before `prePrepareData` rules can't go stale — and it genuinely fixes the bloodied HP-threshold failsafe reading last cycle's `hp.max`.
- `appliesInPrePrepareData` introspecting the prototype (no per-rule metadata) plus the data-aware `appliesInPrePrepareDataFor` override for dual-phase numeric/formula speeds is a clean design, and the dedupe key on stable content rather than the regenerating `id` is the right call.
- Test coverage is thorough across every new path — key extraction, phase detection, the once-per-content warning, and the builder banner (including the no-preview-domain case).

Issue #683, opened 2026-04-08
[EPIC] Berserker automation


Fronix:
All Berserker class automation — Rage/Fury Dice framework, Savage Arsenal abilities, and subclass features (Mountainheart, Red Mist).

Parent: #365

Fronix:
Status after the berserker-automation PR #861 (single branch, engine work + pack data + Migration032):

### Fully automated (16 of 31 features)

| Feature | Mechanism |
|---|---|
| Rage + Fury Dice framework | toggleEffect + dicePool + autoBonus diceConsumer (pre-existing, #595) |
| That All You Got?! | manual diceConsumer, banked damage reduction (pre-existing, #596) |
| Intensifying Fury (both halves) | modifyPool die ladder + NEW addRefills turn-start entry gated on self:raging |
| Rage (2), Deep Rage | pre-existing |
| Bloodlust | manual diceConsumer, DEX x dice movement total |
| Death Blow | manual diceConsumer, 2 x sum bonus damage |
| MORE BLOOD! | NEW onCritReceived dice-refill trigger |
| One with the Ancients | 1/Safe Rest charge gate |
| Boundless Rage (die floor) | NEW modifyPool.minFace |
| Enduring Rage | dyingActionLimit + NEW modifyToggle.turnOn auto-Rage while Dying |
| Mighty Endurance, Eager for Battle | pre-existing |
| Stone's Resilience | NEW modifyConsumer (+@sum on DR spends) |
| Mountain's Endurance | pre-existing (#859) |
| Blood Frenzy | NEW diceConsumer selectionOutcome `maximize` (player picks the die) + 1/turn charge gate |
| Savage Awareness (Perception) | skillRollMode |
| Onslaught (+2 speed) | speedBonus + self:raging |

### Partial (gate/reminder shipped, effect manual)

Swift Fury (NEW poolGainMessage reminder), Deathless Rage (1/turn gate), Unbreakable (1/encounter gate), Whirlwind, You're Next!, Thunderous Steps, Into the Fray, Titan's Fury (crit half automated, miss half manual), Unstoppable Brutality.

### New engine primitives shipped (all generic, reusable by other classes)

- Turn-boundary dice-pool refill triggers (`onTurnStart`/`onTurnEnd`) + charge `onTurnStart` recovery
- `onCritReceived` refill trigger and the defender-side `onAttackReceived` rule dispatch
- Refill-entry predicates and `modifyPool.addRefills`
[...]

Issue #602, opened 2026-04-08
[Automation] Berserker - Path of the Red Mist subclass checklist


Fronix:
## Context
Checklist for all features of the **Path of the Red Mist** Berserker subclass.

## Features

- [x] **Blood Frenzy** (L3) — "(1/turn) While Raging, whenever you crit or kill an enemy, change 1 Fury Die to the maximum." → `resourceCounter` (1/turn), `conditionalBonus` (Raging), `triggerOnEvent` (onCrit, onKill), dice-pool mutation (set one die to max face) — blocked by #576, #578, #579, #584 — ✅ #861: clicking the feature opens the Fury Dice panel with Blood Frenzy selected and the player picks which die to raise to its maximum (`diceConsumer` with `selectionOutcome: "maximize"`, cost 1). Hard-gated to 1/turn by a charge pool that refreshes at turn start. The crit-or-kill trigger is a card reminder.
- [x] **Savage Awareness** (L3) — "Advantage on Perception checks to notice or track down blood. Blindsight 2 while Raging: you ignore the Blinded condition and can see through darkness and Invisibility within that Range." → `skillBonus` (narrow), `conditionalBonus` (Raging) + `conditionImmunity` (Blinded), vision mode grant — blocked by #579, #585 — ✅ partial, #861: Perception advantage pre-set via `skillRollMode`. Blindsight 2 deferred: no vision-mode rule type exists.
- [ ] **Unstoppable Brutality** (L7) — "While Raging, you may gain 1 Wound to reroll any attack or save." → `conditionalBonus` (Raging), Wound-cost reroll mechanic — blocked by #579
- [ ] **Opportunistic Frenzy** (L11) — "While Raging, you can make opportunity attacks without disadvantage, and you may make them whenever an enemy enters your melee weapon's reach." → `conditionalBonus` (Raging), OA trigger-expansion + disadvantage removal — blocked by #579, #584
- [x] **Onslaught** (L15) — "While Raging, gain +2 speed. (1/round) you may move for free." → `conditionalBonus` (Raging) + `speedBonus` (✅ wired), `resourceCounter` (1/round), `grantAction` (free move) — blocked by #576, #579, #582 — ✅ partial, #861: +2 speed while Raging automated (`speedBonus` with `self:raging` predicate). The (1/round) free move stays in the description.

## Foundations This Checklist Depends On
#576, #578, #579, #582, #584, #585, plus the Berserker Rage framework issue

Part of #365

[...]

Issue #601, opened 2026-04-08
[Automation] Berserker - Path of the Mountainheart subclass checklist


Fronix:
## Context
Checklist for all features of the **Path of the Mountainheart** Berserker subclass.

## Features

- [x] **Stone's Resilience** (L3) — "Whenever you expend Fury Dice to reduce incoming damage, add the value of the die to the amount reduced." → `triggerOnEvent` (onFuryDiceSpentForDR), `damageReduction` augmentation (adds rolled die value on top of That All You Got?! reduction) — blocked by #580, #584, That All You Got?! — ✅ #861: new `modifyConsumer` rule appends `@sum` (the spent die values) to Fury damage-reduction spends, so That All You Got?! reduces `(STR+DEX) × dice + sum of faces` automatically.
- [ ] **Mountainous Tenacity** (L3) — "Whenever you expend your Hit Dice to recover HP, for every 10 HP you would recover, you may heal 1 Wound instead." → `triggerOnEvent` (onHitDiceSpend), Wound-heal conversion — blocked by #584
- [x] **Unbreakable** (L7) — "(1/encounter) While Raging, if you would suffer your last Wound or other negative condition of your choice, you don't." → `resourceCounter` (1/encounter), `conditionalBonus` (Raging), `triggerOnEvent` (onWouldSufferLastWound / onWouldGainCondition), `conditionImmunity` (selective) — blocked by #576, #579, #584, #585 — ✅ partial, #861: 1/encounter use gate (charge pool, encounter-start refresh) + card reminder. Negating the Wound/condition stays manual.
- [x] **Titan's Fury** (L11) — "After you miss an attack or are crit by an enemy, Rage for free." → `triggerOnEvent` (onMiss, onCritReceived), free Rage activation — blocked by #584, Rage framework — ✅ partial, #861: Rage auto-restores when an enemy crit's damage is applied to you (`modifyToggle.turnOn: onCritReceived`). The on-miss half is deferred: misses do not reliably reach any pipeline event.
- [x] **Mountain's Endurance** (L15) — "While Dying, if an attack against you would be a crit, the attack is rerolled instead (when-crit abilities, such as Titan's Fury, still trigger)." → `conditionalBonus` (Dying), `modifyIncomingAttack` (reroll crit), preserve on-crit event — blocked by #579, #583 — ✅ already shipped in #859 (`modifyIncomingAttack` forceReroll on crits while Dying, Migration031).

[...]

Issue #600, opened 2026-04-08
[Automation] Berserker - Savage Arsenal abilities checklist


Fronix:
## Context
Checklist for all 12 **Savage Arsenal** abilities. Players pick these starting at L4, gaining an additional selection at L6, L8, L10, L12, L14, and L16 (total 7 by L16). All interact with the Fury Dice / Raging framework.

Pack files: `packs/classFeatures/core/berserker/savage-arsenal/*.json`

## Abilities

- [x] **Death Blow** — "After you deal damage from a crit, you may expend any number of Fury Dice. Sum the dice and deal double that amount of damage." → `triggerOnEvent` (onCritDamage), `dicePool` consumption, bonus damage = 2 × sum of spent dice — blocked by #578, #584 — ✅ partial, #861: manual Fury spend, chat shows `2 * @sum` bonus damage. The "after a crit" precondition is player-adjudicated (manual spends have no attack-outcome gate).
- [ ] **Rampage** — "(1/turn) After you land a hit, you may treat your next attack this turn as if you rolled that same amount instead of rolling again." → `resourceCounter` (1/turn), `triggerOnEvent` (onHit), next-attack roll-override flag — blocked by #576, #584
- [x] **Swift Fury** — "Whenever you gain one or more Fury Dice, move up to DEX spaces for free, ignoring difficult terrain." → `triggerOnEvent` (onFuryDieGained), `grantMovement` — blocked by #581, #584 — ✅ partial, #861: new `poolGainMessage` rule posts "move up to DEX spaces free" whenever Fury Dice are gained. Movement itself stays manual.
- [x] **Deathless Rage** — "(1/turn) While Dying, you may suffer 1 Wound to gain 1 action." → `resourceCounter` (1/turn), `conditionalBonus` (Dying), Wound cost, `grantAction` — blocked by #576, #579, #582 — ✅ partial, #861: 1/turn use gate (charge pool, turn-start refresh) + activation reminder. The Wound cost and extra action stay manual (blocked by #582).
- [x] **Eager for Battle** — "Gain advantage on Initiative. Move 2×DEX spaces for free on your first turn each encounter." → `initiativeRollMode` (✅ wired), `triggerOnEvent` (onFirstTurnOfEncounter), `grantMovement` — blocked by #581, #584 — ✅ partial, already shipped: `initiativeRollMode` advantage + `initiativeMessage` carrying 2×DEX. The movement is manual (#407).
[...]

PR #861, opened 2026-07-26
feat: Berserker automation - Fury Dice features, subclasses, and supporting engine primitives (#683)

### LINKED ISSUE #597: [Automation] Berserker - Bloodlust: spend Fury Dice for free movement
## Context
Berserker class feature **Bloodlust** (Level 3). A resource-swap mechanic: Fury Dice spent for movement instead of damage/reduction.

Pack file: `packs/classFeatures/core/berserker/berserker-progression/bloodlust.json`

## Rules from Rulebook

> "Bloodlust. Expend 1 or more Fury Dice on your turn, move DEX spaces per die spent for free."

## Implementation
- Once-per-turn-or-unlimited? Rules say "on your turn" without a 1/turn tag — treat as unlimited per turn but gated by available dice
- Player prompt: how many Fury Dice to spend
- Grants free movement = DEX × dice spent (no action cost)
- Consumes the chosen Fury Dice from the pool

## Rule Types
- `dicePool` (Fury Dice consumption)
- `grantMovement` (#581) — scaled by dice spent
- `grantAction` (#582) — free activation

## Blocked By
- #578 dicePool
- #581 grantMovement
- #582 grantAction
- Berserker Rage & Fury Dice framework (this batch)

## Open Question
How should the UI prompt the player to choose dice-spend amount for a movement ability that competes with That All You Got?! and Savage Arsenal damage spends?

⚠️ **Work cannot start until the foundational rule types above and the Rage framework are implemented.**

Part of #365

### LINKED ISSUE #598: [Automation] Berserker - Enduring Rage: dying-state action and save bypass
## Context
Berserker class feature **Enduring Rage** (Level 4). A conditional bundle that activates while Dying, modifying action economy and bypassing a save requirement.

Pack file: `packs/classFeatures/core/berserker/berserker-progression/enduring-rage.json`

## Rules from Rulebook

> "Enduring Rage. While Dying, you Rage automatically for free at the beginning of your turn, have a max of 2 actions instead of 1, and ignore the STR saves to make attacks."

## Implementation
- **Dying-state predicate** — gates all three sub-effects
- Sub-effect 1: `triggerOnEvent` onTurnStart → auto-activate Rage (free, no action cost) while Dying
- Sub-effect 2: Action-economy override — max actions while Dying becomes 2 instead of 1 (this is normally 1 while Dying per core rules)
- Sub-effect 3: Bypass the STR save that Dying creatures normally make to act
- Interacts with Level 18 Deep Rage (Rage does not end when dropping to 0 HP)

## Rule Types
- `conditionalBonus` (#579) — Dying predicate
- `triggerOnEvent` (#584) — onTurnStart auto-Rage
- `grantAction` (#582) — extra action slot while Dying
- Dying-state save bypass may require a new flag or `note` + engine handling

## Blocked By
- #579 conditionalBonus
- #582 grantAction
- #584 triggerOnEvent
- Berserker Rage & Fury Dice framework (this batch)

⚠️ **Work cannot start until the foundational rule types above and the Rage framework are implemented.**

Part of #365

Fronix:
## Summary

Implements the Berserker automation epic (#683): all remaining features that are feasible with current engine primitives, plus a set of new generic rule mechanisms (turn-boundary pool triggers, refill predicates, modifyConsumer, modifyToggle.turnOn, minFace, maximizeDie, poolGainMessage) that unlock them. Part of #365.

## Changes

- Pack rules for Bloodlust, Death Blow, Onslaught, Savage Awareness, One with the Ancients, and reminder note nodes on Whirlwind, You're Next!, Unstoppable Brutality, Thunderous Steps, Into the Fray
- Turn-boundary plumbing: combat document emits `nimbleCombatTurnStart` from `_onStartTurn` (GM-only); dice pools gain `onTurnStart`/`onTurnEnd` refill dispatch, charge pools gain `onTurnStart` recovery; the old all-clients `combatTurn` emitter is removed
- Refill-entry predicates and `modifyPool.addRefills` (Intensifying Fury turn-start die while Raging) and new `onCritReceived` trigger (MORE BLOOD!)
- New `modifyConsumer` rule type: augments dice-spend effect formulas (Stone's Resilience adds `@sum` to Fury damage-reduction spends)
- `modifyPool.minFace` die floor (Boundless Rage floors Fury Dice at 6)
- `modifyToggle.turnOn` + defender-side `onAttackReceived` rule dispatch (Enduring Rage auto-Rages while Dying; Titan's Fury restores Rage when crit)
- `maximizeDie` pool activation action + 1/turn and 1/encounter charge gates (Blood Frenzy, Deathless Rage, Unbreakable)
- New `poolGainMessage` rule type: chat reminder on pool gains (Swift Fury)
- `Migration032BerserkerAutomation` backfills all new rules and activation nodes onto existing actor-owned copies (idempotent, source-id match with class+name fallback)
- `docs/system/rules.md` updated; new English lang strings in `en.json` (flagged for human review per AGENTS.md)

## Test Plan

Unit coverage: new/extended tests for every mechanism (turn triggers, refill predicates, addRefills merge, minFace clamping, modifyConsumer application, toggle auto-turn-on, onAttackReceived dispatch, maximizePoolDie, poolGainMessage, speedBonus predicate gating).

In Foundry, with a Berserker test character (auto-apply conditions setting on):

1. Drag the Berserker class to a fresh character at level 3+ with Path of the Red Mist; activate **Rage**, then click **Bloodlust**: the Fury spend panel opens; spending N dice posts a chat card with DEX x N movement.
2. In combat while Raging, advance to the Berserker's turn start: **Intensifying Fury** rolls one free Fury Die (GM client performs the roll). End Rage: no die at next turn start.
3. As GM, apply a critical hit's damage to the Berserker: **MORE BLOOD!** (if owned) adds a Fury Die; **Titan's Fury** (Mountainheart) re-enables the Rage effect with a chat announcement.
4. Drop the Berserker to 0 HP with wounds below max (Dying), advance to their turn start: **Enduring Rage** re-enables Rage automatically.
5. With **Stone's Resilience** owned, use **That All You Got?!**: the spend preview and chat total include the spent die faces (`+ @sum`).
6. At level 20 (**Boundless Rage**), click Rage repeatedly: rolled Fury faces below 6 land as 6, chat flavor notes the raise.
7. Click **Blood Frenzy** twice in one turn: first use raises the lowest Fury Die to max, second is blocked (no charges); the charge refreshes next turn.
8. Load a world with a pre-existing Berserker actor: Migration032 appends the new rules/nodes exactly once; re-running is a no-op.

### Live v13 verification

Verified end-to-end in Foundry v13.351 with a scripted Playwright run driving the real UI (sheet feature clicks, top-tracker End Turn, dice-pool spend panel, chat-card Apply Damage): Rage activation with Rage(2)/die ladder/die floor, Intensifying Fury at turn start, Blood Frenzy maximize + 1/turn gate + turn-start recharge, TAYG + Stone's Resilience spend, MORE BLOOD! and Titan's Fury on an applied crit, and Enduring Rage auto-Rage while Dying. 23/23 checks pass.

Live testing surfaced two real bugs, fixed in the final commit:

1. Refill-entry predicates were read from initialized rule data as Predicate instances (enumerable shape `{ isValid, _source }`), so every predicated refill silently failed at runtime despite green unit tests. `normalizeRefills` now unwraps `_source`.
2. Item activation applied pool effect nodes during `getData`, before the `preUseItem` charge gate fires: a blocked use still mutated the pool. Pool nodes are now deferred and applied only after the gate allows the use (base item + spell flows).

Known-by-design notes confirmed live: a fully absorbed hit (banked damage reduction >= damage) applies no damage and therefore fires no crit-received triggers; Nimble's combat sorting puts characters before NPCs, so a Berserker raging before combat gets their Intensifying Fury die as soon as combat starts.

## Checklist

- [x] Tested in Foundry with no console errors (live v13.351, scripted real-UI run: 23/23 checks)

## Related Issues

Part of #683 and #365. Closes #597, closes #598. #595 and #407 are complete and can be closed independently; checklist updates for #599, #600, #601, #602


src/utils/dicePool/dicePoolRefill.ts
```diff
@@ -242,14 +268,18 @@ async function rollDieIntoPool(
...
-	const face = roll.total ?? 1;
+	const rolledFace = roll.total ?? 1;
+	const face = applyFaceFloor(rolledFace, pool.minFace);
 
 	if (!options.suppressChat) {
 		const ChatMessageCls = (globalThis as unknown as { ChatMessage: typeof ChatMessage })
 			.ChatMessage;
+		const baseFlavor = options.flavor ?? pool.label;
+		const flavor =
+			face !== rolledFace ? `${baseFlavor} (rolled ${rolledFace}, raised to ${face})` : baseFlavor;
```
trevlar: hardcoded English in a user-facing chat flavor: `` `${baseFlavor} (rolled ${rolledFace}, raised to ${face})` ``. AGENTS.md requires `localize()` for all user-facing strings, and your PR checklist claims this is clean. You already added `NIMBLE.dicePoolTracker.panel.useFeature.maximizedNote` for the sibling case — add an equivalent key here (e.g., `NIMBLE.dicePool.refill.raisedToFloor` with `${baseFlavor} (rolled ${rolledFace}, raised to ${face})` and use it here.

src/view/sheets/components/DicePoolPanel.svelte.ts
```diff
@@ -105,6 +105,41 @@ export function createDicePoolPanelState(
...
 
+	/** True when the chosen dice are transformed in place rather than spent. */
+	const isMaximizeOutcome = $derived(selectedConsumer?.selectionOutcome === 'maximize');
+
+	/**
+	 * How many dice the chosen consumer may act on. Transforming outcomes are
+	 * capped at the consumer's cost so a "change 1 die" feature offers exactly
+	 * one pick; spending outcomes stay uncapped ("expend 1 or more").
+	 */
+	const selectionCap = $derived.by((): number | null => {
```
trevlar, line 116: `selectionCap`, `canSelectDie` (line 126), the cap-eviction branch in `toggleDie` (line 210), and `applyMaximizeOutcome` (line 278) are ~90 lines of new branching logic with zero unit coverage, and this is the path Blood Frenzy actually uses in the pack. The repo has precedent for testing rune state modules (`CharacterLevelDownDialogState.svelte.test.ts`, `topTrackerStore.svelte.test.ts`, `characterCreation/state.svelte.test.ts`) — add a `DicePoolPanel.svelte.test.ts` covering: cap derived from `cost`, cap-1 behaving as a radio group, at-max dice being unselectable, and `applyMaximizeOutcome` writing only the selected sub-max faces.

src/utils/dicePool/dicePoolRefill.ts
```diff
@@ -334,6 +364,66 @@ async function rollPoolFresh(
...
+	/** Why nothing changed, for the chat card's skip label. */
+	reason?: 'unknownPool' | 'poolEmpty' | 'allAtMax';
+};
+
+/**
+ * Set the lowest `count` faces in a pool to the die's maximum face value
+ * (always the optimal choice for the player). Reports whether any face
+ * changed and, if not, why. Driven by the `maximizeDie` pool-node action.
+ */
+async function maximizePoolDie(
```
trevlar, line 391: `maximizePoolDie` and the `maximizeDie` pool-node branch (`ItemActivationManager.ts:748`, plus the builder option at `itemActivationEffectsConfigTabHelpers.ts:129` and the `poolEmpty`/`allAtMax` skip reasons) have no pack consumer — `grep -rl maximizeDie packs/` returns nothing; Blood Frenzy went with `selectionOutcome: "maximize"` instead. That leaves two independent implementations of "raise dice to max", and the tested one isn't the one that ships. Either drop the pool-node path (YAGNI) or have `applyMaximizeOutcome` delegate to `maximizePoolDie` so there's one implementation behind both entry points. It might be needed for future work, so please confirm before making the change.

src/managers/ItemActivationManager.ts
```diff
@@ -476,9 +479,37 @@ class ItemActivationManager {
...
 		this.activationData.effects = dependencies.reconstructEffectsTree(updatedEffects);
 
+		// Pool nodes mutate actor state, so their application is deferred until
+		// the caller confirms the use is allowed (the preUseItem gate fires
+		// after getData). Collect them from the RECONSTRUCTED tree: it deep
+		// clones every node, and results must land on the objects the chat card
+		// serializes. See applyDeferredPoolNodes().
+		this.#deferredPoolNodes = ItemActivationManager.#collectPoolNodes(
+			this.activationData.effects as EffectNode[],
+		);
```
trevlar, line 496: deferring is right, but collecting from the reconstructed tree changes which nodes get applied. `reconstructEffectsTree` drops any node whose parent isn't `damage`/`savingThrow` (and any non-damage child in `sharedRolls`), so a pool node parented under, say, a healing node used to fire from the old flat loop and now silently won't. Worth confirming the builder can't produce that parenting, or falling back to the flattened list for collection and only reading results off the reconstructed clones.

src/utils/dicePool/dicePoolConsumers.ts
```diff
@@ -26,6 +27,72 @@ function readEffectFormula(consumer: DiceConsumerRuleLike): string | null {
...
+				type?: string;
+				disabled?: boolean;
+				poolIdentifier?: string;
+				effectTypeFilter?: string;
+				appendFormula?: string;
+				priority?: number;
+				appliesTo?: () => boolean;
+			};
+			if (rule.type !== 'modifyConsumer' || rule.disabled) continue;
+			if (normalizeIdentifier(rule.poolIdentifier) !== poolIdentifier) continue;
```
trevlar, line 66: `getConsumerModifiers` matches modifiers on `poolIdentifier` alone, while the consumer loop 70 lines down also checks `(consumer.poolScope ?? 'item') !== pool.scope`. An actor-scoped and an item-scoped pool sharing an identifier would cross-contaminate. Pass `pool.scope` through and filter on it for consistency.

src/documents/combat/combat.svelte.ts
```diff
@@ -701,6 +701,20 @@ class NimbleCombat extends Combat {
 		return super.createEmbeddedDocuments(embeddedName, normalizedData, operation);
 	}
 
+	override async _onStartTurn(
```
trevlar, line 710: `_onStartTurn` emits `nimbleCombatTurnStart` with no backstop, but `#ensureOutgoingCharacterRefilled` (line 779) documents that under Nimble's expanded solo-monster turn order, "Foundry's cached turn history can desync from our directly-resynced turn index, so its turn-event dispatcher computes an empty interval and skips the outgoing hero's end-of-turn entirely." An empty interval also skips `_onStartTurn` for the incoming combatant. The old `combatTurn` listener you removed fired on every turn change regardless of that interval, so this is a regression for the charge-pool `onTurnStart` recovery that already ships on `dev` — and it lands squarely on Blood Frenzy and Deathless Rage, whose 1/turn charge pools recharge on `onTurnStart`. Mirror the end-turn pattern: capture the incoming combatant in `nextTurn`/`nextRound` and add an `#ensureIncomingCharacterTurnStarted` backstop (claim-before-await so a late `_onStartTurn` no-ops). Your Playwright run wouldn't have caught this — no solo monster in the scripted encounter.

trevlar (changes requested):
**🔧 Push back — the new turn-start emitter drops the backstop your own combat code says it needs, and the maximize path that actually ships is the untested one**

Big, well-organised piece of work — the engine primitives are cleanly separated, the migration is properly idempotent, and the unit coverage on the rule/util layer is genuinely thorough. Three things need to change before merge.

**What's good**

- `Migration032BerserkerAutomation` is properly idempotent — signature-based rule dedupe with an id fallback, node dedupe by id, and a sensible source-id → class+name match order.
- Catching the `Predicate`-instance-vs-`_source` unwrap and the pre-gate pool mutation in live testing (and fixing both) is exactly the kind of thing green unit tests hide.

---

1. Backstop `nimbleCombatTurnStart` against Foundry's skipped-interval case.
2. Localize the "rolled X, raised to Y" chat flavor.
3. Add unit tests for the DicePoolPanel maximize path.
4. Collapse the two maximize implementations into one.
5. Verify deferred pool-node collection can't drop nodes.
6. Filter `modifyConsumer` matches by pool scope too.

src/utils/dicePool/dicePoolRefill.ts
Fronix: Fixed

src/view/sheets/components/DicePoolPanel.svelte.ts
Fronix: Fixed

src/utils/dicePool/dicePoolRefill.ts
Fronix: Fixed

src/managers/ItemActivationManager.ts
Fronix: Fixed

src/utils/dicePool/dicePoolConsumers.ts
Fronix: Fixed

src/documents/combat/combat.svelte.ts
Fronix: Fixed

Fronix:
## Review fixes

All six points are addressed. Commits: `ff1fe993`, `e4e8e41b`, `cc7f2257`, `fcde3637`. `pnpm check` is green at 2806 tests.

---

### 1. Turn-start hook had no backstop

This was a real bug, and it is now verified live in Foundry v13.351. Details and reproduction are in the section below.

The fix mirrors the turn-end pattern already in `combat.svelte.ts`. Both `_onStartTurn` and a new post-advance backstop claim the combatant id synchronously before any `await`, and only whichever claims first emits the hook. The synchronous claim matters because Foundry dispatches turn events fire-and-forget, so a late `_onStartTurn` could otherwise double-fire.

One correction to the review's framing. On `dev`, nothing listens to `nimbleCombatTurnStart`. The emitter shipped but the recovery listener did not, so this is not a regression against `dev`. It is a latent hole in code that this PR makes load-bearing for the first time, which is a good reason to fix it now.

### 2. Hardcoded English in the chat flavor

Correct, and my PR checklist claim was wrong. The "rolled X, raised to Y" string now goes through `NIMBLE.dicePoolTracker.raisedToFloor`.

### 3 and 4. Two maximize implementations, and the tested one was unused

Also correct. `maximizePoolDie` now takes an optional list of explicit indices and reports which faces it raised. The spend panel delegates to it instead of writing faces itself, so there is one implementation behind both entry points.

You asked to confirm before removing the pool-node path. Keeping it was the call, so I collapsed the duplication rather than deleting anything.

New file `DicePoolPanel.svelte.test.ts` covers the panel path that Blood Frenzy actually uses:

- selection cap derived from the consumer's cost
- a cap of 1 behaving like a radio group, where a second pick replaces the first
- dice already at their maximum being unselectable
- the raise applying only to the picked dice
- the consume path still spending dice instead of raising them

The tests run against the real pool, consumer and refill code with a fixture actor rather than mocks. Mocking `#`-aliased imports does not reach a module loaded through a Svelte harness in this setup, and the fixture turned out to be the better test anyway because it asserts on the persisted faces.

### 5. Deferred pool-node collection

Changed as suggested. Collection now walks the flat list and prefers the reconstructed clone by id, so results still land on the objects the chat card serializes.

That said, the scenario cannot actually happen today, and I would rather say so than claim a fix for a live bug:

- The effect builder only offers `pool` at the top level, so the UI cannot create a pool node under a healing node.
- `flattenEffectsTree` only recurses into `on` and `sharedRolls` of damage and saving-throw nodes, and it recomputes `parentNode` from the traversal position. A pool node nested anywhere else never reaches the flat list at all.

So the change is cheap insurance rather than a bug fix. The test reflects that and does not pretend to reproduce a drop.

### 6. modifyConsumer pool scope

Correct. The rule gains a `poolScope` field defaulting to `item`, and matching now filters on it the same way consumer enumeration does. The shipped Berserker `fury` pool is item-scoped, so Stone's Resilience keeps working with no pack edit and no migration change.

---

## Live reproduction of the turn-start bug

Verified in Foundry v13.351. Every action under test is a real DOM click. Only token placement, combatant creation and the pool baseline used the API.

**Setup**

1. Two Berserkers, Sigrid Emberfall and Ulfhild Bloodmist, plus one solo monster, Alaric Draegoth.
2. Nimble expands solo-monster turns so the monster acts after each hero. Three combatants produce a four-entry turn order:
   `Sigrid -> Alaric -> Ulfhild -> Alaric`
3. Sigrid is the right subject because she carries two independent `onTurnStart` consumers:
   - Blood Frenzy's 1/turn charge pool, `recovery: onTurnStart / refresh / 1`
   - Intensifying Fury's Fury Die, `addRefills: onTurnStart / add / 1`, predicated on `self:raging`
4. `Combat.prototype._onStartTurn` was wrapped purely to record whether Foundry called it. The wrapper calls through unchanged.

**Steps**

1. Click "Start Combat" in the Nimble tracker.
2. Open Sigrid's sheet, go to the Features tab, click Rage. She is now raging with Fury `[5, 5]`.
3. Click Blood Frenzy. The charge goes from 1 to 0.
4. Click Blood Frenzy again on the same turn. It is blocked, and produces zero chat cards.
5. Click "Next Turn" four times to walk the full expanded round back to Sigrid.

**What happened**

```
click 1: Sigrid  -> Alaric  (r1t1) | _onStartTurn fired   | hook emitted
click 2: Alaric  -> Ulfhild (r1t2) | _onStartTurn fired   | hook emitted
click 3: Ulfhild -> Alaric  (r1t3) | _onStartTurn fired   | hook emitted
click 4: Alaric  -> Sigrid  (r2t0) | _onStartTurn SKIPPED | hook emitted  <- BACKSTOP
```

At the round wrap Foundry never called `_onStartTurn` for Sigrid. The hook fired anyway, from the backstop, and both consumers ran:

- Blood Frenzy charge went from 0 back to 1
- Fury went from `[5, 5]` to `[5, 5, 3]`

**Why this proves the fix and not just a passing run**

Before the fix, `nimbleCombatTurnStart` was emitted from exactly one place, inside `_onStartTurn`. `git show ff1fe993^` confirms `nextTurn` and `nextRound` contained no reference to it. The probe shows that method never ran at this boundary. On the pre-fix build the hook could not have fired, so Sigrid would have started round 2 with Blood Frenzy still spent and no free Fury Die, silently, with nothing in the log.

**One caveat worth knowing**

The skip lands on the combatant coming in at the wrap, which is turn index 0. It does not land on whoever you happen to be watching. My first run used Ulfhild as the subject, her turn start fired normally, and the backstop was quietly carrying Sigrid instead. I re-ran targeting the hero the skip actually hits. Stopping at the first run would have reported a pass that never exercised the fix.

trevlar (approved):
LGTM 👍🏻

PR #862, opened 2026-07-27
fix(rules-builder): resolve modifyPool pool picker from poolType (#806)

### LINKED ISSUE #806: Charge pool identifiers don't show up in modifyPool in rule builder
Charge pool identifiers should also show up in this list but when set manually it says _not found_

Image

--- 

Should be a straight forward fix but might affect some other places, so worth taking a deeper look where the dicepool identifier is used before throwing in a fix.

Fronix:
## Summary

The `modifyPool` rule can target a dice pool or a charge pool, but its `poolIdentifier` field always rendered the **dice** pool picker. With `poolType: 'charge'` the actor's charge pools never appeared in the dropdown and any stored identifier showed as `⚠ combat-dice (not found)`, because dice and charge pools are separate subsystems with separate identifier namespaces. This adds a conditional-widget primitive to the rules-builder schema layer so `modifyPool` picks its picker from `poolType`, and fixes the related case where an item with no parent actor left the picker permanently disabled.

## Changes

- **`_widgetOption.ts`**: `WidgetExtras.widget` now accepts `WidgetHint | WidgetResolver`, where a resolver is `(data) => WidgetHint` evaluated against the rule's source data. This mirrors the existing `showWhen` predicate. `withWidget()` deliberately does not invoke a resolver: `defineSchema()` runs at system init, before rule data exists.
- **`SchemaFieldRendererState.svelte.ts`**: `widget` becomes a `$derived.by` that resolves the function against `getParentData()`. Everything downstream (the dispatch table, `isHidden`, the unknown-hint warning) still sees a plain string, so the resolved hint is validated at render time and `SchemaFieldRenderer.svelte` needed no change at all.
- **`modifyPool.ts`**: `poolIdentifier` resolves to `chargePoolPicker` when `poolType` is `charge`, `dicePoolPicker` otherwise. Replaces the hardcoded hint and the stale "no conditional-widget primitive yet" comment.
- **`Dice/ChargePoolPicker.svelte`**: when the item has no parent actor (compendium or world item), the pool list is empty and the control was rendered as a permanently disabled `<select>`. It is now a free-text input with the empty-state message as a hint beneath it. The actor-owned path is unchanged. Both pickers also gained a `nimble-pool-picker--dice` / `--charge` marker class so tests can tell them apart (they otherwise share `.nimble-field-input`).
- **`generateReference.gen.ts`**: resolves function-valued widgets against the schema's field initials and appends "(The input changes with the other field values.)" to the description, alongside the existing `showWhen` note. `modifyPool.poolIdentifier` therefore documents as the dice picker, matching `poolType`'s `dice` default.
- **`docs/system/rules.md`**: documents the resolver form and the contract that the fallback branch must match the sibling field's `initial`. Also corrects the widget catalog list, which was already missing both pool pickers.
- **`en.json`**: one new key, `NIMBLE.rulesBuilder.poolIdentifierPlaceholder` ("Type a pool identifier"). **Flagging for human review per AGENTS.md.**

## Test Plan

**Automated** — `pnpm check` passes. New coverage:

- `src/models/rules/modifyPool.test.ts` (new file): the resolver returns `chargePoolPicker` for `poolType: 'charge'`, `dicePoolPicker` for `'dice'`, and `dicePoolPicker` for `{}`. The last one locks in the docs-generator contract, and a companion assertion pins `poolType`'s initial to `dice` so the two can't drift apart.
- `SchemaFieldRenderer.svelte.test.ts`: resolver dispatch for both branches, a resolver returning `'hidden'` (proves `isHidden` composes with resolution), and the free-text fallback for each picker when no actor is resolvable.
- `_widgetOption.test.ts`: a function-valued hint does not warn, and is **not invoked** during validation.

**Manual (v13)** — to reproduce the original bug, check out `dev` and open any item carrying a `modifyPool` rule with `poolType: charge`; the compendium already has one at `packs/classFeatures/core/commander/commander-progression/fit-for-any-battlefield.json` (`poolIdentifier: "combat-dice"`). The Pool identifier field shows `⚠ combat-dice (not found)`.

On this branch:

1. Create a character. Give it one item with a `chargePool` rule (`identifier: combat-dice`, `scope: actor`) and another with a `dicePool` rule (`identifier: fury`, `scope: actor`). Add a third item with a `modifyPool` rule set to `poolType: charge`, `poolIdentifier: combat-dice`.
2. Open that third item → **Rules** tab → **Open Rules Builder**. The Pool identifier field is now a charge picker showing **Combat Dice**, not a "not found" warning.
3. Change **Pool type** to `dice`. The field swaps to the dice picker listing **Fury** (and correctly flags the now-stale `combat-dice` value).
4. Open a **world or compendium** item with the same rule (no parent actor). The identifier is an editable text input pre-filled with the stored value, with "No charge pools defined on this actor" as a hint. Typing a new identifier persists it. On `dev` this control is disabled and unusable.

Steps 2 to 4 were run end-to-end in Foundry v13 through the real UI (sidebar → sheet → Rules tab → Rules Builder button → real `<select>` interaction), 8/8 assertions passing, no console errors beyond the world's pre-existing missing-asset warning.

## Checklist

### Notes for the reviewer

- **KISS check.** AGENTS.md forbids "runtime string-to-function dispatch". This is the inverse: a resolver produces a hint *string*, and the existing explicit `{#if}` chain then dispatches on it. No component lookup table, no dynamic import. Worth a second opinion, since it is the one judgement call here.
- **Deliberately not done.** `DicePoolPicker` and `ChargePoolPicker` are roughly 95% identical (`resolveActor` and `dedupeAndSort` are byte-identical; `registerPoolHooks` differs only in the hook-name array). Deduplicating them meets neither the Rule of Three nor a behavioural need here, and parameterizing `createSubscriber(registerPoolHooks)` incorrectly would leak hook registrations or subscribe to the wrong invalidation events. It would also conflict with the in-flight `feat/berserker-automation` branch. Suggest a follow-up issue once that lands.
- `feat/berserker-automation` does not fix this; it adds `minFace`/`addRefills` to `modifyPool` but leaves the widget block untouched. Its `modifyConsumer` and `poolGainMessage` rules are dice-only by design and need no equivalent change.

## Related Issues

Closes #806

### DIFF
--- a/docs/system/rules.md
+++ b/docs/system/rules.md
@@ -333,9 +333,27 @@ value: new fields.StringField(
 ),
 ```
 
-The closed widget catalog is **`formula | diceFormula | documentUuid | predicate | templateString | richText | hidden`**. `withWidget()` validates the hint in dev and warns on typos.
+The closed widget catalog is **`formula | diceFormula | documentUuid | predicate | templateString | richText | dicePoolPicker | chargePoolPicker | hidden`**. `withWidget()` validates the hint in dev and warns on typos.
 
 For `widget: 'documentUuid'`, set `documentTypes: ['Item.spell']` (or `['Item']`, `['Actor']`) to gate accepted drops.
+
+### Conditional widgets
+
+When the right input depends on a sibling field, pass a function instead of a hint. It receives the same data object as `showWhen`:
+
+```typescript
+poolIdentifier: new fields.StringField(
+  withWidget({
+    required: true,
+    label: 'Pool',
+    widget: (data) => (data.poolType === 'charge' ? 'chargePoolPicker' : 'dicePoolPicker'),
+  }),
+),
+```
+
+The resolved value must still be in the catalog — `withWidget()` cannot check a resolver's return value, so `<SchemaFieldRenderer>` warns at render time if it isn't. `'hidden'` works here too, giving conditional visibility that depends on which widget applies.
+
+Write the fallback branch to match the sibling field's `initial` value: the generated reference docs resolve the function against schema initials, so that branch is what gets documented (the field's description then notes that the input varies).
 
 ### Conditional fields with `showWhen`
 
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -608,6 +608,7 @@
 				"empty": "No charge pools defined on this actor",
 				"notFound": "⚠ {identifier} (not found)"
 			},
+			"poolIdentifierPlaceholder": "Type a pool identifier",
 			"templateValueHintBefore": "Use",
 			"templateValueHintAfter": "to insert the formula result.",
 			"pickerMissingMetaTooltip": "Description missing — flag for review.",
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -609,6 +609,7 @@
 				"notFound": "⚠ {identifier} (not found)"
 			},
 			"poolIdentifierPlaceholder": "Type a pool identifier",
+			"poolPickerNoActor": "This item is not on an actor, so its pools cannot be listed",
 			"templateValueHintBefore": "Use",
 			"templateValueHintAfter": "to insert the formula result.",
 			"pickerMissingMetaTooltip": "Description missing — flag for review.",
--- a/scripts/docs/generateReference.gen.ts
+++ b/scripts/docs/generateReference.gen.ts
@@ -213,13 +213,54 @@ function resolveInitial(rawInitial: unknown): string {
 	return String(initial);
 }
 
-function describeRuleField(name: string, field: any): FieldRow | null {
-	const opts = field?.options ?? {};
-	if (opts.widget === 'hidden') return null;
+/** Raw `initial` value, for feeding widget resolvers. Unlike `resolveInitial`,
+ *  this keeps the value as-is rather than formatting it for display. */
+function rawInitialValue(rawInitial: unknown): unknown {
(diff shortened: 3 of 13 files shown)


trevlar (approved):
**✅ Approved**

Solid fix — the resolver primitive is the right shape for this, and you threaded it through every consumer of `options.widget` (renderer state + docs generator) with nothing left reading the raw value as a string.

- Your KISS call is correct: `WidgetResolver` returns a *hint string* and the existing explicit `{#if}` chain still does the dispatch, so there's no dynamic component lookup. That's not what AGENTS.md forbids.
- Nice touch pinning `poolType.initial` to `'dice'` in the same test that asserts the `{}` fallback — that's exactly the drift the docs-generator contract would otherwise hide. I ran `pnpm docs:generate` on your branch and confirmed `modifyPool.poolIdentifier` renders as **Dice pool picker** with the "(The input changes with the other field values.)" note appended, and no other rows changed.

One non-blocking note: **`src/view/rulesBuilder/components/ChargePoolPicker.svelte:47`** (and `DicePoolPicker.svelte:47`) — the `!hasPools` branch reuses `…chargePoolPicker.empty` ("No charge pools defined on this actor") as the hint, but that branch is also the *no actor at all* path (compendium/world item), where the sentence is false. Your inline comment says "No actor to enumerate pools from" while the condition is really "no pools resolved." A separate key for the unowned case (or a neutral rewording) would read better for pack authors.

PR #871, opened 2026-07-29
feat(damage): apply resistances, vulnerabilities and immunities in the damage pipeline

### LINKED ISSUE #785: Apply damage resistances, vulnerabilities, and immunities in the damage pipeline
### Problem or Motivation

Monster data models (`NPC`, `SoloMonster`, `Minion`) already define `damageResistances`, `damageVulnerabilities`, and `damageImmunities` as string-array fields, and ~100 monster/legendary monster packs populate them. However, **nothing in the damage application pipeline reads these fields**. They are display/data-only. A skeleton "Resistant to piercing, vulnerable to bludgeoning" takes full damage from both. `CharacterDataModel` does not have these fields at all, so PC-side resistances (Stormborn, Divine Grace, Epic Resistance, etc.) cannot be modeled.

Per the Nimble rules (CoreRules):

- **Resistance**: "Take half as much damage."
- **Vulnerable**: "When a creature is vulnerable to a damage type, that kind of damage ignores its armor; if unarmored, they take double the damage instead."
- **Immunity**: referenced indirectly (e.g., Fiery Embrace reduces immunity to resistance). Takes 0 damage of that type.

Note the asymmetry: vulnerability is **not** a simple x2 multiplier. It interacts with armor.

### Proposed Solution

Wire the existing schema fields into the damage application pipeline, and add the same fields to `CharacterDataModel` so PCs can use them.

**Schema**

- Add `damageResistances`, `damageVulnerabilities`, `damageImmunities` (string arrays) to `CharacterDataModel` matching the shape used in `NPCDataModel`, `SoloMonsterDataModel`, and `MinionDataModel`.
- No migration needed for monsters (fields already exist with `initial: []`). Characters will default to empty arrays.

**Pipeline**

The relevant entry point is `chatMessage.applyDamage(value, options)` in `src/documents/chatMessage.ts`, which calls `calculateArmorAdjustedDamage()` per target inside `buildDamageApplicationPlan()`. Today this function only applies the half-damage outcome and the heavy-armor halving.

Extend it to also apply per-damage-type modifiers:

1. Thread damage type through. `DamageRoll` already carries `damageType?: string`. `DamageApplyOptions` already has `roll` and `rolls` (serialized `DamageRoll`). Read the damage type(s) from the serialized roll(s) when building the adjusted damage.
2. Per target, compare each damage component's type against the target's three arrays:
   - **Immune**: that component contributes 0.
   - **Vulnerable**: that component ignores the target's armor (skip the heavy-armor halving for that portion). If the target is unarmored, double the component instead. This also aligns with the existing rule that crits plus vulnerability ignore armor (CoreRules-2:472).
   - **Resistant**: that component is halved (`Math.ceil(x * 0.5)` to match existing rounding convention).
3. Order of operations (per damage component): start with raw dice damage, apply outcome half-damage if any, apply immunity/vulnerability/resistance, apply armor halving unless vulnerable or crit, sum across components, add negative modifiers (untouched).
4. If a roll has no `damageType`, treat it as untyped and skip resistance lookups (current behavior).

**UI**

- Sheets need an editor for the three arrays on Character, NPC, SoloMonster, and Minion (NPC/Solo/Minion may already expose this; verify and add to PC sheet at minimum).
- The chat card needs a per-target breakdown so the player can see *which damage component was modified and why*. Today the card shows total damage and how it was rolled (each `DamageRoll.svelte` summary already carries its own `damageType`, so multi-type attacks render as separate rows), but there is no per-target adjustment view. When damage is applied to selected targets, the resolved per-target outcome must surface:
  - Per target, per damage component: the original component damage, the modifier that was applied (Immune, Resistant, Vulnerable, ignored armor due to vulnerability, etc.), and the final contribution to that target's HP loss.
  - Worked example for a 10 fire + 6 cold attack against a fire-immune, cold-resistant target: render something like `Goblin: fire 10 -> 0 (Immune), cold 6 -> 3 (Resistant). Total: 3`. The current card would just show `16` with no indication that 13 of it was negated.
  - Reuse the existing `DamageOutcomeNode` / `DamageNode` rendering surface where possible (see `src/view/chat/components/DamageOutcomeNode.svelte`, `DamageNode.svelte`). The per-target breakdown likely belongs in or beside the existing outcome display, not in the source roll display.
  - This breakdown should also appear (or be inferable) in `MinionGroupAttackCard.svelte`, which currently calls `messageDocument.applyDamage(normalizedDamage, ...)` once for the group.
- Tooltips on each adjusted number should restate the rule that triggered it (e.g., hover "0 (Immune)" -> "Immune to fire damage").

**Tests**

`chatMessage.test.ts` already has the pattern for testing adjusted damage with rolls. Add cases:
- Resistant target halves typed damage, leaves untyped untouched.
- Immune target zeros that component.
- Vulnerable target bypasses heavy-armor halving; doubles damage if unarmored.
- Multi-component rolls (e.g., 2d6 fire + 1d6 cold) where target resists only one type.
- Interaction with existing `outcome: 'halfDamage'` (the outcome-half stacks multiplicatively with resistance, per rules; verify against rulebook before merging).

### Alternatives Considered

- **Flat multiplier for vulnerability (x2 always).** Rejected: doesn't match the rulebook's armor-bypass wording and would double-dip against armored targets.
- **Apply modifiers on the actor side in `actor.applyDamage(damage)`.** Rejected: by that point damage type and roll context are lost. The chat message layer is where targets and roll metadata coexist.
- **Resistance only, defer vulnerability and immunity.** Considered. Decided against because all three fields already exist in the monster schema and pack data, so leaving two of them inert is more confusing than wiring the full set.

### Additional Context

Rule references:
- Resistance definition
- Vulnerable definition
- Crits and vulnerabilities ignore armor
- Fiery Embrace (immunity reduced to resistance example)

Code references:
- `src/documents/chatMessage.ts:159` (`calculateArmorAdjustedDamage`, the extension point)
- `src/documents/chatMessage.ts:208` (`buildDamageApplicationPlan`)
- `src/documents/chatMessage.ts:451` (`applyDamage` entry point)
- `src/documents/actor/base.svelte.ts:418` (`actor.applyDamage`, untyped HP subtraction; should remain type-agnostic)
- `src/dice/DamageRoll.ts:22` (`damageType?: string` already on the roll)
- `src/models/actor/NPCDataModel.ts:11-22`, `SoloMonsterDataModel.ts:11-22`, `MinionDataModel.ts:11-22` (existing schema)
- `src/models/actor/CharacterDataModel.ts` (needs the three fields added)

Fronix:
## Summary

Monster data models have carried `damageResistances`, `damageVulnerabilities` and `damageImmunities` for a while, and ~100 pack monsters populate them, but nothing in the damage pipeline read them — a skeleton "resistant to piercing, vulnerable to bludgeoning" took full damage from both. This wires all three into `chatMessage.applyDamage`, adds the same fields to characters so PC-side defenses can be modelled at all, and surfaces the result on the chat card and both sheets.

_Note: The target names are T: ..., it's not that each target has T: before it's name_
zen_u2aOVSs9Jm

## Changes

- **Vulnerability in the damage pipeline.** Per the Core Rules glossary, vulnerable damage ignores the target's armor, and an unarmored target takes double instead — not a flat ×2. `calculateArmorAdjustedDamage` resolves it before armor: an armored monster loses its medium/heavy reduction for that component, an unarmored one doubles. Immunity still outranks vulnerability, and a half-damage outcome halves before the doubling applies.
- **Heroes always take the unarmored branch, deliberately.** A hero's `attributes.armor` is a schema object feeding the Defend reaction rather than the monster armor string, so `getActorArmorType` reports `'none'` for them by design — documented at both sites, and pinned by tests using character-shaped actors so it fails if that ever becomes a string.
- **Damage defenses on characters.** `CharacterDataModel` gains the three arrays in the shape monsters already use, so the pipeline reads them off any actor without branching on type. No migration needed.
- **One shared defenses editor.** The three tag groups move out of `NPCMetaConfigDialog` into `DamageDefensesConfig.svelte`, now used by monsters and by a new `actor.configureDamageDefenses()` dialog.
- **Character sheet section.** Damage defenses get their own Core tab section beside the proficiency lists: always readable (including on a locked sheet, so players can see their own), with the edit control following the sheet's edit toggle. Previously the only route was the meta dialog's pencil, which needs ancestry/class metadata to render — a bare character had no route at all.
- **Per-target badges on the chat card.** `getDamagePreviewForTarget` and `getDamageModifiersForTarget` collapse into one `getDamageBreakdownForTarget`, returning per-component `{damageType, damageBeforeDefenses, adjustedDamage, modifiers[]}` plus a total. The target row shows `IMM` / `VUL` / `RES` / `DR` badges; the badge tooltip carries the rule text and the damage total carries the per-type breakdown.
- **Themed badge colours.** New tokens alongside the existing roll-result colours, in both light and dark. Every pair clears WCAG AA (5.1:1 or better).

## Test Plan

There are test actors and a scene for this in the dev world if you want them, but from scratch:

1. **Vulnerability, unarmored.** Make an NPC with no armor and `fire` under Damage Vulnerabilities (sheet → config → Damage Vulnerabilities). Target it, cast a fire attack, click Apply Damage. It should take **double** the rolled damage.
2. **Vulnerability, armored.** Same but set Heavy Armor. It should take the **full** rolled damage — heavy armor would normally halve the dice and drop positive modifiers, and vulnerability bypasses both. Compare against a heavy-armored NPC with no vulnerability to see the difference.
3. **Immunity wins.** Add `fire` to *both* Immunities and Vulnerabilities. It should take **0**.
4. **Multi-type attack.** Point a card with more than one damage type at a target that resists one type and is immune to another. Each component resolves separately; hover the red total on the target row to see `Fire 0 (from 8), Cold 4 (from 7)`.
5. **Badges.** With targets selected, confirm `IMM` / `VUL` / `RES` appear on the rows, and that hovering each gives the full rule text. Badges are GM-only.
6. **Character defenses.** Open a PC → Core tab → Damage Defenses. Unlock the edit toggle, click the pencil, toggle some types, confirm they render in the section. Lock the toggle again: the values stay visible, the pencil disappears. Try it on a character with no ancestry or class — that case had no route to the editor before.
7. **Themes.** Check the badges in both Foundry light and dark themes.

## Checklist

### Notes for the reviewer

- English `en.json` strings were added/changed and need a human read.
- NPCs still reach the shared editor through `NPCMetaConfigDialog` while characters reach it through the Core tab section, so the two actor types have differently-shaped routes to the same control. Left alone as the monster route already worked.

## Related Issues

Closes #785


src/documents/chatMessage.ts
```diff
@@ -316,9 +367,120 @@ function actorIsImmuneToDamage(actor: Actor.Implementation, damageType?: string)
...
+		'system.attributes.damageVulnerabilities',
+	);
+	return (
+		Array.isArray(vulnerabilities) && Boolean(damageType && vulnerabilities.includes(damageType))
+	);
+}
+
+/** Localized damage type name, falling back to the raw key for unknown types. */
+function getDamageTypeLabel(damageType: string): string {
+	const key = (CONFIG.NIMBLE.damageTypes as Record<string, string>)[damageType];
```
trevlar: `getDamageTypeLabel` is now the *third* copy of the same "look up `CONFIG.NIMBLE.damageTypes[type]`, localize, fall back to the raw key" helper: this file, `PlayerCharacterCoreTab.svelte:38`, and the pre-existing `NPCCoreTab.svelte:362`. AGENTS.md's Rule of Three fires at exactly three occurrences across different files, and the Code Promotion Rules require the promoting PR to update all existing call-sites. Pull it into `src/utils/` and rewire all three — that also fixes `NPCCoreTab.svelte:364` calling `game.i18n.localize` directly instead of the `localize()` util.

src/view/chat/components/Targets.svelte
```diff
@@ -195,43 +238,75 @@
...
+			--nimble-damage-badge-background-color: var(--nimble-damage-badge-reduction-background-color);
 		}
 	}
 
 	.nimble-target-list {
 		--nimble-button-padding: 0;
 
-		--nimble-card-content-grid: 'img title armor button';
-		--nimble-card-column-dimensions: 1.75rem 1fr 1rem 2rem;
+		--nimble-card-content-grid: 'img title badges armor button';
```
trevlar, line 316: the armor icon comes from `{@html getArmorIcon(token)}` and has no `grid-area`, so it's auto-placed. With badges present it lands in `armor` (col 4) as intended, but on a row with no badges — the common case — it auto-places into the now-empty `badges` column and leaves the 1rem `armor` column dead, adding ~1.5rem of gap before the trash button. A `:global(.nimble-armor-icon) { grid-area: armor; }` pins it in both cases.

trevlar (changes requested):
**🔧 Push back — one style-guide extraction to do; the damage math itself checks out**

I traced every branch of `calculateArmorAdjustedDamage` / `calculateAdjustedDamage` against the rule text you quoted and the ordering holds up: armored+vulnerable returns the outcome-scaled roll total (so heavy armor's dice-halving *and* its positive-modifier drop are both bypassed), unarmored doubles, immunity zeroes after, and resistance halves after that. The `excludedPrimaryDieValue` / crit / negative-modifier handling stays consistent with the existing `ignoreArmor` path. Only one thing genuinely needs to change.

**What's good**
- The vulnerability semantics are the part that's easy to get wrong, and modelling it as an armor interaction rather than a `×2` multiplier — with immunity outranking it and half-damage outcomes halving first — is the correct reading. The armored/unarmored/ignoreArmor/crit matrix is pinned by tests, including character-shaped actors so a future schema change fails loudly.
- Collapsing `getDamagePreviewForTarget` + `getDamageModifiersForTarget` into one per-component `getDamageBreakdownForTarget` is a real simplification: the bank-consumption logic now mirrors `buildDamageApplicationPlan` in one place instead of two, and the immune-target case stopped falsely crediting the bank as a reason.

---

1. Promote the damage-type-label helper to `src/utils/`, update all 3 call-sites.
2. Pin the armor icon with `grid-area: armor` so badge-less rows don't gap.

src/view/chat/components/Targets.svelte
Fronix: Fixed

src/documents/chatMessage.ts
Fronix: Fixed

PR #875, opened 2026-07-30
feat(sheet): move wounds to a segmented track under the hit point bar

Fronix:
## Summary

Wounds were a column of droplet pips floating over the character portrait, hidden at `opacity: 0` until you hovered the image. A healthy character gave no indication the mechanic existed at all, and because the pips sat on top of the artwork they could not simply be made permanent. This replaces them with a segmented track directly beneath the hit point bar: always visible, always in the same place as the other vitals.

image

image

image

## Changes

- Add `WoundTrack.svelte` alongside the other sheet vitals components (`HitPointBar`, `ManaBar`), rendered into a new `woundTrack` row of the defense grid scoped to the HP column
- Remove the portrait pip rail and the now-redundant wound chip in the Hit Points heading — the track's own label carries the count
- Hovering segment N previews the resulting total by lighting segments 1..N, via `:has(~ …)`; already-earned wounds keep their full colour underneath the preview
- Delete `_wounds.scss` and the dead `--nimble-wound-button-opacity` / `--nimble-wound-button-point-events` hover block in `_image.scss` that only existed to drive it
- Add themed `--nimble-wound-track-*` tokens to both `_variables.scss` (light) and `_theme.scss` (dark); no colour literals in the component
- Label reuses `.nimble-heading--wounds` so its size, weight, casing and colour track Hit Points automatically, with a `fa-droplet` icon matching the wound glyph already used by the combat tracker and minion token actions
- Flush the `0.75rem` hp/mana heading indent inside the defense section so every label and box in the block shares one left edge; scoped locally because `--hp` is shared with the NPC sheet
- Replace the hardcoded `"Toggle Wound"` string with `NIMBLE.wounds.*` keys

Net −166 lines of source.

## Test Plan

1. Open a character sheet. The wound track is visible under the HP bar at zero wounds — this is the main thing the old design could not do.
2. Click the 3rd segment → 3 wounds, label reads `WOUNDS 3/6`.
3. Click the 3rd segment again → drops to 2. The existing toggle-back behaviour from `toggleWounds()` is unchanged.
4. Hover a segment → it and everything to its left light up as a preview; segments already earned stay at full colour.
5. Click a segment, then move the mouse away — no leftover focus ring. Then `Tab` onto a segment — a focus ring *should* appear.
6. For the >9 case, set `system.attributes.wounds.bonus` to 4 on any character (max becomes 10). The track stays a single flush row; the old rail wrapped into two columns past 9.

Verified live in Foundry v13 driving the real DOM: 6 segments at max 6, click → 3, re-click → 2, click last → 6, and 10 segments at max 10. `pnpm check` passes (format, lint, circular deps, type-check, 2863 tests).

## Checklist

### Notes on the unchecked box

No unit tests were added. The wound mutation logic lives in `toggleWounds()` in `PlayerCharacterSheet.state.svelte.ts` and is untouched by this PR; `WoundTrack.svelte` is presentational and takes the toggle as a prop. A component test in the `CustomReactionsPanel.test.ts` + `testHarness.svelte` style could cover segment count, filled count and the emitted wound level — happy to add one if you'd like it before merge.

### Reviewer notes

- **Light mode is unverified.** Light-theme token values are real (not copies of the dark ones) and were picked for contrast against the light sheet background, but every screenshot and live check was in dark mode. Worth an eyeball.
- `en.json` gains `NIMBLE.wounds.heading` / `NIMBLE.wounds.toggle`. English only, flagged for human review per AGENTS.md; no other locale files touched.


trevlar (changes requested):
**🔧 Push back — the new theme tokens don't meet the style guide's 3:1 contrast floor, and the fill/empty distinction is the whole point of the track**

The structural work here is good — the move into the defense grid is clean, the deletions are correct, and your specificity comment on the hp/mana indent is right (I compiled it: the override lands at `(0,4,0)` against the global rule's `(0,3,0)`). The problem is the colour values. You flagged light mode as unverified; it turns out dark mode has the bigger issue.

All numbers below are WCAG relative-luminance contrast ratios computed from the token values in this diff. The two segment-vs-segment ones don't depend on the sheet background at all.

**Must fix before merge**

- **`src/scss/base/_theme.scss:199`** — In dark mode, filled vs empty segment is **1.24:1**. `--nimble-wound-track-filled-color: hsl(1, 62%, 42%)` (`#ae2b29`, L=0.108) and `--nimble-wound-track-empty-background: hsl(293, 12%, 44%)` (`#7b637e`, L=0.145) are almost identical in luminance — the distinction is carried by hue alone. The gradient's mid-tone is no better at 1.29:1. [docs/STYLE_GUIDE.md](docs/STYLE_GUIDE.md) requires **3:1** for "UI components and graphics", and this is the single signal the track exists to convey. `--nimble-wound-track-empty-background: hsl(293, 10%, 70%)` gets you to 3.02:1 against the filled red and 8.05:1 against the dark sheet background.

- **`src/view/sheets/components/WoundTrack.svelte:87`** — The hover preview is effectively invisible. The 55% `color-mix` against the empty background measures **1.20:1** in dark and **2.14:1** in light. This is the headline interaction of the PR and it can't be perceived in either theme. Note that fixing the token above only brings the mix to ~1.9:1, because a 55% blend lands halfway between the two colours by construction — push the mix well higher (~85%) or give the preview its own token rather than deriving it.

- **`src/scss/base/_variables.scss:261`** — In light mode the empty segment (`hsl(293, 8%, 74%)`, `#c1b7c2`) is **1.71:1** against the light sheet background (measured against the `hsl(40, 20%, 94%)` fallback in `--nimble-sheet-background`). At zero wounds the track is barely visible — which is precisely the thing this redesign exists to fix. Worth knowing before you pick values: no flat grey satisfies both `empty-vs-background` and `filled-vs-empty` at 3:1 in light mode, because the filled red is dark (L=0.092) and the background is light (L=0.879). Give the segments a real 1px border token so the track's extent is defined independently of the fill.

- **`src/view/sheets/components/WoundTrack.svelte:123`** — `padding-block-end: 0` is dead, and the comment above it asserts it isn't. It compiles to `.nimble-wound-track__label.svelte-hash` at `(0,2,0)`, which loses to `.system-nimble .nimble-sheet .nimble-heading--wounds` at `(0,3,0)` setting `0.25rem` — so the label keeps 4px of padding you think you removed. Either drop the declaration and fold the space into `margin-block`, or add `.nimble-heading` to the compound the same way you did for hp/mana in the sheet.

- **`src/view/sheets/components/WoundTrack.svelte:106`** — Hovering a segment at or below the current count gives no preview at all (the `--filled` rules correctly win over the preview rules), but `toggleWounds()` sets `woundLevel - 1` in that case. So hovering segment 3 at 5 wounds shows nothing and clicking drops you to 2. The preview is silent exactly in the heal-a-wound case. Consider previewing the un-fill when `woundLevel <= currentWounds`.

**What's good**

- Net −166 lines, and the deletions are genuinely dead — the `--nimble-wound-button-*` block in `_image.scss` and the `.nimble-monster-tracker .nimble-wounds-list` selector both had no remaining markup. I grepped for every removed class and found no orphans.
- The `--filled` / `:has(~ …)` source-order reasoning is correct, including the non-obvious part: Svelte scopes the `:has()` argument with a zero-specificity `:where()`, so `.seg:has(~ .seg:hover)` reaches `(0,4,0)` and genuinely needs the enumerated `--filled:has(…)` variants to be overridden.

---

1. Raise dark-mode filled-vs-empty segment contrast to 3:1
2. Make the hover preview actually visible in both themes
3. Define the track's extent in light mode (border token)
4. Remove or fix the no-op `padding-block-end` on the label
5. Preview the un-fill when hovering at or below current wounds

Fronix:
> **🔧 Push back — the new theme tokens don't meet the style guide's 3:1 contrast floor, and the fill/empty distinction is the whole point of the track**

image
image
image

Issue #582, opened 2026-04-08
[Foundation] Action-economy cost & grant rules


Fronix:
## Purpose

Tracking issue for the action-economy foundation: the cost and grant rules that unblock Commander, Songweaver and Zephyr flagship features, plus parts of Mage and The Cheat.

Rescoped 2026-07-31 — the original `grantAction` proposal was rejected; see the comment below for the rationale. A rulebook sweep across all five books found the demand splits four ways, and the literal "+1 action" grant of the original title is the *smallest* bucket:

| Bucket | Cases | Served by |
|---|---|---|
| A. Activation cost override ("costs 0 / 1 less / 1 more") | 31 | `actionCost` rule (#882, #883) |
| B. Free standard maneuver | 42 | costs axis here (#881, #883); the 21 free-Move cases are primitive 02, v14-gated, out of scope |
| C. Cross-actor grant (incl. hostile denial) | 20 | `actionDelta` targeting + GM relay (#884); activation offers (#885) |
| D. Raw pool delta ("+1 action") | 21 | `actionDelta` (#884); caps already ship as `dyingActionLimit` |

Key design decisions:

- **Cost and use-limit are orthogonal axes** (CoreRules: "If you've already Defended this round, you can't use abilities requiring it again *(even for free)*"). The limit axis already ships (`chargePool` recovery triggers, `*Available` flags); the new work is the cost axis only, and the new rules never touch limit state.
- **Soft-block, not hard gate**: costing more actions than remain prompts a confirmation with a force override, following the existing reaction-confirmation pattern.
- `type: 'action', quantity: 0` is the representation for "free" — no new cost type.

## Children

- [ ] #881 — allow zero-cost activations (PR #876)
- [ ] #882 — `actionCost` rule for item activations (PR #877)
- [ ] #883 — `actionCost` for heroic reactions + insufficient-action confirmation (PR #878)
- [ ] #884 — `actionDelta` rule + pending action grants (PR #879)
- [ ] #885 — `grantActivation` granted activation offers (PR #880)

## Explicitly out of scope

- **Free Move** — primitive 02 (one-shot movement budget), v14-gated pending the native movement cost model.
- **Grim Coronet's "take a turn immediately"** — automate the +3 actions; the extra turn stays a GM control.
- **Duration-based `max` changes** (Dragonform) — `max` is deliberately never touched; `addAdditionalAction` covers it.
[...]

PR #876, opened 2026-07-31
feat(items): zero-cost activations and activation cost labels

Fronix:
## Summary

Makes a zero action cost representable so features worded "for free" stop debiting an action. Previously `activation.cost.quantity` was floored at 1 in the schema, in `resolveItemActionCost`, and again in `consumeCombatantAction`, so every free-use feature charged 1 action that the player had to refund by hand.

Review follow-up then pulled in the activation cost label itself: six sites rendered it by hand, which is why the "Free" case had to be added to each one separately, and two cost shapes rendered wrong or not at all.

### Zero-cost activations

- `activation.cost.quantity` schema `min: 1` -> `min: 0` (duration untouched; no migration needed since relaxing a `min` is strictly widening). The config tab keeps the per-unit floor: `min` is 0 for `action`, 1 for the elapsed-time units.
- `resolveItemActionCost` new contract: returns `0` for non-action cost types and an explicit `quantity: 0`; `1` remains only the default for a *missing* quantity on an action-type cost. This also fixes a latent bug where `type: 'none'` items reported cost 1.
- `resolveMinionAttackActionCost` is a separate helper for group attacks, so the two contracts stay distinct: a missing or `none` cost defaults to 1 (the legacy pack convention), while an explicit `{type: 'action', quantity: 0}` stays free. The minion gate in `combatCommon.ts` and both consumption sites use it, so the skip check still fires for all 184 shipped `monsterFeature`s carrying `cost.type: 'none'`.
- `consumeCombatantAction` drops its `cost >= 1` floor, early-returns with no write at cost 0, and routes its write through `queueCombatantMutationWithFreshDocument`. The current-actions read and the `Math.max(0, current - cost)` now happen *inside* the mutation callback against the fresh combatant, so two overlapping deductions from 3 serialize 3 -> 2 -> 1 instead of both landing on 2.
- `NimbleCharacter.activateItem` and `activateItemMacro` both skip the deduction when the resolved cost is 0, the macro by calling `resolveItemActionCost` rather than rebuilding the contract inline.

### Activation cost labels

- New `formatActivationCostLabel()` in `src/utils/`, now the single source for "1 Action", "2 Actions", "10 Minutes" and "Free". Replaces six hand-rolled copies (`PlayerCharacterSpellsTab`, `CastSpellActionPanel`, `SpellReferenceCard`, `LevelUpSpellCard`, `generatePdfContent`, `CustomReactionsPanel`), three of which had no test.
- The `${quantity || 1}` falsy-zero fallback is gone from every call site. It survives once inside the formatter, guarding the elapsed-time units only: `min: 0` is shared across every cost type and Foundry's `NumberField.min` clamps rather than rejects, so a zero can still arrive through pack JSON or a migration, and "1 Minute" beats "0 Minutes".
- **Reactions now label correctly.** Four sheets read `activationCostTypes['reaction']`, but that map has no `reaction` key, so anything authored against the legacy `type: 'reaction'` shape rendered the literal string `undefined`. Reactions read as "Reaction", "Reaction (2 Actions)" or "Free Reaction" for both shapes (`isReaction` on an action cost, and the legacy cost type). One reaction costs N actions rather than being a resource you spend several of, so the count never pluralises the reaction itself and the cost is only spelled out when it differs from the default of 1 (CoreRules-2:438, "Reactions cost 1 action"). `CustomReactionsPanel` deliberately opts out and keeps showing the bare cost, since every row there is already a reaction.
- The spells tab surfaced its reaction-trigger tooltip only for the legacy cost type, so a reaction authored the current way had no trigger on hover despite the config tab offering the field. It now keys off either shape.
- **`turn` leaves the cost dropdown.** "1/turn" is how often an ability may be used, not what it costs to use it, so it belongs to a usage limit rather than to `activationCostTypes` (CoreRules-2:500 "1/Turn Abilities", :818, GMguide-2:1662 `ACTION: (1/turn)`). It was selectable but rendered no label at all, and nothing in `packs/` authored it (0 of 1032 authored activation costs, counting items embedded in monster actors). Frequency stays unmodelled.
- The PDF export printed "1 special" for the one shipped spell with a `special` cost (Redeem), since its guard excluded only `none` and `mana`. Only `extractSpellDetails` renders an activation cost, so the five features that also carry a `special` cost never hit this path. Those types carry no quantity, so the formatter returns `null` and the caller prints the bare type label. Side effect worth a look: casting times in the PDF are now capitalized ("1 Action AOE" rather than "1 action AOE"), which matches the "Free"/"Range 1-5"/"1 Mana" already around them.
- English strings only, flagged for review: added `activationCosts.free`, `.reaction`, `.freeReaction`, `.reactionWithCost`; removed the now-unused `activationCosts.turn` and `activationCostsPlural.turn`.

No new cost type was added for "free": `{type: 'action', quantity: 0}` is the representation, keeping the item classified as an action-economy item (it still shows in the relevant panels) while costing nothing.

## Test Plan

- `pnpm check` green on this branch, and on the top of the stack (#880) after restacking: 187 files, 3029 tests.
- New unit coverage: `formatActivationCostLabel.test.ts` (13 cases, including the reaction shapes and the legacy cost type that used to render `undefined`), `resolveMinionAttackActionCost.test.ts`, and `combatCommon.test.ts` (9 cases, including `cost.type: 'none'` at `currentActions: 0`, which had no test before and is why CI stayed green through the earlier behavior flip).
- Live-verified in Foundry v13 (world session, real UI clicks): a feature with `{type:'action', quantity: 0}` activated in combat leaves the action pips untouched and still posts its chat card; the same feature at quantity 1 debits exactly 1.
- **Not yet live-verified:** the label changes in the last two commits are unit-tested only. Worth a reviewer eyeballing the spell list, the level-up spell cards, the character-creator spell cards and a PDF export.

Reviewer repro:

1. Author a feature, set its activation cost quantity to 0 in the config tab, activate it during the character's combat turn, and watch the action tracker.
2. Check "Is Reaction" on a spell with a 1-action cost and look at the chip in the spell list: it should read "Reaction", not "1 Action". Set the quantity to 2 for "Reaction (2 Actions)", 0 for "Free Reaction". Add trigger text under Activation > Core and hover the info icon.
3. Open the activation cost dropdown and confirm "Turn" is no longer offered.

## Checklist

- [x] Tested in Foundry with no console errors (zero-cost activation path; see the Test Plan note on the label changes)

## Related Issues

Relates to #582 (first of five PRs for the action-economy foundation).

### DIFF
--- a/src/documents/actor/character.ts
+++ b/src/documents/actor/character.ts
@@ -16,7 +16,9 @@ import type { NimbleCharacterData } from '../../models/actor/CharacterDataModel.
 import calculateRollMode from '../../utils/calculateRollMode.js';
 import { consumeCombatantAction } from '../../utils/combatTurnActions.js';
 import getRollFormula from '../../utils/getRollFormula.js';
-import resolveItemActionCost from '../../utils/resolveItemActionCost.js';
+import resolveItemActionCost, {
+	type ItemWithActivationCost,
+} from '../../utils/resolveItemActionCost.js';
 import CharacterArmorProficienciesConfigDialog from '../../view/dialogs/CharacterArmorProficienciesConfigDialog.svelte';
 import CharacterLanguageProficienciesConfigDialog from '../../view/dialogs/CharacterLanguageProficienciesConfigDialog.svelte';
 import CharacterLevelDownDialog from '../../view/dialogs/CharacterLevelDownDialog.svelte';
@@ -1846,9 +1848,7 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
 		const result = await super.activateItem(id, options);
 
 		if (result && item && !options.skipActionDeduction) {
-			const actionCost = resolveItemActionCost(
-				item as unknown as Parameters<typeof resolveItemActionCost>[0],
-			);
+			const actionCost = resolveItemActionCost(item as ItemWithActivationCost);
 			if (actionCost > 0) {
 				const combat = game.combat as Combat | null;
 				const combatant =
--- a/src/documents/actor/character.ts
+++ b/src/documents/actor/character.ts
@@ -16,6 +16,7 @@ import type { NimbleCharacterData } from '../../models/actor/CharacterDataModel.
 import calculateRollMode from '../../utils/calculateRollMode.js';
 import { consumeCombatantAction } from '../../utils/combatTurnActions.js';
 import getRollFormula from '../../utils/getRollFormula.js';
+import resolveItemActionCost from '../../utils/resolveItemActionCost.js';
 import CharacterArmorProficienciesConfigDialog from '../../view/dialogs/CharacterArmorProficienciesConfigDialog.svelte';
 import CharacterLanguageProficienciesConfigDialog from '../../view/dialogs/CharacterLanguageProficienciesConfigDialog.svelte';
 import CharacterLevelDownDialog from '../../view/dialogs/CharacterLevelDownDialog.svelte';
@@ -1845,10 +1846,10 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
 		const result = await super.activateItem(id, options);
 
 		if (result && item && !options.skipActionDeduction) {
-			const activation = (
-				item.system as { activation?: { cost?: { type: string; quantity: number } } }
-			).activation;
-			if (activation?.cost?.type === 'action') {
+			const actionCost = resolveItemActionCost(
+				item as unknown as Parameters<typeof resolveItemActionCost>[0],
+			);
+			if (actionCost > 0) {
 				const combat = game.combat as Combat | null;
 				const combatant =
 					combat?.combatants?.find(
@@ -1858,7 +1859,7 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
 					await consumeCombatantAction({
 						combat,
 						combatantId: combatant.id,
-						actionCost: activation.cost.quantity ?? 1,
+						actionCost,
 					});
 				}
 			}
--- a/src/models/item/common.ts
+++ b/src/models/item/common.ts
@@ -21,7 +21,7 @@ export const activation = () => ({
 				required: true,
 				initial: 1,
 				nullable: false,
-				min: 1,
+				min: 0,
 				integer: true,
 			}),
 			type: new fields.StringField({
--- a/src/utils/combatTurnActions.ts
+++ b/src/utils/combatTurnActions.ts
@@ -375,11 +375,9 @@ export async function requestSwapCombatTurn(params: {
 export async function consumeCombatantAction(params: {
 	combat: Combat;
 	combatantId: string;
-	fallbackCombatant?: Combatant.Implementation | null;
 	actionCost?: number;
 }): Promise<number> {
-	const combatant =
-		params.combat.combatants.get(params.combatantId) ?? params.fallbackCombatant ?? null;
+	const combatant = params.combat.combatants.get(params.combatantId) ?? null;
 	if (!combatant) return 0;
 
 	// With action-tracking automation off, item use never deducts actions. The
@@ -387,18 +385,26 @@ export async function consumeCombatantAction(params: {
 	// callers must not branch on it.
 	if (!isActionTrackingAutomationEnabled()) return getCombatantCurrentActions(combatant);
 
-	const currentActions = getCombatantCurrentActions(combatant);
-	if (currentActions < 1) return 0;
+	// Cheap early-out on a possibly stale snapshot. The queued mutation below
+	// recomputes from the fresh document before writing, so at worst a stale
+	// zero here skips a deduction; it can never cause an incorrect write.
+	const snapshotActions = getCombatantCurrentActions(combatant);
+	if (snapshotActions < 1) return 0;
 
 	const cost = Number(params.actionCost ?? 1);
 	const normalizedCost = Number.isFinite(cost) && cost >= 0 ? cost : 1;
-	if (normalizedCost === 0) return currentActions;
+	if (normalizedCost === 0) return snapshotActions;
 
-	const nextActions = Math.max(0, currentActions - normalizedCost);
 	const appliedActions = await queueCombatantMutationWithFreshDocument({
 		combat: params.combat,
 		combatantId: params.combatantId,
 		mutation: async (currentCombatant) => {
+			// Recompute from the fresh document so overlapping deductions each
+			// apply on top of the latest persisted value instead of a shared
+			// pre-queue snapshot clobbering one another.
+			const currentActions = getCombatantCurrentActions(currentCombatant);
+			const nextActions = Math.max(0, currentActions - normalizedCost);
+			if (nextActions === currentActions) return currentActions;
 			await currentCombatant.update({
 				[COMBATANT_ACTIONS_CURRENT_PATH]: nextActions,
 			} as Record<string, unknown>);
@@ -408,7 +414,7 @@ export async function consumeCombatantAction(params: {
 	// A skipped write (the combatant vanished before the queued mutation ran)
 	// leaves the action count unchanged, so report the value that is actually
 	// persisted rather than the deduction we intended to make.
-	return appliedActions ?? currentActions;
+	return appliedActions ?? snapshotActions;
 }
 
 export async function maybeAdvanceTurnForCombatant(params: {
--- a/src/utils/combatTurnActions.ts
+++ b/src/utils/combatTurnActions.ts
@@ -2,6 +2,7 @@ import { isActionTrackingAutomationEnabled } from '../settings/automationSetting
 import { getActorDyingActionLimit, isActorDying } from './actorHealthState.js';
 import { combatantActionMutationQueue } from './combatantActionMutationQueue.js';
 import { isCombatantDead } from './isCombatantDead.js';
+import { queueCombatantMutationWithFreshDocument } from './queueCombatantMutationWithFreshDocument.js';
 
 export const COMBATANT_ACTIONS_CURRENT_PATH = 'system.actions.base.current';
 export const COMBATANT_ACTIONS_MAX_PATH = 'system.actions.base.max';
@@ -390,13 +391,19 @@ export async function consumeCombatantAction(params: {
 	if (currentActions < 1) return 0;
 
 	const cost = Number(params.actionCost ?? 1);
-	const normalizedCost = Number.isFinite(cost) && cost >= 1 ? cost : 1;
+	const normalizedCost = Number.isFinite(cost) && cost >= 0 ? cost : 1;
+	if (normalizedCost === 0) return currentActions;
+
 	const nextActions = Math.max(0, currentActions - normalizedCost);
-	const actionUpdate: Record<string, unknown> = {
-		_id: params.combatantId,
-		[COMBATANT_ACTIONS_CURRENT_PATH]: nextActions,
-	};
-	await params.combat.updateEmbeddedDocuments('Combatant', [actionUpdate]);
+	await queueCombatantMutationWithFreshDocument({
+		combat: params.combat,
+		combatantId: params.combatantId,
+		mutation: async (currentCombatant) => {
+			await currentCombatant.update({
+				[COMBATANT_ACTIONS_CURRENT_PATH]: nextActions,
+			} as Record<string, unknown>);
+		},
+	});
 	return nextActions;
 }
 
--- a/src/utils/combatTurnActions.ts
+++ b/src/utils/combatTurnActions.ts
@@ -395,16 +395,20 @@ export async function consumeCombatantAction(params: {
 	if (normalizedCost === 0) return currentActions;
 
 	const nextActions = Math.max(0, currentActions - normalizedCost);
-	await queueCombatantMutationWithFreshDocument({
+	const appliedActions = await queueCombatantMutationWithFreshDocument({
 		combat: params.combat,
 		combatantId: params.combatantId,
 		mutation: async (currentCombatant) => {
 			await currentCombatant.update({
 				[COMBATANT_ACTIONS_CURRENT_PATH]: nextActions,
 			} as Record<string, unknown>);
+			return nextActions;
 		},
 	});
-	return nextActions;
+	// A skipped write (the combatant vanished before the queued mutation ran)
+	// leaves the action count unchanged, so report the value that is actually
+	// persisted rather than the deduction we intended to make.
+	return appliedActions ?? currentActions;
 }
 
 export async function maybeAdvanceTurnForCombatant(params: {
--- a/src/utils/resolveItemActionCost.test.ts
+++ b/src/utils/resolveItemActionCost.test.ts
@@ -2,25 +2,36 @@ import { describe, expect, it } from 'vitest';
 import resolveItemActionCost from './resolveItemActionCost.js';
 
 describe('resolveItemActionCost', () => {
-	it('returns 1 when item is null', () => {
-		expect(resolveItemActionCost(null)).toBe(1);
+	it('returns 0 when item is null', () => {
+		expect(resolveItemActionCost(null)).toBe(0);
 	});
 
-	it('returns 1 when system is missing', () => {
-		expect(resolveItemActionCost({} as Parameters<typeof resolveItemActionCost>[0])).toBe(1);
+	it('returns 0 when system is missing', () => {
+		expect(resolveItemActionCost({} as Parameters<typeof resolveItemActionCost>[0])).toBe(0);
 	});
 
-	it('returns 1 when activation is missing', () => {
-		expect(resolveItemActionCost({ system: {} })).toBe(1);
+	it('returns 0 when activation is missing', () => {
+		expect(resolveItemActionCost({ system: {} })).toBe(0);
 	});
 
-	it('returns 1 when cost is missing', () => {
-		expect(resolveItemActionCost({ system: { activation: {} } })).toBe(1);
+	it('returns 0 when cost is missing', () => {
+		expect(resolveItemActionCost({ system: { activation: {} } })).toBe(0);
 	});
 
-	it('returns 1 when cost type is not action', () => {
+	it('returns 0 when cost type is not action', () => {
 		expect(
 			resolveItemActionCost({ system: { activation: { cost: { type: 'bonus', quantity: 2 } } } }),
+		).toBe(0);
+		expect(
+			resolveItemActionCost({ system: { activation: { cost: { type: 'none', quantity: 1 } } } }),
+		).toBe(0);
+	});
+
+	it('returns 1 when quantity is missing on an action cost', () => {
+		expect(
+			resolveItemActionCost({
+				system: { activation: { cost: { type: 'action' } } },
+			}),
 		).toBe(1);
 	});
 
@@ -40,12 +51,12 @@ describe('resolveItemActionCost', () => {
 		).toBe(1);
 	});
 
-	it('returns 1 when quantity is zero', () => {
+	it('returns 0 when quantity is explicitly zero', () => {
 		expect(
 			resolveItemActionCost({
 				system: { activation: { cost: { type: 'action', quantity: 0 } } },
 			}),
-		).toBe(1);
+		).toBe(0);
 	});
 
 	it('returns the correct quantity for valid action costs', () => {
--- a/src/utils/resolveItemActionCost.ts
+++ b/src/utils/resolveItemActionCost.ts
@@ -10,7 +10,7 @@ interface ItemWithActivationCost {
 
 export default function resolveItemActionCost(item: ItemWithActivationCost | null): number {
 	const cost = item?.system?.activation?.cost;
-	if (!cost || cost.type !== 'action') return DEFAULT_ACTION_COST;
+	if (!cost || cost.type !== 'action') return 0;
 	const quantity = Number(cost.quantity ?? DEFAULT_ACTION_COST);
-	return Number.isFinite(quantity) && quantity >= 1 ? quantity : DEFAULT_ACTION_COST;
+	return Number.isFinite(quantity) && quantity >= 0 ? quantity : DEFAULT_ACTION_COST;
 }
--- a/src/utils/resolveItemActionCost.ts
+++ b/src/utils/resolveItemActionCost.ts
@@ -1,6 +1,6 @@
 const DEFAULT_ACTION_COST = 1;
 
-interface ItemWithActivationCost {
+export interface ItemWithActivationCost {
 	system?: {
 		activation?: {
 			cost?: { type?: string; quantity?: number };
--- a/src/view/sheets/pages/ItemActivationCoreConfigTab.svelte
+++ b/src/view/sheets/pages/ItemActivationCoreConfigTab.svelte
@@ -47,6 +47,7 @@
 				{#if ['action', 'minute', 'hour'].includes(activationCostType)}
 					<input
 						type="number"
+						min="0"
 						value={activationCost}
 						onchange={({ target }) =>
 							document.update({
--- a/src/view/sheets/pages/ItemActivationCoreConfigTab.svelte
+++ b/src/view/sheets/pages/ItemActivationCoreConfigTab.svelte
@@ -47,7 +47,7 @@
 				{#if ['action', 'minute', 'hour'].includes(activationCostType)}
 					<input
 						type="number"
-						min="0"
+						min={activationCostType === 'action' ? 0 : 1}
 						value={activationCost}
 						onchange={({ target }) =>
 							document.update({
--- a/docs/STYLE_GUIDE.md
+++ b/docs/STYLE_GUIDE.md
@@ -1489,6 +1489,7 @@ export const myStore = writable<MyStoreState>(initialState);
 | `getSubclassChoices()` | `src/utils/getSubclassChoices.ts` | Get available subclass options |
 | `resolveItemActionCost()` | `src/utils/resolveItemActionCost.ts` | Get an item's activation action cost (0 for non-action cost types; missing quantity defaults to 1) |
 | `resolveMinionAttackActionCost()` | `src/utils/resolveMinionAttackActionCost.ts` | Action cost for group-attack activations; a missing or `none` cost defaults to 1 (legacy pack convention), an explicit zero stays free |
+| `formatActivationCostLabel()` | `src/utils/formatActivationCostLabel.ts` | Render an activation cost as "1 Action", "2 Actions", "10 Minutes" or "Free"; returns `null` for the types that carry no quantity so the caller can label them |
 | `spell/*` | `src/utils/spell/` | Spell-related utilities |
 | `treeManipulation/*` | `src/utils/treeManipulation/` | Tree data structure utilities |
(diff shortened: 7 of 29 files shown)


src/utils/resolveItemActionCost.ts
```diff
@@ -10,7 +10,7 @@ interface ItemWithActivationCost {
 
 export default function resolveItemActionCost(item: ItemWithActivationCost | null): number {
 	const cost = item?.system?.activation?.cost;
-	if (!cost || cost.type !== 'action') return DEFAULT_ACTION_COST;
+	if (!cost || cost.type !== 'action') return 0;
```
trevlar, line 13: Don't reactions also cost 1?


My AI told me this:

returning `0` for `type: 'none'` breaks `resolveMinionAttackSkipReason` (`combatCommon.ts:194`, unchanged in this PR). It feeds the result in as `requiredActions`, so the guard at `combatCommon.ts:198` becomes `currentActions >= 0` — always true. A minion group member with **0 actions left can now attack, every time, without limit**. The gate is gone, not just the debit. I counted the packs on the `dev` tip: 184 of the 190 `monsterFeature` items with `subtype: 'action'` carry `cost.type: 'none'`; only 6 use `action`. So this hits essentially every shipped monster. Your PR body flags the "no longer debits" half of this, but not that the block itself stops firing.
  Either keep the gate's floor at the call site (`Math.max(1, resolveItemActionCost(...))`, or a separate `resolveRequiredActions` helper so the two contracts stay distinct), or land the pack-data migration setting `type: 'action'` on those features in this same PR. Worth noting the config tab hides the activation-cost editor for `monsterFeature` entirely, so GMs can't correct this on their own monsters from the sheet.

src/utils/resolveItemActionCost.test.ts
trevlar, line 1: `resolveMinionAttackSkipReason` has no test file anywhere, which is why CI stayed green through a behavior flip this large. Whichever way you resolve the above, it needs a test covering `cost.type: 'none'` at `currentActions: 0`.

src/models/item/common.ts
```diff
@@ -21,7 +21,7 @@ export const activation = () => ({
 				required: true,
 				initial: 1,
 				nullable: false,
-				min: 1,
+				min: 0,
```
trevlar, line 24: I think this is accurate to represent free actions. There are places in our code where we fall through to a default of 1.

AI told me this:

relaxing `min` to `0` makes `quantity: 0` representable, but four spell UIs render it with `${activationCost || 1}`, so an explicit `0` is falsy and prints **"1 Action"**. The item costs nothing while the sheet tells the player it costs one. Sites: `PlayerCharacterSpellsTab.svelte:56`, `CastSpellActionPanel.svelte.ts:110`, `SpellReferenceCard.svelte.ts:82`, `LevelUpSpellCard.svelte.ts:29`. You added `Free` handling to `CustomReactionsPanel` — these four need the same, and `NIMBLE.activationCosts.free` is already there for it. `generatePdfContent.ts:314` has the same bug in a different shape (`cost.quantity > 1 ? ... : '1 ${cost.type}'`).

src/utils/combatTurnActions.ts
```diff
@@ -384,13 +385,19 @@ export async function consumeCombatantAction(params: {
...
+	const normalizedCost = Number.isFinite(cost) && cost >= 0 ? cost : 1;
+	if (normalizedCost === 0) return currentActions;
+
 	const nextActions = Math.max(0, currentActions - normalizedCost);
-	const actionUpdate: Record<string, unknown> = {
-		_id: params.combatantId,
-		[COMBATANT_ACTIONS_CURRENT_PATH]: nextActions,
-	};
-	await params.combat.updateEmbeddedDocuments('Combatant', [actionUpdate]);
+	await queueCombatantMutationWithFreshDocument({
```
trevlar: Is this AI indicating a race condition?

Routing through `queueCombatantMutationWithFreshDocument` here doesn't do what the wrapper is for. `nextActions` is computed at line 391 from the *pre-queue* snapshot, so the fresh document you fetch inside the mutation is read and then ignored — two rapid 1-action activations from 3 both compute `2`, and the second write clobbers the first. Compare `combat.svelte.ts:871` (`useHeroicReactions`), which recomputes from the fresh combatant *inside* the mutation. Move the `getCombatantCurrentActions` read and the `Math.max(0, current - cost)` into the mutation callback.

trevlar: Additionally this one...

`fallbackCombatant` is now dead for the write path. The old `updateEmbeddedDocuments` call worked off the id alone, but `queueCombatantMutationWithFreshDocument` does its own `combat.combatants.get(id)` and returns `undefined` on a miss — yet `consumeCombatantAction` still returns `nextActions` as though the write landed. `minionGroupTokenActions.ts:2270` is passing a fallback that can no longer be used. Either thread it through or drop the parameter.

src/view/sheets/pages/ItemActivationCoreConfigTab.svelte
```diff
@@ -47,6 +47,7 @@
 				{#if ['action', 'minute', 'hour'].includes(activationCostType)}
 					<input
 						type="number"
+						min="0"
```
trevlar: Strange that this is shared with minute and hour which are not equations of cost.

src/documents/actor/character.ts
```diff
@@ -1845,10 +1846,10 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
 		const result = await super.activateItem(id, options);
 
 		if (result && item && !options.skipActionDeduction) {
-			const activation = (
-				item.system as { activation?: { cost?: { type: string; quantity: number } } }
-			).activation;
-			if (activation?.cost?.type === 'action') {
+			const actionCost = resolveItemActionCost(
+				item as unknown as Parameters<typeof resolveItemActionCost>[0],
```
trevlar: `item as unknown as Parameters<typeof resolveItemActionCost>[0]` is a double cast through `unknown`. Widening `ItemWithActivationCost` (or exporting it and accepting the structural shape) reads better and keeps the compiler useful here.

trevlar (changes requested):
**🔧 Push back — the `resolveItemActionCost` contract flip silently disables the minion action gate for 184 of the 190 shipped monster attack features, and free items still render as "1 Action" in every spell UI**

The core idea is right and the schema/`consumeCombatantAction` work is clean. The problem is downstream: `resolveItemActionCost` has three other callers written against the old "default 1" contract, and only the character path was updated. Those callers don't appear in the diff — they don't need to change for their behavior to change.

**What's good**

- Fixing `type: 'none'` → `0` is genuinely the right contract, and the `resolveItemActionCost` tests now read as a clear spec of it.
- Dropping the `cost >= 1` floor with an explicit no-write early return at 0 is the correct shape, and `combatTurnActions.test.ts` covers it properly.

---

1. Restore the minion `noActionsRemaining` gate (or migrate pack data)
2. Add tests for `resolveMinionAttackSkipReason` at 0 actions
3. Render "Free" instead of "1 Action" in the four spell UIs
4. Fix the same falsy-zero bug in the PDF export
5. Compute `nextActions` inside the fresh-document mutation
6. Make `fallbackCombatant` work again, or remove it
7. Only allow `min="0"` for the `action` cost type
8. Drop the `as unknown as` double cast in `character.ts`

src/utils/resolveItemActionCost.ts
Fronix: Yeah I'm not sure about this either, we do have free action stuff as well. Might be better to default to everything costing 1 action and it having to opt in to be free instead.

src/view/sheets/pages/ItemActivationCoreConfigTab.svelte
Fronix: I don't think this is an issue, this had no min at all before so this was just adding some strictness. Before you could theoretically type -9999

src/utils/resolveItemActionCost.ts
Fronix: Fixed a separate helper so the two contracts stay distinct: new src/utils/resolveMinionAttackActionCost.ts treats a missing cost or type: 'none' as 1 action (legacy pack convention) while an explicit {type: 'action', quantity: 0} stays free. It's now used at the gate in combatCommon.ts and both consumption sites (combatMinionAttacks.ts, minionGroupTokenActions.ts), so the block fires again for all 184 'none'-cost monster features.

One note: the comment's claim that the config tab hides the cost editor for monsterFeature is incorrect — MonsterFeatureSheet.svelte has an Activation tab with the cost editor, so GMs can set explicit costs on their own monsters.

src/utils/resolveItemActionCost.test.ts
Fronix: Fixed: new src/documents/combat/combatCommon.test.ts with 9 cases, including exactly the requested one (cost.type: 'none' at currentActions: 0 → noActionsRemaining), plus explicit-zero-at-0-actions → allowed, and the other skip reasons. The helper also has its own 6-case test file.

src/models/item/common.ts
Fronix: Fixed: all four spell UIs (PlayerCharacterSpellsTab, CastSpellActionPanel, SpellReferenceCard, LevelUpSpellCard) now render the localized "Free" for an explicit action-cost 0 instead of || 1-ing it, and generatePdfContent.ts prints "Free" as well. Missing quantity still defaults to 1.

trevlar (changes requested):
**🔧 Push back — six of eight landed cleanly, but the fresh-document race is still there and the five new "Free" display branches have no tests**

The `resolveMinionAttackActionCost` split is the right shape: two contracts, two helpers, both tested, and `combatCommon.test.ts` pins the `type: 'none'` at 0 actions case that was silently flipping. The gate is genuinely restored. Items 1, 2, 3, 4 and the double cast are all resolved, and the `min` on the config tab is now action-only.

Two things are still open — one of them a blocker from last round that didn't get a reply.

**Must fix before merge**

- **src/utils/combatTurnActions.ts:391** — `nextActions` is still computed from the pre-queue snapshot (`currentActions`, read at line 384) and then written inside the mutation, so the fresh combatant you fetch is read and discarded. Two overlapping activations from 3 actions both compute `2`; the queue serializes the writes but the second clobbers the first and the player keeps an action they spent. Reachable on a double-click of an action card, and it's the exact failure the wrapper exists to prevent. Move the `getCombatantCurrentActions` read and the `Math.max(0, current - normalizedCost)` inside the `mutation` callback and return that value — see `combat.svelte.ts:876` (`useHeroicReactions`), which recomputes `usageState` from the fresh combatant inside the mutation.

- **src/view/sheets/components/CustomReactionsPanel.test.ts:82** — the new `quantity === 0 → "Free"` branch at `CustomReactionsPanel.svelte.ts:49` has no test, even though this very test covers the sibling branches and `createItem` already accepts `quantity`. A `quantity: 0` case asserting `Free` renders is one line. Same gap in the four spell UIs and the PDF export — `SpellReferenceCard.svelte.test.ts` exists and its fixture already carries an `activation` block, so at minimum that one and the reactions panel need a zero case. As it stands the user-visible half of this PR is entirely unexercised by CI, which is what let the original falsy-zero bug through.

- **src/utils/combatTurnActions.ts:381** — you fixed the dishonest return value (returning `currentActions` on a skipped write is right, and the new test pins it), but `fallbackCombatant` now only feeds the pre-read. If the combatant isn't in the collection the guard passes off the fallback and then the write silently no-ops, so `minionGroupTokenActions.ts:2270` is passing it for nothing end to end. Either thread it into the mutation as the fallback target, or drop the parameter and the `latestCombatant` computation at the call site.

- **src/utils/resolveMinionAttackActionCost.ts:14** — the legacy default only catches a missing type and `'none'`; `minute`, `hour`, `special` and `turn` fall through to `resolveItemActionCost`, come back `0`, and re-open the same always-true gate for those types. Nothing in packs hits it today (I tallied `monsterFeature`: 184 `none`, 6 `action`, no others), but the docblock reads as "anything that isn't an explicit action cost defaults to 1", so `cost.type !== 'action'` matches the stated intent and closes the hole.

- **src/view/sheets/character/pdfExport/generatePdfContent.ts:316** — hardcoded `'Free'` while `NIMBLE.activationCosts.free` exists and the file already calls `game.i18n.localize` at line 619. Your checklist claims no hardcoded user-facing strings; this is the one.

**What's good**

- The two-helper split with a docblock explaining *why* the minion path defaults to 1 beats a `Math.max(1, ...)` at the call site — the two contracts stay legible instead of one being papered over.
- `combatCommon.test.ts` covers all five skip reasons, not just the one case I asked for.

---

1. Compute `nextActions` inside the fresh-document mutation
2. Test the new "Free" branches (reactions panel + spell card)
3. Thread or drop `fallbackCombatant`
4. Default every non-`action` cost type to 1 for minions
5. Localize the PDF export's `'Free'`

src/utils/combatTurnActions.ts
Fronix: Fixed in 83d220b8. The current-actions read and the `Math.max(0, current - cost)` computation now happen inside the `queueCombatantMutationWithFreshDocument` callback against the fresh combatant, and the function returns the value it actually persisted. Added a regression test asserting two overlapping deductions from 3 serialize 3→2→1 instead of both landing on 2. The pre-queue `< 1` check remains as a cheap early-out only: a stale zero can skip a deduction (self-correcting), but can never produce an incorrect write since the authoritative recompute is inside the mutation.

Fronix: Fixed in 83d220b8 by dropping the parameter (option 2), along with the `latestCombatant` computation at the `minionGroupTokenActions.ts` call site. Threading it into the mutation was the other option, but the wrapper's re-fetch by id from `combat.combatants` *is* its staleness guarantee — a fallback document could only ever receive a write outside that guarantee, i.e. exactly the stale write the wrapper exists to prevent. Dropping it leaves the honest contract: no combatant in the collection, no write.

Fronix:
All five items from the second review are addressed in 83d220b8:

1. **Fresh-document race** — the read and the deduction computation moved inside the mutation callback, returning the persisted value (per the `useHeroicReactions` pattern). New regression test pins overlapping deductions serializing 3→2→1. Replied in the thread with details.
2. **"Free" branch tests** — added `quantity: 0` → "Free" cases to `CustomReactionsPanel.test.ts` and `SpellReferenceCard.svelte.test.ts`, plus a new `CastSpellActionPanel.svelte.test.ts` (its factory exposes `getSpellMetadata` directly), and pinned the missing-quantity → "1 Action" default. Three sites remain untested: `LevelUpSpellCard` (`$effect` in the factory needs an effect root and the metadata logic is a non-exported internal), `PlayerCharacterSpellsTab` (logic inline in the component, needs a full sheet harness), and the PDF export (`extractSpellDetails` is internal and needs a full character actor). Happy to add harnesses for those in a follow-up if you want the coverage — flagging the gap rather than papering over it.
3. **`fallbackCombatant`** — dropped, plus the `latestCombatant` computation at the call site; rationale in the thread reply.
4. **Minion legacy default** — broadened to `cost.type !== 'action'` → 1, so `minute`/`hour`/`special`/`turn` can't re-open the gate; docblock restated to match, tests extended (`minute` q10 → 1, `special` q0 → 1, action-with-missing-quantity → 1).
5. **PDF `'Free'`** — now `game.i18n.localize('NIMBLE.activationCosts.free')`, matching the file's existing idiom. Checklist claim is true again.

`pnpm check` green at the bottom of the stack (2889 tests) and at the top (3014); all five branches restacked and pushed. One structural note: the unified in-mutation recompute now lives here in #876, so #879's "drive-by correctness fix" bullet no longer applies to its diff — its body has been updated accordingly.


---

**Live verification (Foundry v13, this branch @ 83d220b8).** All scenarios driven through the real UI (clicks on actual DOM elements); the API was used only to seed documents and read state for assertions.

- **Overlapping-deduction race:** two feature activations clicked back-to-back on a turn with 3 actions ended at **1** in 4/4 runs (the pre-fix clobber would leave 2); a 3-click variant ended at 0. Both/all chat cards posted each run.
- **Zero-cost activation:** a `{type:'action', quantity:0}` feature posts its card with no action decrement; a default-quantity feature consumes exactly 1.
- **Minion gate:** with actions available, a legacy `{type:'none'}` attack rolls and consumes exactly 1 (consumption still lands after the `fallbackCombatant` removal). At 0 actions: `'none'` and `'minute'` both skip with `noActionsRemaining` (tried the `'none'` one twice — the unlimited-attack repro is closed), while an explicit `{type:'action', quantity:0}` attack succeeds and consumes nothing.
- **"Free" rendering:** spells tab, Cast-a-Spell panel, and Custom Reactions panel all show the localized "Free" for zero-cost items; a default spell still shows "1 Action".
- **PDF export:** reached through the sheet's export dialog; the column content shows the localized "Free" (the generated jsPDF binary itself was not parsed).

One UI observation for a possible follow-up, not a regression: the minion panel's shared Roll button disables when every member is at 0 actions, so a lone free-cost minion at 0 can't fire from the panel even though the engine allows it; in a mixed group it rolls fine.

No console errors attributable to this PR (two pre-existing world-data errors noted and excluded).

trevlar (changes requested):
**🔧 Push back — the zero case needs its own branch by unit, not one nested inside the joined one**

The runtime pipeline is solid and last round's five items all landed: `nextActions` is computed inside the fresh-document mutation, `fallbackCombatant` is gone (a no-op anyway — `minionGroupTokenActions.ts:2169` resolved it from the same `combatants.get`), the PDF export localizes through `NIMBLE.activationCosts.free`, and `resolveMinionAttackActionCost` defaults every non-`action` type to 1. Multi-action costs work end to end — `resolveItemActionCost` returns the authored quantity and `consumeCombatantAction` deducts it, both tested.

**Must fix before merge**

- **src/view/sheets/pages/PlayerCharacterSpellsTab.svelte:51** — `if (['action', 'minute', 'hour'].includes(activationType))` runs one quantity path for every unit, then nests `if (activationType === 'action' && activationCost === 0)` inside it. All of these are casting times, so the line to draw isn't cost vs. time — it's the unit. Action-economy units admit zero, and zero is exactly what "free" means; elapsed-time units (`minute`, `hour`, and any `second` later) don't — there is no 0 minutes and no 0 hours. Give each its own branch: the action branch handles 0 → Free, 1 → Action, 2+ → Actions and needs no `|| 1`; the elapsed-time branch handles quantity ≥ 1 and has no Free case at all. Same shape at `SpellReferenceCard.svelte.ts:76`, `CastSpellActionPanel.svelte.ts:104`, `LevelUpSpellCard.svelte.ts:24` and `generatePdfContent.ts:314`.

  Once split, the `${activationCost || 1}` falsy-zero fallback at `PlayerCharacterSpellsTab.svelte:61` and `LevelUpSpellCard.svelte.ts:30` is dead, and the blanket `min: 0` at `common.ts:24` stops being load-bearing for the timed types. `resolveItemActionCost` already separates the units correctly (`type !== 'action'` → 0); the formatters are the last place they're still run through one path.

- **src/view/sheets/pages/PlayerCharacterSpellsTab.svelte:52** — do the split once. The Free branch is copy-pasted six times (the five above plus `CustomReactionsPanel.svelte.ts:49`), and three of the six have no test. A shared `formatActivationCostLabel()` in `src/utils/` turns the must-fix into a one-place change and makes the missing coverage a single test file.

- **src/view/sheets/pages/ItemActivationCoreConfigTab.svelte:39** — `{#if document.type !== 'monsterFeature'}` hides the whole Activation Cost section, and line 53 is the only activation-cost editor in the system, so a free monster feature is only authorable via pack JSON. The runtime handles it fine; it's purely the missing affordance. Probably a later PR in the #582 series, but worth a line in the body.

- **src/macros/activateItemMacro.ts:45** — still hand-rolls the contract you centralized (`type === 'action'` then `quantity ?? 1`). Correct today only because `deductActionPips(actor, 0)` writes an unchanged value that Foundry diffs away. `resolveItemActionCost(item)` with an `> 0` guard would match `NimbleCharacter.activateItem:1851`.

**What's good**

- Free action reactions fall out for free: `isReaction` stays orthogonal to the cost, so `{type: 'action', quantity: 0, isReaction: true}` resolves to 0 and renders "Free" without a special case. Worth keeping the test at `CustomReactionsPanel.test.ts:95` — it's the only thing pinning that interaction.
- The two-helper resolver split is the right shape, and both the docblock and the STYLE_GUIDE entry say *why* the legacy pack default exists — that's the difference between a shim someone deletes in six months and one they understand.

---

1. Branch casting-time labels by unit: action-economy vs elapsed time.
2. Drop the `|| 1` fallback the split makes redundant.
3. Extract one shared label formatter and test it.
4. Expose the activation cost editor for monster features (or defer explicitly).
5. Use `resolveItemActionCost` in `activateItemMacro`.

Fronix:
Thanks for the second pass. Taking these slightly out of order, because your non-blocking #1 turned out to be the one that mattered.

**3. Extract one shared label formatter and test it — done, and it subsumes the must-fix.**

New `formatActivationCostLabel()` in `src/utils/`, now the single source for "1 Action" / "2 Actions" / "10 Minutes" / "Free". All six copies route through it (`PlayerCharacterSpellsTab`, `CastSpellActionPanel`, `SpellReferenceCard`, `LevelUpSpellCard`, `generatePdfContent`, `CustomReactionsPanel`), with 13 cases in `formatActivationCostLabel.test.ts` covering the three that had no test. You filed this as non-blocking, but it was the real defect: six copies is *why* the "Free" branch had to be bolted onto each site separately last round.

**2. Drop the `|| 1` fallback — done at every call site, kept once, deliberately.**

It's gone from all six. It survives in one documented place inside the formatter, guarding the elapsed-time units only. Reason it isn't deleted outright: Foundry's `NumberField.min` *clamps* rather than rejects, so `{type: 'minute', quantity: -5}` becomes `0` and reaches the formatter. Without the fallback that renders "0 Minutes". So I'd push back on "the blanket `min: 0` stops being load-bearing for the timed types" — removing the fallback is what would make `min: 0` load-bearing, by letting a clamped zero through to the label.

**1. Branch casting-time labels by unit — not taken, and I'd like to make the case.**

I enumerated every input across all six sites. The split changes exactly one output: `{type: 'minute'|'hour', quantity: 0}`, which renders "1 Minute" today.

| type | qty | today | after the split |
|---|---|---|---|
| action | 0 | Free | Free |
| action | 1 / 2+ | 1 Action / N Actions | unchanged |
| minute, hour | 1+ | N Minutes | unchanged |
| **minute, hour** | **0** | **1 Minute** | "0 Minutes" if `\|\| 1` goes, per the instruction |

And that state isn't reachable: the config tab already draws the unit line at the input, `ItemActivationCoreConfigTab.svelte:50`, `min={activationCostType === 'action' ? 0 : 1}` — that's the one line this PR added to that file. Nothing in `packs/` authors it either (0 of 1032 authored activation costs, counting items embedded in monster actors).

On `resolveItemActionCost` as precedent: I read that as a different contract rather than the same split. `type !== 'action' -> 0` means *a spell that takes an hour costs zero action pips*, which is right for the action economy but says nothing about how to label a casting time.

Happy to be overruled — if you want the branches split for readability regardless, it's now a one-place change inside the formatter rather than six.

**5. `activateItemMacro` — done.** Now `resolveItemActionCost(item)` with an `> 0` guard, matching `NimbleCharacter.activateItem`.

**4. monsterFeature activation cost editor — you're right, and my earlier reply was wrong.**

I said on the `resolveItemActionCost.ts` thread that `MonsterFeatureSheet.svelte` has an Activation tab with the cost editor. The tab exists, but it routes `MonsterFeatureSheet.svelte:61` -> `ItemActivationConfigTab` -> `ItemActivationCoreConfigTab`, and the `{#if document.type !== 'monsterFeature'}` at line 39 wraps the entire Activation Cost section. So it is hidden, exactly as you said, and a free monster feature is only authorable via pack JSON. It's pre-existing (this PR touched only line 50 of that file), so I've left it for the later PR in the #582 series rather than widening this one. Sorry for the bad steer.

---

Two things came out of the discussion that weren't in your list, both UI-only:

- **Reactions rendered `undefined`.** Four sheets read `activationCostTypes['reaction']`, but that map has no `reaction` key, so anything authored against the legacy `type: 'reaction'` shape printed the literal string. Both shapes now resolve: `isReaction` on an action cost, and the legacy type. They read "Reaction" / "Reaction (2 Actions)" / "Free Reaction" — never "2 Reactions", since one reaction costs N actions rather than being a resource you spend several of (CoreRules-2:438). `CustomReactionsPanel` opts out and keeps the bare cost, since every row there is already a reaction. Related: the spells tab showed its reaction-trigger tooltip only for the legacy type, so a reaction authored the current way had no trigger on hover despite the config tab offering the field.
- **`turn` left the cost dropdown.** "1/turn" is how often an ability may be used, not what it costs (CoreRules-2:500 "1/Turn Abilities", :818, GMguide-2:1662 `ACTION: (1/turn)`). It was selectable but rendered no label at all, and nothing authors it. Frequency stays unmodelled; it wants its own field beside the cost, not inside it.

Also worth a look while reviewing: casting times in the PDF export are now capitalized ("1 Action AOE" rather than "1 action AOE") as a side effect of sharing the formatter. It matches the "Free" / "Range 1-5" / "1 Mana" already around them, but say the word if you'd rather keep the lowercase prose style.

The label changes are unit-tested but not yet live-verified in Foundry — flagged in the PR body with the three spots worth eyeballing.

trevlar (approved):
**✅ Approved with one fix — the reaction predicate is duplicated in two places that now disagree**

Third round and everything from the last two landed. `formatActivationCostLabel()` is the shared formatter, the zero case is confined to the action unit, the `|| 1` falsy-zero fallback is gone from all six call sites, and the reaction shapes that used to render the literal `undefined` read correctly now. You also declined to drop the shared `min: 0` and explained why in the docblock — one schema field, `NumberField.min` clamps rather than rejects. That was the right call.

One real thing:

- **src/view/sheets/pages/PlayerCharacterSpellsTab.svelte:59** — `carriesTrigger = isReaction || activationType === 'reaction'` is a second copy of the formatter's `isReactionCost()`, minus its `type === 'action'` requirement. The "Is Reaction" checkbox at `ItemActivationCoreConfigTab.svelte:77` isn't gated on cost type and the trigger field shows whenever it's ticked (line 88), so `{type: 'minute', quantity: 10, isReaction: true}` is authorable — and it renders "10 Minutes" (your formatter has an explicit test for ignoring the flag there) with a "Reaction Trigger: …" tooltip attached. Export the predicate from `formatActivationCostLabel.ts` and use it in both spots so the label and the tooltip can't drift apart.

Optional, take or leave:

- **PlayerCharacterSpellsTab.svelte:44** — `getSpellMetadata` is the only new logic not covered by the formatter's test, and it's in a `<script>` block past the 30-line threshold in STYLE_GUIDE:403. A plain `.ts` extraction (`:409` — no runes needed) would make the four tooltip branches testable.
- **PlayerCharacterSpellsTab.svelte:70** — `data-tooltip="${tooltip}"` puts unescaped item text into a string that reaches `{@html meta}` at line 333. Pre-existing, but you're rewriting the block.
- **combat.svelte.test.ts:3749** — the two fixture `cost` additions from 8559793f went stale once `resolveMinionAttackActionCost` restored the legacy default; reverting them puts the costless monster-feature shape back in the integration path.
- **formatActivationCostLabel.ts:70** — the `special` fallback ended up hand-copied at five call sites; returning the bare type label for it would collapse all five.
- **ItemActivationCoreConfigTab.svelte:39** — still no cost editor for `monsterFeature`, so a free monster feature isn't authorable from the sheet. Fine to leave for later in the #582 series.

Needs a restack onto current `dev` before merge — #886 touched the same two lines in `consumeCombatantAction`, so keep its `isActionTrackingAutomationEnabled()` gate above your `normalizedCost === 0` early return when you resolve it.

PR #886, opened 2026-07-31
feat(settings): automation settings group with per-family toggles

Fronix:
## Summary

Replaces the single `automation.autoApplyConditions` world setting (which despite its name gated the entire rule-event dispatcher and defaulted to off) with an Automation settings group: 8 world-scope family toggles, all defaulting to **on**, managed from a GM-only **Configure Automation** dialog in the system settings tab. GMs can now pick and choose what the system automates, and fresh worlds get working automation out of the box.

image

## Changes

- **New settings module** (`src/settings/automationSettings.ts` + `registerAutomationSettings.ts`): 8 `automation.*` world keys (`applyRuleEffects`, `derivedConditions`, `resourceRecovery`, `resourceSpending`, `actionTracking`, `healthStateSync`, `combatConvenience`, `chatNotifications`), fail-open getters, and a read-only legacy migration: a world that explicitly stored the old toggle keeps that behavior as the `applyRuleEffects` default; the legacy key stays registered hidden.
- **Dispatcher refactor**: the 10 per-handler `isAutoApplyEnabled()` guards are replaced by a single gate inside `dispatch()`. New `static alwaysDispatchedEvents` on rule classes exempts lifecycle methods that are core plumbing with no manual fallback: `diceConsumer` and `toggleEffect` keep receiving `onItemActivated` even with rule automation off, so dice spend prompts and toggle activation (e.g. Rage) can never be bricked. Activation cards are only suppressed when the replacement flow is guaranteed to run. `src/utils/isAutoApplyEnabled.ts` is deleted.
- **Per-family guards** at minimal chokepoints: derived conditions (AE pre-create/post-delete), charge/dice pool recovery (one shared eligibility predicate each), mana + charge spending, combat action consumption, bloodied sync (Last Stand stays always-on), auto-initiative on combat start, and chat notifications (initiative reminders, pool gain notices, toggle effect announcements).
- **New dialog** (`src/view/settings/AutomationSettingsDialog.svelte`): four sections (Rules & Effects, Resources, Combat, Chat), seeds from stored values, writes only changed keys, applies live with no reload. Registered as a restricted `registerMenu` submenu.
- **Docs**: GM settings guide gets an Automation section (always-on behaviors, mid-combat-flip caveat, migration note); conditions page and `docs/system/rules.md` updated to the new semantics.
- **Localization**: 8 new `NIMBLE.settings.<key>.{name,hint}` blocks + `automationMenu` chrome block in `en.json`; legacy strings reworded to "(Legacy)". Flagged for human review per AGENTS.md.

## Test Plan

- `pnpm check` (175 test files / 2912 tests, 14 new covering the gate exemptions, fail-open getters, legacy seeding truth table, and one off-state case per family) and `pnpm build:system` pass at each of the three commits.
- Reviewer smoke test in Foundry (v13):
  1. Open **Game Settings → Configure Settings → Nimble** as GM: a **Configure Automation** button appears (players must not see it). All 8 toggles default on.
  2. In a world that had the old "Auto-Apply Conditions from Rules" stored as `false`: Apply Conditions and Effects from Rules starts off; the other 7 start on.
  3. Turn **Apply Conditions and Effects from Rules off**, then activate Rage and Bloodlust on a Berserker: Rage still toggles on and the Fury Die spend panel still opens (exempt core plumbing), while rule-driven condition application stops.
  4. Turn **Health-State Sync off**, drop a character below half HP: no Bloodied. Drop a Last Stand monster to 0 HP: Last Stand still fires.
  5. Turn **Resource Spending off**, cast a tier-1 spell: mana untouched. **Action Tracking off**, use an item in combat: action pips untouched. **Combat Convenience off**, begin combat: no auto-rolled character initiative.
  6. Toggles apply immediately, no reload.
- All of the above was also verified live via scripted real-UI Playwright runs against a v13 world.

## Checklist


trevlar (changes requested):
**🔧 Push back — the design is right, but three of the eight family gates and the entire legacy migration probe ship untested**

The dispatcher refactor is genuinely better than what it replaces: collapsing ten scattered `isAutoApplyEnabled()` guards into one gate inside `dispatch()`, with `alwaysDispatchedEvents` as the explicit escape hatch for core plumbing, is the right shape. I traced every handler you unguarded (`handleDamageApplied`, `handleCombatTurn`, `handleActorUpdate`, `handleSaveResolved`, `handleRest`, `handleInitiativeRolled`, `handleUseItem`, `handleUpdateCombat`, `handleDeleteCombat`, `handleConditionApplied`) — all of them are pure context-building around a `dispatch()` call, so moving the gate inward is behavior-preserving. Good.

What I can't sign off on is the coverage story. The PR body says "one off-state case per family"; that's three families short, and the one piece of logic that can silently change an existing world's behavior has no test at all.

**Must fix before merge**

- **src/settings/registerAutomationSettings.ts:51** — 
- **src/documents/combat/combat.svelte.ts:661** — the combat-convenience gate has no off-state test, and the on-state one already exists next door at `combat.svelte.test.ts:1624` ("auto-rolls unrolled character initiative and resets non-character actions at combat start"). Add the mirror case: settings stubbed to `false`, assert `combat.rollInitiative` is not called while `#applyNpcActionResetUpdates` still runs.
- **src/utils/combatTurnActions.ts:386** — the action-tracking gate has no test, and `combatTurnActions.test.ts` is already there to hang it on. Assert `combat.updateEmbeddedDocuments` is never called when the toggle is off.
- **src/documents/item/spell.ts:68** — the mana-deduction gate has no test and no test file for `spell.ts` at all. This is the one gate that touches persisted actor resources, so it's the one you least want to regress. A focused test asserting `actor.update` isn't called for a tier-1 spell with spending off would cover it.

- **src/view/settings/AutomationSettingsDialog.svelte:48** — `save()` awaits `setAutomationToggle` in a bare loop. If `game.settings.set` rejects (permissions, socket hiccup), the GM gets no confirmation, no error, a dialog that stays open, and an unhandled rejection out of the click handler — with some toggles already written and others not. Wrap the loop in try/catch and surface `ui.notifications.error` on failure.
- **src/utils/combatTurnActions.ts:385** — the comment says "report the untouched pool so callers proceed normally", but both call sites (`character.ts:1858`, `minionGroupTokenActions.ts:2267`) discard the return value, and a combatant genuinely at 0 actions returns `0` here — the same value the out-of-actions path returns, so the two states are indistinguishable to any future caller. Either drop the justification or return something callers can actually tell apart.
- **docs/documentation/gm/settings.md:16** — the doc says Resource Spending off means "resource costs are not enforced", but `src/utils/spell/applyUpcastDeltas.ts:85` still rejects upcasting with "Insufficient mana" regardless of the toggle. Small wording fix, or gate the upcast check too.

**What's good**

- `AutomationSettingsMenu` mirrors the existing `CustomSpellSchoolsMenu` submenu pattern exactly (same `GenericDialog` wrapper, same `openEditor` focus guard, same `restricted: true` registration) — easy to follow, nothing invented.
- Reading every toggle live at the call site instead of caching means "applies immediately, no reload" falls out for free, and the fail-open `readAutomationToggle` means an unregistered key can't brick a world.

---

1. Test the legacy storage probe, both shapes
2. Add off-state test for combat-convenience initiative gate
3. Add off-state test for action-tracking gate
4. Add off-state test for mana-spending gate
5. Handle `setAutomationToggle` rejection in the dialog
6. Clarify or fix `consumeCombatantAction`'s early-return value
7. Reconcile upcast mana check with the docs

Fronix:
All 7 items are addressed in e4b8ba50.

**Must-fix**

1. **Legacy storage probe** — new `src/settings/registerAutomationSettings.test.ts` covers both shapes and the edges: stored `false` seeds `applyRuleEffects` default `false`, stored `true` and fresh-world (no stored doc) both seed `true`, a missing world storage collection doesn't throw (defaults `true`), and all 8 toggles + the submenu register under the system namespace.
2. **Combat-convenience gate** — added the mirror off-state test next to the on-state one in `combat.svelte.test.ts`: with the toggle stubbed `false`, `rollInitiative` is never called while the non-character action resets still run.
3. **Action-tracking gate** — `combatTurnActions.test.ts` now stubs the setting explicitly (no more riding on the fail-open default) and asserts `updateEmbeddedDocuments` is never called and the pool comes back untouched.
4. **Mana gate** — new `src/documents/item/spell.test.ts` tests through the real `activate()`: off-state performs no `actor.update` while the activation card still fires (so the test isn't vacuous), plus on-state deduction, clamp-at-zero, and cantrips-free cases.

**Non-blocking**

5. Dialog `save()` now wraps the write loop in try/catch: on failure it logs, surfaces `ui.notifications.error`, and keeps the dialog open. Each successful write also updates the snapshot, so retrying Save only rewrites the toggles that didn't land.
6. Went with dropping the justification: the comment now states outright that the gated return value is indistinguishable from the blocked-use `0` and callers must not branch on it. Both call sites discard the value today, so changing the signature felt speculative — happy to switch to a distinguishable return if you'd prefer.
7. Doc wording fixed rather than gating the upcast check: sheet mana is still a manually-tracked number with spending off, so capping upcast against it stays coherent. The doc now says costs are neither deducted nor checked when off, with the upcast cap called out as the one exception.

trevlar (changes requested):
**🔧 Push back — round-1 feedback is all addressed; one coverage gap left, on the exemption mechanism itself**

Re-reviewing after `e4b8ba5`. Every item from the last pass landed: `registerAutomationSettings.test.ts` covers the legacy storage probe across all four shapes including the missing-collection case, the combat-convenience / action-tracking / mana-spending off-states all have tests now, `save()` wraps the write loop in try/catch with `ui.notifications.error`, `consumeCombatantAction`'s early return carries an honest comment instead of a wrong justification, and the upcast-cap exception is called out in the GM docs. Good turnaround.

**Must fix before merge**
- **src/models/rules/diceConsumer.ts:94** (and **src/models/rules/toggleEffect.ts:175**) — the two `static override alwaysDispatchedEvents` declarations are the only thing standing between "rule automation off" and a bricked Fury Die spend panel, and nothing tests them. Both new tests define their own synthetic class with the literal inline (`ruleEventDispatch.test.ts:114`, `shouldSuppressActivationCard.test.ts:28`), so they prove the *mechanism* works while proving nothing about the two classes that actually use it. A typo or a future rename of `onItemActivated` compiles, passes CI, and silently reintroduces exactly the failure this was built to prevent. Assert it directly — `expect(DiceConsumerRule.alwaysDispatchedEvents).toContain('onItemActivated')` and the same for `ToggleEffectRule` — or drive a real `DiceConsumerRule` instance through the dispatcher with the toggle off.

- **src/models/rules/base.ts:169** — `readonly string[]` accepts anything. `readonly (keyof NimbleBaseRule)[]` turns a misspelled or renamed lifecycle method into a compile error, and the dispatcher already narrows `methodName` to `keyof NimbleBaseRule`, so the two types line up. That would also make the must-fix above much less load-bearing.
- **src/models/rules/toggleEffect.ts:405** (and **:537**, **src/documents/combat/handlers/initiativeMessageHandler.ts:30**) — Chat Notifications promises three behaviors in its hint; only the pool-gain notice has an off-state test. `toggleEffect.test.ts` already exists, so a case there is cheap.

**What's good**
- Per-family guards sit at real chokepoints — `isEligibleForAutomaticRecovery` / `isEligibleForAutomaticRefill` each collapse three call sites — rather than being sprinkled at every caller.
- The always-on carve-outs are principled and written down: Last Stand staying outside the health-state gate, and the mid-combat-flip caveat in the GM docs, are the details that would have bitten someone later.

---

1. Test the real rule classes' `alwaysDispatchedEvents` values
2. Type `alwaysDispatchedEvents` as `keyof NimbleBaseRule`
3. Cover the toggle-effect and initiative chat-notification gates

Fronix:
Thanks @trevlar! All three items are addressed in 3f13f58d (`pnpm check` green: 177 files / 2931 tests, 8 new).

**1. Real classes' `alwaysDispatchedEvents` tested directly (must-fix)** — `diceConsumer.test.ts` and `toggleEffect.test.ts` now assert `DiceConsumerRule.alwaysDispatchedEvents` / `ToggleEffectRule.alwaysDispatchedEvents` contain `'onItemActivated'`, so a typo or rename in the actual classes fails CI rather than silently re-gating the exempt dispatches.

**2. `alwaysDispatchedEvents` typed as `keyof NimbleBaseRule`** — done in `base.ts`, both real overrides, the dispatcher's cast (which also let us drop the `methodName as string` cast), and the synthetic test class in `shouldSuppressActivationCard.test.ts`. One wrinkle: statics inside the generic class body can't reference the bare class name, so `base.ts` uses a small unexported `RuleLifecycleEvent = keyof NimbleBaseRule` alias. As you noted, this makes item 1 belt-and-suspenders — a misspelled event is now a compile error before the tests even run.

**3. Remaining chat-notification gates covered** — new off-state tests for the toggle-effect start/end announcements in `toggleEffect.test.ts` (toggle still flips, no announcement posts, with fail-open control tests so the off-state assertions aren't vacuous) and for the initiative bonus reminder via `initiativeMessageHandler` in `initiativeMessage.test.ts` (no reminder with the toggle off, posts by default). Each uses a scoped `game.settings.get` mock returning false only for the `chatNotifications` key, restored afterward so the fail-open tests stay undisturbed.

PR #887, opened 2026-07-31
feat(rules): offer dice-consumer spends on the attack card

Fronix:
## Summary

Lets a `diceConsumer` be spent from the attack card it modifies instead of the sheet's dice pool panel, for spends the rules tie to how an attack landed. Berserker's Death Blow adopts it and becomes pure pack data.

image

image

## Changes

- **New `cardOffer` field on `diceConsumer`** (`hit` / `criticalHit`). When set, the owner gets a button on their own attack cards whose outcome matches, and the spend folds into that card's damage total.
- **Folding, not a second roll.** Applying the bonus as a separate hit would re-run the target's armor, resistances and flat damage reduction against the same attack. The bonus is appended as a flavored numeric term, matching how banked pool faces already ship, so `getDiceDamageTotal` counts it as dice.
- **Attacker-side offers on the existing `incomingReactions` framework**, whose schema already advertised `kind` as an extension point. Previously every entry was defender-side.
- **Execution routed through the primary GM**, like the other reaction kinds. Not a convention: a chat message is updatable only by its author or a GM (owning the acting actor implies neither), and one writer is what stops a concurrent reaction from clobbering the entry array.
- **Stale-pick rejection.** The picker reports the faces it displayed, so a pool that moved mid-dialog is refused rather than spending whichever dice now sit at those indices.
- **A consumer that opts in loses its sheet flow entirely** - the panel can't see the attack's outcome, so it would offer a crit-only spend on any activation.
- **Death Blow** sets `cardOffer: 'criticalHit'`, with `Migration034` backfilling existing copies and dropping the UI note from its description.
- Docs: `cardOffer` guidance and a worked example in the `diceConsumer` partial, plus a player-facing account in Dice Rolls & Chat Cards. The reference field table generates itself from the schema.
- Drive-by: the reaction socket channel was hardcoded `system.nimble`, which silently breaks under the `nimble-dev` rolling build. Now derived from `SYSTEM_ID`.

## Test Plan

**In Foundry**, as a Berserker with Rage active and at least one Fury Die:

1. Attack a target and land a **critical hit**. An `Add Damage: Death Blow` button appears on the card, visible to the Berserker's owner and the GM.
2. Click it, pick some Fury Dice, confirm. The card's damage total rises by double the picked sum, the dice leave the pool, and the card records what was spent.
3. Click **Apply Damage** once. The target should take the folded total with armor and any damage reduction applied a single time.
4. Land an **ordinary hit**: no button (Death Blow is crit-only).
5. Activate Death Blow from the sheet: it no longer opens the pool panel, and it is absent from the panel's feature list.
6. As a **player** (not the GM), repeat step 2 to exercise the GM socket relay.

**Automated:** `pnpm check` (2902 tests). New suites cover the fold, the offer collector, the outcome gate, and the executor end to end against a real character-actor fixture.

**Known limits**, both documented rather than fixed:
- The bonus takes the attack's damage type, so a feature dealing a specific type (radiant, fire) can't use this yet.
- No melee/ranged filter on the offer.

## Checklist


src/utils/dicePool/dicePoolConsumers.ts
```diff
@@ -138,6 +139,14 @@ function getDicePoolConsumers(
 			if (consumer.mode !== 'manual') continue;
 			if (normalizeIdentifier(consumer.poolIdentifier) !== poolIdentifier) continue;
 			if ((consumer.poolScope ?? 'item') !== pool.scope) continue;
+			// A consumer that opted into a chat-card offer is spent from the card
+			// the spend modifies, not from the sheet panel, where it would be
+			// ungated by the attack's outcome. The card executor asks for them
+			// explicitly.
```
trevlar, line 156: You added two gates here, and `dicePoolConsumers.test.ts` is untouched by this PR. The `includeCardOffers` filter is only covered indirectly, and the new `if (typeof rule.appliesTo === 'function' && !rule.appliesTo()) continue;` on line 149 changes shipped behavior for **every** `diceConsumer`, not just card offers — a consumer whose predicate fails used to appear in the sheet panel and now doesn't. That's a good change, but it needs its own tests: a `cardOffer` consumer hidden by default and returned with `{ includeCardOffers: true }`, and a predicate-failing consumer dropped from the panel list.

src/view/dialogs/PoolSpendOfferDialog.svelte
```diff
@@ -0,0 +1,235 @@
...
+	});
+
+	let pool = $derived.by(() => {
+		subscribePool();
+		return getDicePools(actor).find((p) => p.id === poolId) ?? null;
+	});
+	let consumer = $derived(
+		pool
+			? (getDicePoolConsumers(actor, pool, { includeCardOffers: true }).find(
+					(c) => c.ruleId === ruleId,
```
trevlar: The consumer lookup matches on `ruleId` alone. Both of the other two lookups disambiguate by item as well (`IncomingReactionPrompts.svelte:162`, `chatMessage.ts:1316`) precisely because, as your own comment says, rule ids are only unique within an item. Pass the item ID through and match it here too, or the preview can come from a different consumer than the one the GM will execute.

src/view/chat/components/IncomingReactionPrompts.svelte
```diff
@@ -126,6 +214,9 @@
 				messageId: messageDocument.id ?? '',
 				entryId: entry.id,
 			});
+		} catch (error) {
+			console.error('Nimble | Failed to use a chat card reaction', error);
+			ui.notifications?.error(localize('NIMBLE.chat.incomingReactions.useFailed'));
```
trevlar, line 230: For a player, `requestIncomingAttackReaction` returns as soon as the socket is emitted, so `busy` clears while `used: true` is still in flight from the GM. Re-opening the offer and confirming inside that window folds the bonus a second time (the GM's `used` check hasn't been written yet, and both requests compute `remainingFaces` from the same starting pool — so the dice come out once and the damage goes on twice). Keeping `busy` set for the spend path until the entry reads `used`, or optimistically hiding the entry after submit, closes it.

docs/system/rules.md
```diff
@@ -428,6 +428,14 @@ Multiple `diceConsumer` rules can target the same pool — e.g. an `autoBonus` c
 - **Refill predicates**: each `dicePool.refills` entry accepts an optional `predicate`, evaluated against the actor's live domain when the trigger fires (e.g. `{ "self": "raging" }` to gate a turn-start refill on an active toggle). Entries without a predicate always apply.
 - **`modifyPool.addRefills`**: a modifier can contribute refill entries to the target pool without editing the base pool rule — the granting feature carries its own trigger. The modifier's rule-level predicate gates whether the entries are included at all; entry-level predicates are evaluated at trigger time (prefer these for state that flips mid-combat, since rule-level gating churns the stored pool definition).
 - **`modifyPool.minFace`**: a minimum face value for dice rolled into the target pool. Rolls below the floor are raised to it at every roll point (refills, activation rolls, initial seeding); manual face edits stay unclamped. The highest floor among contributing modifiers wins.
+#### Card-side pool spends
```
trevlar: Add a new line to separate the header from the list.

trevlar (changes requested):
**🔧 Push back — the migration rewrites descriptions too broadly, and the two new gates in `getDicePoolConsumers` ship untested**

The feature design is right: folding the bonus into the existing `DamageRoll` instead of posting a second packet, routing every kind through the primary GM, and reporting `expectedFaces` so a moved pool is refused rather than silently respent. The docs and the executor tests are genuinely good. Two things need to change before merge.

**What's good**

- The folding rationale is documented where it matters (`foldBonusIntoPrimaryDamage.ts:42-52` is honest that "counts as dice" is a reading, not a settled rule), and `chatMessageSpendOffer.test.ts` drives a real character actor through the real pool helpers rather than mocking past the interesting parts.
- Deriving the socket channel from `SYSTEM_ID` is a good catch — that would have broken silently on `nimble-dev`.

---

1. Guard `source.system` before writing the description.
2. Test the `includeCardOffers` and `appliesTo` gates in `dicePoolConsumers`.
3. Match the consumer by item id in the picker dialog.
4. Close the double-submit window while the GM write is in flight.
5. Move the new docs heading out of the bullet list.

Fronix:
All five addressed in `b72db96e..5f31248b`.

**1. Migration rewrote descriptions too broadly** (`b72db96e`)

You were right on both halves, and the name match was the worse of the two: any homebrew feature called "Death Blow" would have had its text replaced. It now requires our own `death-blow-fury-consumer` rule id *in addition* to the name-or-source-id match, guards `source.system` before touching it, and replaces only the exact description shipped before the card offer existed. A description a GM has since rewritten is theirs to keep.

**2. `includeCardOffers` and `appliesTo` gates untested** (`b981497e`)

Agreed, and the `appliesTo` point is the one that needed saying: that gate changes shipped behaviour for every `diceConsumer`, not just card offers. Five tests added: a `cardOffer` consumer hidden by default, returned with `{ includeCardOffers: true }`, and a sheet-panel consumer on the same pool left alone; plus a predicate-failing consumer dropped and a predicate-passing one kept.

**3. Picker matched the consumer by rule id alone** (`677d9c2f`)

Correct, and inconsistent with the other two lookups for no reason. `findOfferPoolId` is now `findOfferTarget` and returns the item id it had already resolved; the dialog matches on rule id and item id, same as the card button and the executor.

**4. Double-submit window** (`677d9c2f`)

The analysis holds: the request returns at socket-emit, a second confirm inside that window passes the unused-entry check, both reads see the same unspent pool, and the bonus folds twice for one set of dice.

I fixed the correctness half in the executor rather than only in the card. `resolveSpendPoolForDamageOffer` now takes an in-memory lock keyed by message and entry for the length of the resolve, which is sufficient because every request lands on the one primary GM's client. Client-side hiding alone would have left the same race open to a second owner of the actor.

The card does hold the offer back too, as you suggested, since a silently dropped confirm is a poor experience. The hold lapses after a bounded wait, so a spend the GM refuses as stale becomes usable again instead of disappearing (the refusal path writes nothing, so there is no update to release on).

**5. Docs heading** (`5f31248b`)

Blank line added.

One gap worth naming: there is no migration test harness in the repo, so item 1 ships without a test. `pnpm check` is green at 2907 tests across 176 files, clean typecheck, no dependency violations.

trevlar (changes requested):
**🔧 Push back — the outcome gate only runs at card creation, so an interactive `forceReroll` on the same card can both invalidate and erase a spend**

The five items from the last pass are properly addressed, and the design is right: folding into the existing `DamageRoll` instead of posting a second packet, routing every kind through the primary GM, `expectedFaces` for stale-pick rejection, the client hold plus GM-side `inFlightSpendOffers`. What's not handled is the interaction with the *other* offer that can already be sitting on the same card.

**Must fix before merge**

- **`src/documents/chatMessage.ts:1313`** — `#applySpendPoolForDamageOffer` never re-checks `entry.outcomeTrigger`. The gate runs once, in `offerSurvives`, at card creation. Concrete break: a Berserker crits a target defended by a Cheat with Pocket Sand (`pocket-sand.json` is an *interactive* `forceReroll` with `rerollTrigger: "always"`, so its button is on every attack card). The Cheat rerolls, the new roll is an ordinary hit, and `resolveForceRerollReaction` duly writes `isCritical: false` — but the crit-only "Add Damage: Death Blow" button is still there and still spends. Re-check the trigger at execution time against the card's current `system.isCritical` / `system.isMiss`; the reroll resolver already keeps both current.

- **`src/documents/chatMessage.ts:1237`** — the reverse order silently eats the spend. If the Berserker folds first and the Cheat rerolls after, the reroll rebuilds from `serialized.originalFormula`, which carries no folded term: the bonus vanishes from the total, the Fury Dice are already gone from the pool, and the entry keeps `used: true` with a `usedAmount` attribution line still claiming damage that is no longer in the roll. You have what you need to fix it — after building `newRollJson`, re-apply `foldBonusIntoPrimaryDamage` for each used `spendPoolForDamage` entry's `usedAmount`, or refund from `usedFaces` and reset the entry.

- **`src/view/chat/components/IncomingReactionPrompts.svelte:176`** — the new UI logic ships untested, in a file that already has `IncomingReactionPrompts.test.ts` and a harness. Uncovered: the `spendPoolForDamage` heading/label/attribution branches, `findOfferTarget`'s pool-and-item resolution, and the `heldEntryIds` gate that is the client half of the double-spend guard. Everything else in this PR is tested unusually well — this is the one gap.

- **`src/documents/chatMessage.ts:1350`** — `poolSpendStale` and the silent `return`s around it fire inside a `game.user?.isGM` guard, so the rejection lands on the GM's screen while the player who made the pick just sees the button vanish for `SPEND_HOLD_DURATION` and come back unexplained. Relaying the failure to `requestingUserId` would let the hold clear immediately instead of waiting out the 10s timer.

- **`src/documents/chatMessage.ts:1326`** — `entries` is read here and written back at :1388, several awaits later. `inFlightSpendOffers` is keyed per `messageId:entryId`, so two *different* entries on one card resolving inside that window each write their own stale copy of the array and one `used` flag is lost. Pre-existing shape (`resolveForceRerollReaction` does the same at :1225/:1267), but the doc comment credits single-writer GM routing with preventing exactly this, and it doesn't quite.

- **`docs/system/rules.md:432`** — placement is still off: `modifyConsumer`, `poolGainMessage`, the refill-trigger bullet, `maximizeDie`, and the paragraphs through :445 all belong to "Pool modifiers and refill gating" but now render under `#### Card-side pool spends`. Moving the new `####` section below them fixes it.

- **`src/view/sheets/components/AttackActionPanel.svelte:182`** — dropping `max-height: 300px` is unrelated to this feature and isn't in the description; with no max-height the sibling `overflow-y: auto` has nothing to clamp against. Worth a line in the body either way so it doesn't surprise anyone in the changelog.

- **`src/view/dialogs/PoolSpendOfferDialog.svelte:11`** — inline `interface Props`. STYLE_GUIDE "Props Definition" wants these in `types/components/*.d.ts`; new file, so worth starting it in the right place.

**What's good**

- `chatMessageSpendOffer.test.ts` drives a real character fixture through the real pool helpers and covers the GM-only gate, the relayed-GM-identity spoof, the non-owner request, the stale pick, and the crit/hit/miss gate against the *reroll* result rather than the discarded roll.
- Charging the pool only after the fold is known to be possible, and storing the used-spend attribution as components rather than a pre-localized sentence — both small, both easy to get backwards, both right.

---

1. Re-check `outcomeTrigger` at spend execution, not just at stamping
2. Re-fold or refund a used spend when a reroll rebuilds the roll
3. Add component tests for the new spend-offer UI
4. Report the stale-pick rejection to the spending player
5. Note the shared stale-`entries` write-back window
6. Move the new docs heading below the bullets it split
7. Split out or justify the `max-height` removal
8. Move `PoolSpendOfferDialog` props into `types/components/`

trevlar (approved):
**✅ Approved**

Every must-fix from the last pass is genuinely closed, not papered over: the outcome gate is now re-derived from the live rule at execution (`chatMessage.ts:1544`), `#carryFoldedSpendsAcrossReroll` re-folds or refunds a spend a reroll invalidates, `queueReactionWrite` serializes the whole `incomingReactions` read-modify-write per card rather than per entry, refusals relay back to the spending player, and the spend-offer UI has real component tests.

- `#carryFoldedSpendsAcrossReroll` picks the right ordering of bad options — re-fold what the new outcome still earns, refund what it doesn't, and prefer an unearned bonus over silently charging the player when the refund can't land.
- Test coverage tracks the risk: stale picks, spoofed GM identity, non-owner requests, crit/hit/miss gate flips against the reroll result, banked-reduction and maximize rejections, and the concurrent-resolve serialization.

Still not in the PR body: the `max-height: 300px` removal in `AttackActionPanel.svelte:182`. Unrelated to this feature and worth one line in the description so it doesn't surprise anyone reading the changelog.

Issue #685, opened 2026-04-08
[EPIC] Commander automation


Fronix:
All Commander class automation — Combat Dice pool, Coordinated Strike!, Commander's Orders, Combat Tactics, and subclass features (Bulwark, Vanguard, Spellblade).

Parent: #365

Fronix:
> *This was generated by AI during triage.*

## Stale blocker sweep

5 checklists under this epic cite a closed foundation as a blocker. **A closed foundation does not mean the feature is unblocked.** #943 is the reference for re-reading those lines.

| Checklist | Cites (all closed) | Needs care |
|---|---|---|
| #615 | #576, #579 | #576 |
| #617 | #575, #576, #581, #584 | #576, #581, #584 |
| #618 | #574, #575, #576 | #576 |
| #619 | #576, #578, #579 | #576 |
| #620 | #584 | #584 |

"Needs care" flags the citations that shipped under a different name or a narrower scope than the checklist assumes:

- **#576 `resourceCounter`** never existed under that name. It is `chargePool` + `chargeConsumer`. Usually genuinely unblocked.
- **#581 `grantMovement`** grants a persistent movement **mode**, not a one-off "move N spaces for free". Free-move features that cite it are still blocked, now by #935.
- **#582** was rescoped before closing. Raw action grants and cost overrides are served; self-directed activation grants and free Move are not.
- **#584 `triggerOnEvent`** was never built. Each rule type has its **own** trigger vocabulary and they differ. #943 carries the table; this citation cannot be resolved generically.

### Class specifics

- **Reposition!** and **Advance!** grant free movement to *other actors*, which is **#935**. Both cite #581; that closure does not unblock them.
- **Inerrant Strike** rerolls a missed attack and modifies the new roll, which is **#936** buckets B and D combined.
- **Heavy Strike**'s size-scaled push is forced movement, **#935**.
- #620 cites only #584. Check the trigger table in #943 for the specific event before assuming it is unblocked.
- **Rigorous Training** is **#938**.

Berserker's three checklists (#600, #601, #602) have already been corrected in place and can be used as the worked example.

Issue #619, opened 2026-04-08
[Automation] Commander - Champion of the Vanguard subclass checklist


Fronix:
## Context
Checklist for all features of the **Champion of the Vanguard** Commander subclass (mobility / ally-action focus).

## Features

- [ ] **Advance!** (L3) - "(1/round) After you move toward an enemy, gain advantage on the first melee attack you make against it. When you use your Coordinated Strike, you and all allies within 12 spaces can first move up to half their speed for free." -> `resourceCounter` (1/round), `triggerOnEvent` (onMoveToward, onCoordinatedStrike), `conditionalBonus` (advantage on first melee), `grantMovement` (multi-ally) - blocked by #576, #579, #581, #584
- [ ] **Experienced Commander** (L7) - "Your Coordinated Strike may target 1 additional ally. Gain +1 use of Coordinated Strike/Safe Rest." -> Coordinated Strike target count +1, pool size +1 - modifies flagship Coordinated Strike issue
- [ ] **Survey the Battlefield** (L11) - "When you roll Initiative, regain 1 use of Coordinated Strike. +1 max Combat Dice." -> `triggerOnEvent` (onInitiative), Combat Dice pool cap +1 - blocked by #578, #584
- [ ] **As One!** (L15) - "Attacks made with your Coordinated Strike also grant advantage and ignore all disadvantage. Your chosen allies gain 1 additional action to use on their next turn." -> `conditionalBonus` (advantage + ignore disadvantage on CS attacks), `grantAction` (+1 action to all CS allies) - blocked by #579, #582

## Foundations This Checklist Depends On
#578, #579, #581, #576, #582, #584

Part of #365

Fronix:
Partially delivered by #895 and #896 (merged).

**Experienced Commander is ticked.** The +1 use of Coordinated Strike per Safe Rest landed as a cross-item `modifyPool` on the subclass feature. The other clause, "may target 1 additional ally", needs nothing built: `grantActivation` offers the attack to every token the user currently has targeted, and the base count of 1 is not enforced either, since who to target is left to the table. There is no remaining work behind this box.

**Survey the Battlefield** is not ticked. The "+1 max Combat Dice" half landed as a `modifyPool` on the `combat-dice` pool. The "regain 1 use of Coordinated Strike when you roll Initiative" half needs predicated charge-pool recoveries and is not built.

[...]

Issue #618, opened 2026-04-08
[Automation] Commander - Champion of the Bulwark subclass checklist


Fronix:
## Context
Checklist for all features of the **Champion of the Bulwark** Commander subclass (shield/armor focus).

## Features

- [x] **Armor Master** (L3) - WIRED (grantProficiency) - "You are proficient with plate armor."
- [ ] **Shield Expert** (L3) - "While wearing a shield, you may Defend 2 each round. The first time each round you block all of the damage from an attack, you may make an opportunity attack against the attacker for free." -> `conditionalBonus` (shield equipped), `resourceCounter` (2/round Defend), `triggerOnEvent` (onFullBlock), `grantAction` (free OA) - blocked by #576, #579, #582, #584
- [ ] **Juggernaut** (L7) - "When you use Coordinated Strike, you deal extra damage equal to your armor, and you can add 1 to your primary die." -> `damageBonus` (= armor value), primary-die bonus, hooked to Coordinated Strike - blocked by #574
- [ ] **Taunting Strike** (L11) - "(1/turn) You may Taunt a creature you hit until the end of their next turn." -> `resourceCounter` (1/turn), `applyCondition` (Taunted), `triggerOnEvent` (onHit) - blocked by #575, #576, #584
- [ ] **Shield Wall** (L15) - "Allies within 2 spaces gain ALL the benefits of the shield you have equipped." -> Aura effect sharing equipped shield's benefits to allies within 2 - needs aura/grantItem-to-others pattern

## Foundations This Checklist Depends On
#574, #575, #576, #579, #582, #584

Part of #365

Fronix:
Touched by #895 (merged). **Taunting Strike** is not ticked: only its use limit landed.

It now tracks 1/turn (`chargePool` with `max: 1` recovering on turn start, plus a `chargeConsumer`), and the card offers the Taunt as a condition button. Two things keep it short of done:

- There is no on-hit trigger, so the player activates it manually after the hit rather than it firing from the attack.
- The `taunted` condition itself is inert. Nothing in the system reads it, so applying it marks the target and changes no mechanics.

Worth knowing for whoever picks this up: #900 (open) fixes the condition button applying to the clicking user's selected tokens rather than the card's targets, which this feature hits directly, and gives applied conditions a source so the readout names who taunted whom.

Issue #617, opened 2026-04-08
[Automation] Commander - Commander's Orders checklist (6 tactical orders)


Fronix:
## Context
Checklist for the 6 selectable Commander's Orders. Each is an ally-targeted command with its own resource pattern. Coordinated Strike is flagship and NOT in this checklist.

**Already filed as dedicated issue and NOT in this checklist:**
- Coordinated Strike! - this batch

## Orders

- [ ] **Commanding Presence** (also listed as Combat Tactic) - See Combat Tactics checklist. Cross-listed here because rulebook lists it under Commander's Order heading as well.
- [ ] **Face Me!** - "Reaction (after an ally is crit within 12 spaces): Taunt that enemy until you drop to 0 HP." -> `triggerOnEvent` (onAllyCrit within range), `applyCondition` (Taunted, duration=self-at-0) - blocked by #575, #584
- [ ] **Hold the Line!** - "(1/encounter) Reaction (when an ally drops to 0 HP): Command them to continue the fight! Set their HP to 3x your LVL." -> `resourceCounter` (1/enc), `triggerOnEvent` (onAllyDropToZero), HP-set heal - blocked by #576, #584
- [ ] **I Can Do This ALL DAY!** - "(1/encounter) Reaction (when you would drop to 0 HP): You may expend any number of Hit Dice and set your HP to the sum rolled instead (do not add your STR)." -> `resourceCounter` (1/enc), `triggerOnEvent` (onWouldDropToZero), Hit Dice spend override - blocked by #576, #584
- [ ] **Move it! Move it!** - "When you roll Initiative you may give yourself and an ally advantage on the roll and +3 speed for 1 round." -> `triggerOnEvent` (onInitiative), `initiativeRollMode` (exists), `speedBonus` (exists, 1-round) - blocked by #584
- [ ] **Reposition!** - "Action/Reaction (on an ally's turn): Command 1 ally to move up to their speed (or 2 allies up to half their speed) for free." -> `grantMovement` on other actors, dual-mode action/reaction - blocked by #581, #582

## Foundations This Checklist Depends On
#575, #576, #581, #582, #584

Part of #365

Fronix:
Touched by #895 (merged). No boxes ticked, since in each case only the use limit landed and the effect itself is still manual. Recording it so the remaining work is clear and the file changes are not read as regressions.

[...]

PR #895, opened 2026-08-04
feat(commander): automate Coordinated Strike and repair the Commander packs

Fronix:
## Summary

Automates the Commander's Coordinated Strike! and the progression around it, using the
action-economy rule types from #582 (`actionCost`, `actionDelta`, `grantActivation`) for the
first time in pack data. Also repairs ten Commander pack files that were emitting spurious
damage packets on every chat card.

## Changes

- **Coordinated Strike!** is a zero-cost activation with two use limits (INT per Safe Rest,
  once per round), grants the targeted ally an immediate weapon attack via `grantActivation`,
  and pays for both attacks with `actionDelta` so neither side spends an action.
- **Level scaling**: Master Commander's +1 use at L9/13/17, Experienced Commander and Survey
  the Battlefield's pool bonuses, all as cross-item `modifyPool` rules.
- **Capstone riders**: As One! (L15 Vanguard), Unparalleled Tactics (L18-19) and Captain of
  Legions (L20) grant allies an action on their next turn.
  - **As One!** has no per-encounter limit in the rules ("Your chosen allies gain 1 additional
    action to use on their next turn", full stop), and ships without one.
  - **Unparalleled Tactics** and **Captain of Legions** do say "The first time each encounter",
    and that gate is **not in this PR**. It lands in #896, the branch stacked on this one, as a
    hidden once-per-encounter charge pool predicated on the two L18+ riders only. See the
    caveat in `23bc1ff` for why it could not be expressed at the time.
- **New `subclass:` domain tag** so a rule can be gated on the subclass a character picked,
  not just the class. Generic, five lines, and useful to every class.
- **Use tracking** on Hold the Line!, I Can Do This ALL DAY! and Taunting Strike.
- **Pack repairs**: ten files carried placeholder `acid`/`0` damage nodes that existed only to
  hang a reminder note off. Since #890 those emit a real zero-damage packet on every card.
  Move it! Move it! was worse, emitting `thunder` damage for what the rulebook describes as an
  initiative roll. I Can Do This ALL DAY! also carried a malformed `maxHitDice` rule that
  raised max Hit Dice for a feature about spending them. Master Commander's description listed
  level 5 among the levels granting +1 use of Coordinated Strike; Heroes-2:383 puts only the
  initiative regain and ignore-disadvantage clauses at 5, so the text implied a use the rules
  correctly do not grant.
- Migrations 036 to 038 carry all of it to existing worlds, each guarded on the exact
  superseded value so a hand-edited copy is never stomped.

Features that could not be expressed are recorded in the relevant commit bodies rather than
faked. The main ones: Weapon Mastery needs target-armor-tier predicates and a rule-driven
`ignoreArmor`; Reposition!, Advance! and Sweeping Strike need the v14 movement and multi-target
primitives; Master Commander's initiative regain needs predicated charge-pool recoveries.

Two roll-mode clauses attached to Coordinated Strike are also unimplemented, and are not called
out in their commit bodies:

- **Master Commander (L5)**: "Attacks made from your Coordinated Strikes also now ignore
  disadvantage."
- **As One! (L15)**: "Attacks made with your Coordinated Strike also grant advantage and ignore
  all disadvantage."

Neither is expressible today. There is no attack roll-mode rule type (only `initiativeRollMode`,
`savingThrowRollMode` and `skillRollMode`), and `conditionalBonus.advantage` cannot be scoped to
a granted activation that resolves on a different actor. Both descriptions still state the rule,
so the clauses are visible to the player and applied at the table.

## Test Plan

`pnpm check` passes (208 files, 3364 tests).

Live-tested in a real v13 world across three Playwright suites; scripts and screenshots are in
`_bmad-output/classtests/`. To reproduce:

1. `pnpm build`, launch v13, open a world on this branch and let the migrations run.
2. `node _bmad-output/classtests/scripts/seed-commander-demo.mjs` seeds a fully levelled L20
   Champion of the Vanguard with two allies in an active combat, then leaves it in place.
3. Join as Gamemaster. On the Commander's Features tab, Coordinated Strike! shows one charge
   badge reading `8/8` (INT 4 + 3 from Master Commander + 1 from Experienced Commander).
4. Target an ally, click the feature. Expect: no action debited, the badge drops to 7/8, the
   chat card shows a granted weapon-attack offer plus one line per recipient, and the top
   tracker shows a pending marker on the allies.
5. Accept the offer as the ally. Their action count should net to zero, not gain one.
6. Use it twice in one round. The second is refused with a visible notification naming the
   per-round limit.
7. At L20, use it twice in one encounter. Both uses grant the capstone action: the
   once-per-encounter gate is not in this PR. Verifying that gate belongs to #896, where a
   49-assertion suite covers it.

## Checklist

Note on the unit-test box: per project convention this PR ships no migration tests. Engine
changes and pack-parity are covered by ordinary unit tests.

## Related Issues

Part of #365, and of the Commander epic #685.

Refs #613, #615, #616, #617, #618, #619 - each is a checklist issue covering more than this PR
delivers, so none is closed here. The unexpressible items are listed above.

Builds on the action-economy foundations from #582.


trevlar (changes requested):
Conflicts need resolved.

trevlar (dismissed):
**🔧 Push back — one description fix. The rules implementation itself checks out against the rulebook, feature by feature.**

I checked this against `Commander Levels.md`, `Combat Abilities & Weapon Mastery.md` and both subclass files, not just for internal consistency. Every mechanic you shipped matches the rules as written. I also verified migration↔pack byte parity across all 15 feature specs and every `superseded*` guard against `origin/dev`; local `pnpm check` is green (208 files, 3364 tests).

**Must fix before merge**

- **`packs/.../commanders-order/coordinated-strike.json:153`** — The summary describes all three capstone riders as "gated once per encounter", and Test Plan step 7 tells a reviewer the capstone grant fires on the first use only. That's wrong in both directions:
  - **As One!** (Vanguard L15) has *no* once-per-encounter limit in the rules — "Your chosen allies gain 1 additional action to use on their next turn," full stop. Shipping it ungated is **correct**; the summary misdescribes working behaviour.
  - **Unparalleled Tactics** (L18) and **Captain of Legions** (L20) *do* say "The first time each encounter", and that gate is genuinely absent — as commit `23bc1ff` states, it isn't expressible, and I confirmed `actionEconomySystem.ts` has no per-encounter tracking for `actionDelta`.

  Split the bullet so As One! isn't lumped in, and fix step 7 — as written it will make a reviewer file a false bug against you.

- **`packs/.../commander-progression/master-commander.json:87`** — Pre-existing description text, but it's what tripped me up: "Levels 5, 9, 13, 17: Gain +1 use of Coordinated Strike per Safe Rest." The rulebook puts the +1 only on **Master Commander (2)/(3)/(4)** at L9/13/17; L5 is the initiative-regain and ignore-disadvantage clauses. **Your three `modifyPool` rules are right** and the `(2)/(3)/(4)` labels match the book exactly — the description is the thing that's wrong. Since you're in the file, dropping the stray `5` would stop the next reader (or the next agent) concluding you're short a use.

- **`packs/.../commanders-order/coordinated-strike.json:117`** — Two rules clauses aren't in your "recorded rather than faked" list, and by your own standard they should be: Master Commander L5's "Attacks made from your Coordinated Strikes also now ignore disadvantage", and As One!'s "Attacks made with your Coordinated Strike also grant advantage and ignore all disadvantage". Not implementing them looks right to me — there's no attack roll-mode rule type (only `initiativeRollMode`/`savingThrowRollMode`/`skillRollMode`), and `conditionalBonus.advantage` can't be scoped to a granted activation resolving on another actor — but the commit bodies are silent on both.

**What's good**

- Rules fidelity is high across the batch: Coordinated Strike's "(1/round) Free action… INT times/Safe Rest", 1/encounter on Hold the Line! and I Can Do This ALL DAY!, 1/turn on Taunting Strike, and the Combat Tactics flavour line all match the book. The L18–19 predicate band is a neat, exactly-right encoding of "(replaces Unparalleled Tactics)".
- Move it! Move it! was unambiguously broken — rolling `1d20+initiative` as *thunder damage* against the target for what the book calls an initiative roll — and the replacement note restates the rule faithfully. Same for dropping `maxHitDice` off a feature about *spending* Hit Dice.
- The `superseded*` guard design is the right call and implemented precisely — I verified every guard against `origin/dev` and they're exact, so a GM's hand-edit aborts the swap rather than getting stomped.

---

1. Split As One! out of the once-per-encounter claim; fix step 7.
2. Drop the stray "5" from Master Commander's description text.
3. Record the two unimplemented advantage/disadvantage clauses.

Fronix:
Thanks for checking this against the rulebooks feature by feature rather than just for internal consistency, and for verifying the migration/pack parity and the `superseded*` guards independently. All three points are addressed.

**1. As One! and the capstone gate.** You're right on both halves, and the second half has a wrinkle you had no way to see from this PR.

As One! has no per-encounter limit in the rules and ships without one, so the summary was describing working behaviour as something it isn't. Split out now.

The gate on Unparalleled Tactics and Captain of Legions does exist, but in #896, the branch stacked on this one. It's a hidden once-per-encounter charge pool, predicated `level >= 18` and referenced by those two riders only, deliberately not by As One!, which is the same split you asked for. It wasn't expressible when `23bc1ff` was written; making it expressible was part of the follow-up work, so it landed there rather than here.

Step 7 was the actively harmful part and is rewritten: it now says both uses grant the capstone action, because on this branch they do, and points at #896's 49-assertion suite for the gate.

**2. Master Commander's description.** Fixed in a6420583, and thank you, this one was a real trap. `Heroes-2:383` puts level 5 at the initiative regain and the ignore-disadvantage clause only; the `+1 use` arrives with (2)/(3)/(4) at 9/13/17. The stray `5` implied a fourth use the rules don't grant and made the automation look one short.

Since it's a shipped item, the fix needed migration cover, so `Migration036` now carries descriptions too, guarded the same way as the activation data: the swap fires only while the field still holds exactly what the pack shipped before, so a hand-edited copy is never stomped. No new migration file, so `LATEST_SCHEMA_VERSION` is unchanged.

**3. The two roll-mode clauses.** Fair by my own stated standard, and your reasoning for why they're not implemented matches what I found: there's no attack roll-mode rule type, only `initiativeRollMode` / `savingThrowRollMode` / `skillRollMode`, and `conditionalBonus.advantage` can't be scoped to a granted activation that resolves on a different actor.

I've recorded both in the PR description's unexpressible list rather than amending the commit bodies, since rewriting history would force a rebase of the stack for no functional gain. Both descriptions still state the rule, so the clauses are visible to the player and applied at the table.

**Stack note.** Adding a6420583 moved #896's base, so I restacked it (19 commits, no conflicts) and force-pushed. Both PRs are green: #895 at a6420583, #896 at 6dd8386e.

Fronix:
@trevlar could you check #896 as well, that one has the more important stuff

trevlar (approved):
**✅ Approved**

All three items from the last pass are resolved, and the only code delta since `56d1ae4a` is precisely the scope I asked for — the stray `5` dropped from Master Commander's description plus the matching carry-over in `Migration036`. Nothing else in the diff moved.

- The description migration is done right: I diffed `Migration036`'s `description` / `supersededDescription` byte-for-byte against the new pack file and `origin/dev`'s, and both match exactly, so a hand-edited copy still aborts the swap. No new migration file, `LATEST_SCHEMA_VERSION` unchanged, and no overlap with `037`/`038` on that feature.
- The As One! split holds up against the book: `Champion of the Vanguard.md:11` gives As One! no per-encounter limit, and `Commander Levels.md:118` scopes "(replaces Unparalleled Tactics)" to Captain of Legions only — which is exactly what the `subclass:champion-of-the-vanguard` + `level>=15` / `18–19` / `>=20` predicate bands encode. Step 7 now describes what the branch actually does, and the two roll-mode clauses are recorded in the summary.

Recording the unimplemented clauses in the PR body rather than amending commit bodies was the right call — no reason to force a restack of #896 for a documentation fix. I'll pick up #896 next.

Issue #885, opened 2026-07-31
[Foundation] Rule type: grantActivation (granted activation offers)


Fronix:
Declarative rule for "an ally immediately makes a weapon attack": offers stamp onto the granting item chat card, render only for owners of the target actor and GMs, run the recipient normal activation flow (costs resolve via actionCost), and stamp used idempotently through the primary GM with executor-side revalidation. Part of #582. Implemented by #880.

Fronix:
Closed by #880

Issue #881, opened 2026-07-31
[Foundation] Allow zero-cost activations


Fronix:
Make `activation.cost {type: action, quantity: 0}` representable so features worded "for free" stop debiting an action (the schema, `resolveItemActionCost`, and `consumeCombatantAction` all floored the cost at 1). Part of #582. Implemented by #876.

Fronix:
Closed by #876

Issue #365, opened 2026-02-28
[EPIC] AUTOMATION - PHASE 2


Fronix:
## What this epic tracks

Every feature, skill and spell in the Nimble rulebooks that needs some kind of automation work in the FoundryVTT system. "Automation" here means anything that makes a mechanic work in the system: a rule added to a pack file, code to handle an interaction, a sheet widget, or UI plumbing. It does not mean fully autonomous. Manual steps are fine as long as the mechanic is supported somehow.

The goal is a single place to see what's still outstanding across the whole game, what's ready to work on, and what depends on what.

## Structure

GitHub caps parents at 100 sub-issues, so this epic is organised as a tree rather than a flat list.

**Foundation layer**

- [EPIC] Foundation rule types (#681). Defines the rule-engine primitives (damageBonus, applyCondition, resourceCounter, toggleEffect, dicePool, conditionalBonus, damageReduction, grantMovement, grantAction, modifyIncomingAttack, triggerOnEvent, conditionImmunity). Many feature issues across all classes are blocked by items here, so this is the gating work.

**Per-class epics**

One epic per core class, each containing all the automation work for that class:

- [EPIC] Hunter automation (#682)
- [EPIC] Berserker automation (#683)
- [EPIC] The Cheat automation (#684)
- [EPIC] Commander automation (#685)
- [EPIC] Mage automation (#686)
- [EPIC] Oathsworn automation (#687)
- [EPIC] Shadowmancer automation (#688)
- [EPIC] Shepherd automation (#689)
- [EPIC] Songweaver automation (#690)
- [EPIC] Stormshifter automation (#691)
- [EPIC] Zephyr automation (#692)

Inside each class epic you'll find a mix of two kinds of issue:

- **Flagship issues**: one per novel or complex feature (e.g. Hunter's Mark, Rage & Fury Dice, Spellblade Arcane Command). These get their own thread because they have multi-step implementation paths or non-trivial interactions.
- **Checklist issues**: groupings of smaller, similar features (progression passives, ability pools, subclass feature lists). Each checkbox is an individual feature with a rulebook quote and the rule types it needs. Good for knocking out quick wins.

Every issue includes the source rulebook quote, the rule type or types it needs, and which foundation issues block it.

**Out of scope for now**

- Artificer and Hexbinder supplement classes. Will be tracked separately once they come into scope.
[...]

PR #896, opened 2026-08-04
fix(combat): action-economy defects found by live-testing the Commander

Fronix:
## Summary

Fixes six defects in the action-economy engine from #582, found by live-testing the Commander
work in the PR below this one. Two of them made granted actions actively wrong rather than
merely incomplete, and every future class using `grantActivation` or `actionDelta` would have
inherited them.

Stacked on #895. Review that one first; this branch depends on it.

## Changes

**Engine defects**

- **Granted actions landed a turn late.** A pending grant was only folded into the pool by the
  turn-end refill, so a grant written after the recipient's turn had already ended waited for
  the *next* turn end. The bug was order-dependent, which is why unit tests missed it. An
  additive fold now also runs at turn start.
  The turn-end refill is deliberately untouched: heroic reactions are usable only while it is
  not your turn and are paid from the same pool, so between turns that pool already is the
  upcoming turn's pool. Rebuilding it at turn start would refund reaction costs.
- **`allAllies` included the granter**, so a capstone that grants every ally an action also
  handed one to the character using it.
- **Granted attacks used a stale target.** An offer is created by targeting the recipient, so
  accepting without retargeting fired the attack back at them, with an Apply Damage button and
  no warning. The card now names the current target and calls out that case. It does not
  prevent it: who to target is a table decision.

**New generic vocabulary**

- Charge pools emit their state as domain tags, mirroring the dice-pool precedent.
- Charge pools and consumers honour their own `predicate`, which was previously parsed and
  silently ignored. Together these make "the first time each encounter, also do Y" expressible.
- A charge pool can opt out of being displayed, so a pool that only rate-limits a feature stops
  reading as a second budget next to the one the player actually spends.

**Visibility**

- Action grants are reported on the activation chat card and in the top combat tracker.
- The pending-action strings were pluralised and corrected; they said "at the next action
  refill", which is no longer where the action lands.
- Charges now appear on the actions tab, on all four panels that list the character's own items.
  The badge component resolves its own pools when not given them, which is what stops a new
  surface from silently showing none.

Nothing here disables, dims or gates anything. An empty pool renders a normal, clickable row;
the existing notification explains a refusal.

## Test Plan

`pnpm check` passes (210 files, 3312 tests).

Live-tested in a real v13 world; scripts and screenshots in `_bmad-output/classtests/`. The
suites re-run against this branch:

- `action-economy-fixes.mjs` (52/54), `coordinated-strike.mjs` (38/38),
  `coordinated-strike-riders.mjs` (35/36), `coordinated-strike-encounter-gate.mjs` (49/49).

**The check that matters most**, because it is what the turn-start fold could have broken:
take a heroic reaction between turns while a grant is pending, then let the turn start. The
reaction's cost must still be deducted. Observed pool of 2 where an additive fold predicts 2
and a rebuilt pool would show 4, silently refunding the reaction. `action-economy-fixes.mjs`
asserts this.

Also worth exercising by hand: use a granted offer while the recipient is still targeted and
confirm it warns but still works; and check the actions tab shows charge badges that open the
adjust dialog in place.

## Checklist

The `dev` box is unchecked on purpose: this is the upper half of a stack and targets
`feat/commander-coordinated-strike`. It should retarget to `dev` once the lower PR merges.

## Related Issues

Fixes defects introduced by the action-economy foundations in #582 (implemented across
#881 through #885). No separate issues were filed for these defects; they were found by live
testing rather than reported, and the evidence lives in `_bmad-output/classtests/`.

Part of #365.

## Known limitations

- `allAllies` still has no distance filter, so a range-limited grant reaches every ally in the
  encounter. That needs the v14 token-adjacency primitive.
- Backward turn navigation dispatches no turn events, so rewinding onto a combatant who is owed
  a grant defers it to their next turn end rather than folding it. Deferred, not lost.
- All live verification was single-GM. The socket relay path for a player-owned recipient
  accepting an offer on another client is untested.


trevlar (changes requested):
**🔧 Push back — the engine fixes are solid, but the granted-offer targeting UI states something false in the multi-target case and the new charge badges re-resolve the whole actor's pools per row**

The three engine defects (turn-start additive fold, `allAllies` excluding the granter, the pending-action wording) are well-reasoned and well-covered. My concerns are all in the new UI/vocabulary layer on top.

**Must fix before merge**

- **`src/view/chat/components/GrantedActionOffers.svelte:191`** — the `targetingRecipient` notice interpolates *every* target name into a clause that claims they are the one acting. With the recipient and a goblin both targeted it renders "Currently targeting Sir Brannon, Goblin Cutthroat, who is the one taking this action." That's factually wrong, and it's the common case (offers are created by targeting the recipient, so the recipient plus a real target is exactly what a retargeting user produces). Name only the recipient in that clause and list the other targets separately, or split it into two sentences. `resolveGrantedOfferTargeting` already gives you `targetsRecipient` separately from `targetNames` — you just need the recipient's own name too.

- **`src/view/chat/components/GrantedActionOffers.svelte:101`** — `isDiscouraged` tests `CONST.TOKEN_DISPOSITIONS.FRIENDLY` absolutely, not relative to the recipient. Two failures fall out: a GM-run hostile recipient granted an action and correctly targeting a PC gets the discouraged styling, and a granted *support* action aimed at an ally (heal, buff) gets it too. Compare the target's disposition against the recipient token's own disposition, or narrow the dim to `targetsRecipient` only. Related: 0.45 opacity reads as "disabled" to most users, which contradicts your own "nothing here disables, dims or gates anything" — consider a text/icon cue instead of an opacity drop.

- **`src/view/chat/components/ActionDeltaSummaryNode.svelte:25`** — `getAdjustmentLocalizationKey` is eight untested branches of singular/plural key selection, and it duplicates the logic you *did* extract and test as `src/utils/pendingActionWording.ts`. Pull it into a sibling plain-`.ts` helper and give it the same table-driven test; a wrong key here silently renders a raw `NIMBLE.chat.…` string on the card.

- **`src/view/sheets/components/CastSpellActionPanel.svelte:80`** (same at `CustomReactionsPanel.svelte:62`, `AttackActionPanel.svelte:132`/`:149`, `OpportunityAttackPanel.svelte:116`) — these call sites omit `pools`, so each row runs `getPoolsForItem` → `getPools` → `buildEffectiveChargePoolMap(actor)`, which walks every item and rule on the actor and calls `actor.getRollData()` per formula. That's a full rebuild per row, re-run on every invalidation of `actor.reactive`. Your own prop doc on `ChargeIndicatorProps.pools` says to batch for long lists — and the actions tab is the longest one on the sheet. Follow `PlayerCharacterFeaturesTab.svelte:180`: one `allPools = $derived(getPools(actor.reactive))` per panel, then `pools={getPoolsForItem(actor.reactive, id, allPools)}` per row.

- **`src/hooks/actionEconomySystem.ts:207`** — this and `handleUseItemGrantedOffers` (`:240`) now both fire un-awaited `message.update()` on the same ChatMessage in the same tick, and Coordinated Strike carries both `actionDelta` and `grantActivation` rules, so that's the shipping case rather than a hypothetical. Worth folding into one update (or one handler) rather than relying on the server merging two concurrent diffs.

- **`src/utils/chargePool/helpers.ts:324`** — a pool whose predicate goes false is dropped from the definitions, and `persistChargePoolMap`/`buildPoolUpdatePayload` then writes `-=<poolId>` and deletes its stored state. That's harmless for a monotonic gate like `level >= 18`, but `docs/documentation/rules-builder/predicates.md` now advertises pool predicates generally — an author who gates a pool on something reversible (a toggle, equipped armor) gets a silent full refund on the next flip. Worth calling out in the docs you added.

- **`src/view/chat/utils/resolveGrantedOfferTargeting.ts:17`** — the doc comment promises `targetNames` "in target order", but the only call site builds the array as `[...getInvalidTargets(id), ...getTargetedTokens(id)]`, so the recipient's tokens are always hoisted first. The unit tests pass their own ordering and so never see this; either fix the comment or sort at the call site.

- **`src/utils/chargePool/chargePoolTags.ts:363`** — the header says the empty-pool tag is "camelCased", but `capitalize` only upcases the first character, so the shipped pack identifier yields `self:noCoordinated-strike-encounterCharges`. Also worth a line in `docs/system/rules.md`: two item-scoped pools sharing an identifier collapse into one tag key, and `min` requires *every* matched value to pass, so an unrelated empty `uses` pool can falsify a rider predicated on a different item's `uses` pool. Authors should namespace identifiers the way you did here.

**What's good**

- The turn-start fold's "additive, never a reset" reasoning is exactly right, and the regression test that a heroic reaction paid between turns stays deducted is the right test to have written — that's the one a naive rebuild would have broken.
- Both migrations match conservatively on the exact superseded label/predicate and bail on any GM edit, so a retuned copy survives.

---

1. Fix the multi-target `targetingRecipient` notice — it states something false.
2. Make `isDiscouraged` relative to the recipient's disposition, not absolute.
3. Extract and test `getAdjustmentLocalizationKey`.
4. Batch charge-pool resolution in the four new panels.
5. Merge the two concurrent chat-message updates.
6. Document that a false pool predicate deletes stored charges.
7. Fix the "in target order" comment or sort at the call site.
8. Fix the "camelCased" claim; document charge-tag identifier collisions.

PR #899, opened 2026-08-04
Add size options UI control to ancestry sheet

trevlar:
## Summary

Ancestries already store a `system.size` array in their data model, and the character-creation flow reads it to offer a size choice. But the Ancestry item sheet had no control to edit that array, so custom ancestries could only set their size options by editing the underlying JSON or via the console. This PR adds a UI control to set those size options directly on the sheet.

## Changes

- Added a **Size Options** control to the Config tab of `src/view/sheets/AncestrySheet.svelte`.
- The control is a multi-select `TagGroup` (reusing the existing component) populated from `CONFIG.NIMBLE.sizeCategories`, bound to `system.size`.
- Toggling a tag adds/removes it from `system.size`; selections are always persisted in the canonical category order regardless of click order.
- Labels use the existing `NIMBLE.sizeCategories.*` localization keys via `localize()` — no hardcoded strings.
- Added a `Hint` explaining that selecting more than one size lets the player choose during character creation (matching the existing character-creation behavior, which only shows a size picker when an ancestry offers multiple sizes).

## Test Plan

1. Open (or create) an Ancestry item and go to the **Config** tab.
2. Confirm a **Size Options** section appears with a tag for each size (Tiny, Small, Medium, Large, Huge, Gargantuan).
3. Toggle sizes on/off and confirm they persist to `system.size` (reopen the sheet or inspect the item's data).
4. Confirm the stored order is always canonical (e.g. selecting Large then Small yields `["small", "large"]`).
5. Create a character with an ancestry that has **more than one** size selected and confirm the character-creation dialog presents the size choice; with a single size, confirm no size picker appears.
6. Check the console for errors while interacting with the sheet.

## Checklist

- [ ] `pnpm check` passes (format, lint, circular deps, type-check, tests) — could not run locally; `pnpm install` is blocked in this environment by an egress policy on a Git-hosted dependency (`foundry-vtt-types/pixi-utils`). Please rely on CI.
- [x] No hardcoded user-facing strings (used `localize()`) — size labels are localized; the inline `Hint` text follows the same inline-copy pattern already used elsewhere in this sheet.
- [ ] Tested in Foundry with no console errors — not run in this environment.

## Related Issues

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01QXPwjHTf1SFDy2JZyYrUZL)_


Fronix (dismissed):
## Blockers

**1. Deleting your only size changes it to Medium** — `sizeSelection.ts:39`

Half-Giant is `['large']`. GM clicks the ✕ on the Large badge. Result: the ancestry is now **Medium**. The rule being enforced is "at least one size," but the fallback throws away the GM's value instead of keeping it.

Fix: `return remaining.length ? remaining : currentSizes;` — a no-op. That also removes the need for the `revertedTo` state and the wasted `item.update()`.

Same thing via the "All sizes" ✕, except that path shows no explanation at all.

**2. New ancestries display "Medium" but store `[]`** — `AncestryDataModel.ts:8`

`initial: () => []`, so every freshly created ancestry is empty. `effectiveSizes` renders that as Medium, and the summary says "All characters will be Medium" — but nothing is written until the GM clicks something. `prepareAncestryMetadata` doesn't share that fallback, so the ancestry card in character creation shows *no size* for an ancestry whose sheet says Medium.

Fix: `initial: () => ['medium']`. (Also: the comment in `sizeSelection.ts` calls this a legacy case. It isn't — it's the default for everything.)

---

## Worth fixing

**Two helpers disagree about unknown sizes.** `toggleSize` preserves values the config doesn't define (there's a test for it); `toggleAllSizes` discards them, and `sizeSelection.test.ts:96` locks that in. Pick one policy.

**Keyed `{#each}` on a key that can repeat.** Both new lists key on the size string, and `ArrayField` doesn't dedupe — an imported ancestry with `['medium','medium']` throws a Svelte duplicate-key error and blanks the sheet you'd use to fix it. `[...new Set(...)]` in `effectiveSizes` closes it.

**`hasSizeChoice` is now computed two ways.** `state.svelte.ts:110,122` still does `size?.length > 1`; the component derives it from `effectiveSizes`. They agree on current data, but the PR just introduced the canonical helper — the state machine should use it.

**Accessibility:**
- `SizeOptionsSelect.svelte:60` — `aria-label` overrides the visible button text, breaking WCAG 2.5.3. `aria-expanded` already covers the state; drop it and add `aria-controls`.
- `AncestrySizeSelection.svelte:96` — `role="radiogroup"` with `<li>` between group and radios breaks the ARIA relationship. Add `role="none"` to the `<li>`s. Neither list has arrow-key navigation.

**Hardcoded colors** in `SizeOptionsSelect.svelte` (`rgba(255,255,255,0.18)` etc.) against `STYLE_GUIDE.md:1057` "Never hardcode colors."

---

## Nits

- Trigger label reads backwards: at 3 of 6 selected it says "Add a size…", only at 6 of 6 does it say "Edit sizes…".
- `sizeSelection.ts` sits in `sheets/components/` but is imported by a dialog four levels up — it's shared now, so `src/utils/`.
- Bonus: `prepareAncestryMetadata` maps to labels *before* sorting by key index, so `indexOf` is always -1 and the sort is a no-op. Swapping it to the new `sortBySizeOrder` fixes that for free.
- `es.json`/`fr.json` still carry the removed `ancestryOptions.header`.

trevlar:
## Fixed

- **Removing the only size fell back to Medium** — a world copy now reverts to the size the compendium ancestry it was copied from shipped with, so Half-Giant's ✕ gives Large back; only an ancestry written from scratch lands on the system default (`sizeSelection.ts:41`, `AncestrySheet.svelte:42`).
- **"All sizes" ✕ explained nothing** — both removal paths route through one `applyChange`, which names whatever came back (`SizeOptionsSelect.svelte:47`).
- **New ancestries displayed Medium but stored `[]`** — `initial: () => [DEFAULT_SIZE]` (`AncestryDataModel.ts:10`).
- **Ancestry card showed no size where the sheet showed Medium** — `prepareAncestryMetadata` reads through `effectiveSizes`, which also fixes the label-vs-key sort being a no-op (`prepareAncestryMetadata.ts:8`).
- **Helpers disagreed about unknown sizes** — one policy now: a removal keeps whatever is left, undefined values included, and only a genuinely empty result falls back. `toggleAllSizes` no longer discards them in either direction (`sizeSelection.ts:60`).
- **Duplicate `{#each}` keys** — `effectiveSizes` dedupes (`sizeSelection.ts:17`).
- **`hasSizeChoice` computed two ways** — the state machine calls the new `offersSizeChoice` (`state.svelte.ts:107`).
- **`aria-label` overrode the trigger's visible text** — dropped; `aria-controls` points at the panel, which is always rendered and `hidden` when collapsed so the reference isn't dangling (`SizeOptionsSelect.svelte:95`).
- **`<li>` between radiogroup and radios** — the radios are the group's own children now, with roving tabindex and arrow keys carrying the selection (`AncestrySizeSelection.svelte:123`).
- **Hardcoded rgba on the badge chips** — `color-mix` from `currentColor`, which drops the `--all` override too (`SizeOptionsSelect.svelte:367`).
- **Trigger label read backwards** — "Edit sizes…" from two selected, not only at six (`SizeOptionsSelect.svelte:98`).
- **`sizeSelection.ts` under `sheets/components/`** — moved to `src/utils/` with its test.

## Not fixed

- **Stale `ancestryOptions.header` in `es.json`/`fr.json`** — AGENTS.md forbids agents touching non-English locale files. The key is unreferenced, so it's inert until someone regenerates those files.
- **No migration for world ancestries already storing `[]`** — reads go through `effectiveSizes`, so they present and behave as Medium; a migration would only be cosmetic.

## Approach

On the first blocker: the no-op fix drops the rule instead of enforcing it, and "keep the current value" is only correct while the current value is the authored one. Since compendium ancestries are locked, an editable ancestry is a world copy, and its compendium source is the authoritative default — so the ✕ restores *that*, and Medium is the floor for ancestries with no published version. The notice explains which it did.

The unknown-size policy is now the same sentence in both helpers: remove, keep the remainder, fall back only when nothing remains. That makes `toggleAllSizes(['tiny'…'gargantuan','colossal'])` yield `['colossal']` rather than Medium.

Pushed as 3b679b7. Thanks @Fronix — 3383 tests, lint, typecheck, and dep-cruiser all green.

Fronix (approved):
LGTM

PR #900, opened 2026-08-05
fix(conditions): apply to the card's targets and record their source

Fronix:
## Summary

Conditions were applied from two places that each rolled their own logic, so they had drifted
apart. One bypassed condition immunity, both applied to the wrong creatures in one case, and
neither recorded where the condition came from. This centralises application and fixes both.

image

## Changes

- **New shared helper** owning the whole flow: dedupe, the blocking `preApplyCondition` hook,
  creation, duration, and the `conditionApplied` hook. Both callers now use it.
- **A card's condition button applies to the card's targets**, not to whatever tokens the
  clicking user happens to have selected on the canvas. With several tokens selected it used to
  condition all of them; with the wrong one selected, the wrong creature.
- **The button is GM-only**, gated the same way as the card's damage buttons. A card's targets
  are rarely tokens a player owns, so a player's click could only be refused by the server on
  permissions. Players keep the chip and its description tooltip; the apply affordance and hint
  render only for GMs. Each target is also applied independently, so one refusal cannot stop
  the conditions the remaining targets can still take.
- **Conditions record a source.** The readout resolves a source name from the effect's `origin`,
  and nothing was setting one, so every condition this system applied read as "Source: None".
  Preference is the applying item, then the applying actor, and omitted when neither is known,
  so a condition toggled by hand from the token HUD stays sourceless rather than being
  attributed to something it did not come from. Those manual paths are untouched.

Application mirrors what `Actor#toggleStatusEffect` does internally rather than calling it,
because that method accepts no origin. Creation keeps the id, so the static ids reserved for
linked-status conditions still dedupe.

Two incidental improvements over the path it replaces: duration is stamped before creation
rather than patched after, so it is one document write and creation listeners see the real
duration; and dedupe now also catches a disabled effect holding the status, which the actor's
status set does not report because it is built from active effects only.

## Behaviour change worth flagging

**Condition immunity now applies to the chat-card button.** It previously called the status
toggle directly, firing neither `preApplyCondition` nor `conditionApplied`, so
`conditionImmunityGuard` was bypassed entirely: a click could stick a condition on a creature
immune to it, and downstream listeners never heard about it.

The button can therefore now refuse where it previously always succeeded. The old behaviour was
a bug, but it is the kind of change that surprises someone mid-session, so it should not land
silently.

**Players also lose the apply affordance entirely.** Previously a player's click applied the
condition to their own selected tokens, which was the targeting bug above; with targeting fixed
there is nothing useful a player's click can do, so the button is inert for them. If players
ever need to apply conditions to targets they do not own, the right shape is a socket relay to
the active GM, as `requestCombatantActionDelta` does; that is deliberately out of scope here.

An earlier revision of this description claimed a non-GM permission failure "matches the
previous behaviour and is not a regression". Review rightly pointed out that was wrong: the old
path walked controlled tokens, which a player by definition owns, so it could never fail on
permissions. The GM gate resolves this by not offering players the failing path at all.

## Test Plan

`pnpm check` passes (210 files, 3381 tests), 42 of them new or reworked here.

Unit tests cover the helper directly (dedupe including the disabled-effect case and the reserved
static id, the blocking hook preventing both creation and the applied hook, origin from item,
fallback to actor, omitted when neither, duration stamping, `keepId` on create) and the card node
(targets resolved from the card with a deliberately non-empty canvas selection present, to prove
the selection is ignored; one refused target not aborting the rest; the player-facing chip being
inert).

To check it in Foundry:

1. Put two or more tokens on a scene and select all of them, but target only one.
2. As the GM, use a feature whose card offers a condition button, and click it. Only the
   targeted creature should gain the condition. Before this change, every selected token did.
3. Open the conditions readout on that creature. "Source" should name the feature rather than
   reading "None".
4. Toggle a condition by hand from the token HUD. "Source" should still read "None"; manual
   application is deliberately sourceless.
5. Give a creature immunity to a condition, then try the card button on it. It should now be
   refused rather than applied.
6. As a player, the condition chip on a card should render without the apply hint and do
   nothing when clicked.

## Checklist

The Foundry box is unchecked honestly: this is verified at unit level only. The steps above are
what a reviewer should run.

## Related Issues

None filed; both defects were found while live-testing Commander automation rather than reported.

Independent of the open stack (#895, #896) and mergeable on its own: it is based directly on
`dev`, adds no migration, and shares no code with either branch. It does touch
`public/lang/en.json` and `docs/STYLE_GUIDE.md`, which that stack also appends to, so whichever
merges second may need a trivial conflict resolution in those two files.

## Known limitations

- Conditions remain largely inert. `taunted`, `prone`, `frightened` and `restrained` have no
  mechanical handling anywhere in the system; they are markers. This PR makes application correct
  and attributable, not meaningful.
- `src/enrichers/enrichCondition.ts` still renders an inert button inside descriptions with no
  click handler anywhere. Its tooltip no longer promises a click action; giving it a real
  handler is a separate decision.
- `src/hooks/automaticConditions.ts` still applies derived conditions through
  `toggleStatusEffect` with its own immunity check rather than the shared helper, so those
  conditions carry no source yet. Routing that path through the helper is follow-up work; the
  style guide now says so rather than claiming the helper is universal.


trevlar (changes requested):
**🔧 Push back — retargeting the button to the card's targets makes it fail for players, and the PR body reasons about that as a non-regression when it isn't**

The centralisation itself is good work, and the helper is the right shape. One thing needs a decision before merge.

**Must fix before merge**

- **src/view/chat/components/ConditionNode.svelte:69** — This button is rendered to every user, not just the GM (`ItemCardEffects.svelte:35` renders condition nodes unconditionally; the Apply Damage button right below it is gated on `game.user?.isGM`). The old handler walked `canvas.tokens.controlled`, which a player can only ever populate with tokens they own, so `toggleStatusEffect` always had permission. The card's targets are usually creatures a player does *not* own, so `ActiveEffect.create` rejects and Foundry throws up a "lacks permission" toast instead of applying anything. Your PR body calls this "not a regression" because the old path could fail too — it couldn't, for the same reason: controlled tokens are owned by definition. So this is a live path a player will hit on nearly every attack card.

  Pick one and make it deliberate: gate the button on `game.user?.isGM` the way `applyAllDamage` does (`src/documents/chatMessage.ts:995`), relay the application to the primary GM via the existing socket pattern (`src/utils/getPrimaryActiveGmId.ts`, `requestCombatantActionDelta.ts`), or keep it open and catch the rejection with a localized `ui.notifications` message. Any of those is fine; the current unhandled rejection isn't.

- **src/view/chat/components/ConditionNode.svelte:69** — The sequential `await` in the loop means the first target that throws skips every remaining target, so a mixed-ownership target list applies to nobody past the failure. The old `forEach` fired each independently. `Promise.allSettled` or a per-target `try`/`catch` keeps one bad target from eating the rest.
- **docs/STYLE_GUIDE.md:1468** — "The one path every automated condition application goes through" isn't true yet: `src/hooks/automaticConditions.ts:60` still applies derived conditions with `toggleStatusEffect` and its own `isConditionImmune` check rather than the hook. Either route that path through the helper too or soften the wording, otherwise the next person trusts the table and misses a second implementation.
- **src/view/chat/components/ConditionNode.svelte:70** — `applyConditionToActor(targetActor as never, …)` defeats the helper's own docstring ("kept structural so a caller can pass a stand-in document without a cast"). Type `resolveTargetActors()` as returning `ConditionTargetActor[]` and the cast goes away.
- **src/enrichers/enrichCondition.ts:22** — Since you're already fixing this wording on the card, the enricher's footer is both hardcoded English and describes behaviour that doesn't exist (no handler anywhere). Dropping the footer line is smaller than leaving a tooltip that promises a click action.

**What's good**

- Stamping duration via `updateSource` before create instead of patching after is a real improvement — one write, and `_preCreate`/create listeners see the true duration.
- The dedupe genuinely covers what `Actor#statuses` misses (disabled effect holding the status, reserved static `_id`), and each of those cases has its own test. 40 tests for 190 lines of helper is the right ratio.

---

1. Decide what a non-GM click does — gate, relay, or handle the rejection
2. Correct the "not a regression" reasoning in the PR body
3. Don't let one failing target abort the rest of the loop
4. Soften or fulfil the "one path" claim in the style guide
5. Drop the `as never` cast by typing the resolved actors
6. Fix or remove the stale hardcoded enricher tooltip footer

Fronix:
All six points addressed, thanks for the catch on the ownership reasoning — you're right that controlled tokens are owned by definition, so the old path couldn't hit the permission failure and the new one would have on nearly every player click. The PR body now says so explicitly instead of calling it a non-regression.

1. **Non-GM click** — gated on `game.user?.isGM`, matching `applyAllDamage` (06d9c1bd). Players keep the chip and its description tooltip; the apply hint, aria-label, and handler render only for GMs. With targeting fixed there's nothing useful a player click can do, so gating seemed more honest than catch-and-toast; if players ever need this, the socket relay is the follow-up shape, and the body notes that.
2. **PR body** — rewritten: the "not a regression" claim is retracted, the gate is flagged as its own behaviour change, and the test plan gained a player step.
3. **Loop abort** — `Promise.allSettled` with rejected results reported, plus a test where the first target rejects and the second still applies (aded11ba).
4. **"One path" claim** — softened in both the style guide and the helper's own docstring, which had the same overclaim; both now point at `automaticConditions.ts` still going through `toggleStatusEffect`. Routing that path through the helper (which would also give derived conditions a source) is listed as follow-up rather than smuggled into this PR (c49f8f36).
5. **`as never`** — gone; `resolveTargetActors()` returns `ConditionTargetActor[]` (1375ff51).
6. **Enricher footer** — dropped entirely (0c92deb6).

`pnpm check` passes on the result (210 files, 3381 tests).

PR #901, opened 2026-08-05
feat: add a per-item "Skip roll dialog" toggle

### LINKED ISSUE #868: Add a per-item "Skip roll dialog" toggle
# Add a per-item "Skip roll dialog" toggle

## Problem or Motivation

There is no way to author an activated ability that resolves in one click. The activation dialog is opened by structure, not by intent:

- `ItemActivationManager#getData` opens `ItemActivationConfigDialog` whenever the effect tree contains **any** `damage` or `healing` node, even when the formula is a constant with no dice in it (`src/managers/ItemActivationManager.ts:140-148`).
- Every item of type `spell` opens `SpellUpcastDialog` unconditionally (`ItemActivationManager.ts:148`).

So a flat-value ability like Lay on Hands (`formula: "5*@level"`, no dice, `packs/classFeatures/core/oathsworn/oathsworn-progression/lay-on-hands.json:28`) makes the player click through a dialog offering advantage/disadvantage, situational modifiers, and a primary-die override, none of which do anything.

Beyond the no-dice case, authors have no way to say "this item never prompts" for an ability that *is* technically configurable. A signature attack that always rolls straight, a utility spell where the upcast panel is noise, an NPC feature meant to fire fast at the table.

The only escape hatches today are holding **Alt** while activating (`ItemActivationManager.ts:151-163`) undiscoverable, per-click, and impossible to bake into pack content or authoring the ability with no `damage`/`healing` node at all, which also drops the effect.

This came in from Discord: a user building a charge-consuming Lay on Hands, blocked on "is there a way to implement activated abilities without rolling dice?" The charge half already works `chargeConsumer` consumption fires off the `useItem` hook independent of any roll (`src/hooks/chargeSystem.ts:134-154`). It's only the dialog that's unavoidable.

## Proposed Solution

One persisted per-item flag that unconditionally suppresses the dialog. No heuristics, no formula inspection, if the author sets it, the item never prompts, even when the dialog would technically have something to offer.

### 1. Schema

Add to the shared `activation()` schema in `src/models/item/common.ts` (alongside `showDescription`, line 62):

```js
skipRollDialog: new fields.BooleanField({
    required: true,
    nullable: false,
    initial: false,
}),
```

Boolean with `initial: false`,  no data migration needed, existing content keeps today's behavior.

### 2. Manager

In `ItemActivationManager#getData`, treat the flag exactly like the existing `fastForward` option, which already does the right thing: bypass the dialog and build `dialogData` from `#getDefaultDialogData(rollOptions)` (`ItemActivationManager.ts:130-138`). This is a routing change to the branch at `:139-181`, not new machinery.

Precedence, most specific first:

1. `options.fastForward` explicit caller intent (macros, `item.activate()`), wins either way.
2. `system.activation.skipRollDialog` the new flag.
3. Alt held existing per-click skip.
4. Otherwise: today's `hasRolls || isSpell` check.

The flag applies to both dialog types, so a spell with the flag set skips `SpellUpcastDialog` and activates at its base tier. That is the intended reading of "never prompt"; an author who wants upcasting simply doesn't set the flag.

### 3. Sheet UI

The checkbox belongs on the item's **Config** tab, in the general-configuration area, next to identifier / object type / price, where an author looks when deciding how the item behaves. Not on the Activation tab, which is about what the activation *does*.

```
GENERAL CONFIGURATION
  Object Type      [Armor] [Shield] [Weapon] [Consumable] [Miscellaneous]
  Object Size Type [Slots required] [Stackable] [Small]
  Inventory Slots  [1]
  [ ] Skip the roll dialog (activate instantly)
```

`activation` is a shared schema, but every item type renders its own `configTab` snippet and there is no common host, `ObjectSheet.svelte:182` has a "General Configuration" header, `FeatureSheet.svelte:54` has "Feature Type", `MonsterFeatureSheet.svelte:64` has only "Description", and spells use `pages/SpellConfigTab.svelte`. Today a setting that applies to every item type has nowhere to live, and adding one means four hand-written copies that immediately start to drift.

So introduce a shared section component, `src/view/sheets/components/ItemGeneralConfigSection.svelte`, that renders its own `nimble-section-header` ("General Configuration") plus the cross-type fields, reading the item off the existing `document` context like the other config tabs do. Each sheet then mounts it with one line:

```svelte
<ItemGeneralConfigSection />
```

The skip-roll-dialog checkbox is its first occupant; the point of the component is that the *next* generic setting is a one-file change instead of four.

| Sheet | Placement |
|---|---|
| `ObjectSheet.svelte:182` | replaces the hand-rolled "General Configuration" header; type-specific fields (object type, size, slots) stay below it |
| `FeatureSheet.svelte:54` | after "Feature Type" |
| `MonsterFeatureSheet.svelte:64` | in the `configTab` snippet |
| `pages/SpellConfigTab.svelte` | spell config body |

Only these four sheets mount `ItemActivationConfigTab`, so they are exactly the activatable types. The component should degrade gracefully for item types that later mount it without an `activation` schema (ancestry, background, class, subclass, boon), render only the fields that apply, so it stays usable as the home for genuinely universal item settings.

New localization keys in `public/lang/en.json` (section heading, label, hint). The hint should note that Alt-click already skips the dialog for a single activation, so this toggle reads as "always".

### 4. Content

Set the flag on shipped flat-value content,  Lay on Hands is the obvious first case.

### Acceptance criteria

- With the flag set, activating the item never opens `ItemActivationConfigDialog` or `SpellUpcastDialog`; it rolls with defaults (roll mode 0, no situational modifiers) and posts its card.
- This holds even when the item has real dice and even when the spell is upcastable,  the flag is absolute.
- With the flag unset, behavior is byte-for-byte what it is today, including the Alt-click skip.
- `chargeConsumer` still consumes on activation in every path above.
- The toggle appears on the Config tab of all four activatable item sheets (object, feature, monster feature, spell), rendered by a single shared component,  adding a second cross-type setting touches one file.
- Existing content is unaffected until the flag is set.

## Alternatives Considered

- **Auto-skip whenever the resolved formula has no die terms**, plus lifting `canUpcast` (`SpellUpcastDialog.svelte:41-43`) out of the dialog so non-upcastable spells stop prompting. More "correct" in theory, but it makes prompting behavior implicit and formula-dependent,  an author editing `5` into `1d6` silently gains a dialog. It also can't express "never prompt" for an item that does have dice, which is half the ask. An explicit flag covers both cases and is far easier to reason about.
- **A world-level setting.** Too blunt; the same world has both one-click utility features and attacks that need the dialog every time.
- **Document Alt-click and stop there.** Zero code, but undiscoverable and unauthorable into pack content.

## Additional Context

Minor, adjacent: the activation card plays `CONFIG.sounds.dice` whenever `rolls.length > 0`, regardless of whether any die was actually rolled (`src/documents/item/base.svelte.ts:275`). A constant-formula healing card clatters like a dice roll. Easy follow-up, not required for this issue.

Also out of scope,  the other half of the Discord report is that Lay on Hands should let the player **spend any amount** from the pool and heal that much. There is no prompt-for-amount mechanism today:

- `chargeConsumer.cost` is a static formula resolved at activation (`src/models/rules/chargeConsumer.ts:25-33`).
- The activation dialog's charge stepper only renders for pools with a `dieSize` set (`src/view/dialogs/itemActivationConfigDialogHelpers.ts:58-70`), and each spent charge becomes an added *die* on the damage formula (the Fury Dice pattern),  it can't produce a flat healing value.
- Healing formulas cannot reference how much was spent; `situationalModifiers` only feeds the damage/attack formula.

That needs a variable-spend UI feeding a flat value and should be tracked separately. Note the shipped Lay on Hands currently heals the entire 5×LVL pool in one activation, which does not match the rules text ("spend any amount of remaining healing power to restore that many HP").

Fronix:
## Summary

Adds a per-item `system.activation.skipRollDialog` toggle that unconditionally suppresses the activation/upcast dialog, so authors can build one-click abilities like Lay on Hands. Closes the gap where the only skip was the undiscoverable per-click Alt hold.

## Changes

- New `skipRollDialog` boolean (initial `false`, no migration needed) on the shared `activation()` schema, so all four activatable item types get it from one definition.
- `ItemActivationManager#getData` dialog decision extracted into `#resolveDialogData` with precedence: `fastForward` option → item flag → Alt held → existing `hasRolls || isSpell` dialog gate. A flagged spell skips `SpellUpcastDialog` and casts at base tier. Also fixes a pre-existing leak where cancelling the dialog skipped the `keyPressStore` unsubscribe.
- One "Roll Dialog" checkbox in the shared `ItemActivationCoreConfigTab`, so it appears on all four item sheets (object, feature, monster feature, spell) with no per-sheet edits. **Placement deviates from #868's proposal** (Config-tab "General Configuration" section): the Activation > Core tab already hosts the analogous `showDescription` toggle and is a single shared component, so the shared-section component is deferred until a second cross-type setting exists.
- Two new English strings under `NIMBLE.itemConfig` (label + hint noting Alt-click as the one-shot equivalent) — flagged for human review per AGENTS.md.
- Shipped Lay on Hands now sets the flag.
- Dialog classes are now routed through the manager's existing `testDependencies` injection (behavior-identical) because module mocks cannot intercept imports preloaded by `tests/setup.ts`.

## Test Plan

- `pnpm test` — 5 new routing tests in `ItemActivationManager.test.ts` (flag on item, flag on spell, flag unset opens dialog, dialog cancel, fastForward precedence); full suite green.
- In Foundry: import Lay on Hands onto a character; on Activation > Core its "Skip the roll dialog" checkbox is pre-checked; activating posts the healing card with no dialog. Uncheck it and the config dialog returns; Alt-click still skips. Set the flag on any spell and it casts without the upcast dialog at base tier. A charge-consuming feature with the flag still decrements its pool on activation.
- Verified live on v13 via Playwright driving the real UI (checkbox clicks, sheet activation clicks); all cases above pass with no system console errors.

## Checklist

## Related Issues

Closes #868

### DIFF
--- a/packs/classFeatures/core/oathsworn/oathsworn-progression/lay-on-hands.json
+++ b/packs/classFeatures/core/oathsworn/oathsworn-progression/lay-on-hands.json
@@ -31,6 +31,7 @@
 				}
 			],
 			"showDescription": true,
+			"skipRollDialog": true,
 			"targets": {
 				"count": 1,
 				"restrictions": ""
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -2542,8 +2542,9 @@
 			"targetRestrictions": "Target Restrictions",
 			"attackType": "Attack Type",
 			"distance": "Distance",
+			"activationOptions": "Activation Options",
 			"skipRollDialog": "Skip the roll dialog (activate instantly)",
-			"skipRollDialogHint": "Always activate this item with default roll settings, without opening the roll or upcast dialog. Holding Alt while activating does the same for a single activation.",
+			"skipRollDialogHint": "Always activate this item with default roll settings instead of opening the roll or upcast dialog. Holding Alt while activating inverts this setting for that activation.",
 			"attackTypes": {
 				"none": "None",
 				"melee": "Melee",
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -2542,6 +2542,8 @@
 			"targetRestrictions": "Target Restrictions",
 			"attackType": "Attack Type",
 			"distance": "Distance",
+			"skipRollDialog": "Skip the roll dialog (activate instantly)",
+			"skipRollDialogHint": "Always activate this item with default roll settings, without opening the roll or upcast dialog. Holding Alt while activating does the same for a single activation.",
 			"attackTypes": {
 				"none": "None",
 				"melee": "Melee",
--- a/src/managers/ItemActivationManager.test.ts
+++ b/src/managers/ItemActivationManager.test.ts
@@ -1,6 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import type { EffectNode } from '#types/effectTree.js';
 import { MockRollConstructor } from '../../tests/mocks/foundry.js';
+import { keyPressStore } from '../stores/keyPressStore.js';
 import { findNodesByContexts } from '../utils/treeManipulation/findNodesByContexts.js';
 import { ItemActivationManager, testDependencies } from './ItemActivationManager.js';
 
@@ -77,15 +78,6 @@ const MockDamageRoll = vi.fn(function DamageRollMock(
 const DamageRoll = MockDamageRoll;
 const NimbleRoll = MockNimbleRoll;
 const getRollFormula = mockGetRollFormula;
-
-vi.doMock('../stores/keyPressStore.js', () => ({
-	keyPressStore: {
-		subscribe: vi.fn(() => vi.fn()),
-	},
-}));
-vi.doMock('../documents/dialogs/ItemActivationConfigDialog.svelte.js', () => ({
-	default: vi.fn(),
-}));
 
 // Dialog constructor mocks, injected through testDependencies because module
 // mocks cannot intercept the manager's dialog imports (tests/setup.ts loads
@@ -1299,6 +1291,7 @@ describe('ItemActivationManager.getData (rolls)', () => {
 	describe('Dialog routing', () => {
 		beforeEach(() => {
 			dialogState.result = undefined;
+			keyPressStore.set({ ctrl: false, shift: false, alt: false });
 		});
 
 		it('should skip the config dialog and complete activation when skipRollDialog is set', async () => {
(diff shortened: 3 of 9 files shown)


trevlar (changes requested):
**🔧 Push back — solid feature and a clean refactor, but one new user-facing string isn't localized**

The `#resolveDialogData` extraction is a genuine readability win over the nested `if/else` it replaces, the precedence order (`fastForward` → item flag → Alt → `hasRolls || isSpell`) is right, and moving `unsubscribe()` up to immediately after `subscribe()` is a real fix for the cancel-path leak. Only one thing actually blocks.

**Must fix before merge**

- **src/view/sheets/pages/ItemActivationCoreConfigTab.svelte:43** — the new section heading `Roll Dialog` is a hardcoded user-facing string. AGENTS.md ("Use `localize()` for all user-facing strings") and STYLE_GUIDE.md ("Always use localization for user-facing strings") both require this, and your checklist claims no hardcoded strings. Add a key (e.g. `NIMBLE.itemConfig.rollDialogSection`) next to the two you already added at `public/lang/en.json:2545` and wrap it in `localize()`. The neighbouring headings being hardcoded is pre-existing debt — new strings shouldn't add to it.

- **src/view/sheets/pages/ItemActivationCoreConfigTab.svelte:41** — a whole `<section>` + heading for one checkbox whose label already reads "Skip the roll dialog" is redundant, and it'll multiply the next time a toggle lands here. Consider folding it into the first section and renaming that heading to something generic and localized (e.g. "Activation Options"), which also lands closer to #868's grouping intent than a per-toggle section does.
- **src/managers/ItemActivationManager.ts:825** — the flag takes precedence over the Alt check, so once a spell is flagged there's no way to reach `SpellUpcastDialog` from the UI at all; the only escape is opening the sheet and unchecking. Letting Alt *force* the dialog for flagged items would make that recoverable. If you'd rather keep it unconditional, say so in the hint text at `public/lang/en.json:2546` — as written it implies Alt is only ever an equivalent, never an override.
- **src/managers/ItemActivationManager.ts:811** — `#resolveDialogData(rollOptions)` has an implicitly-`any` parameter despite the JSDoc `@param`. The shape is known (`{ domain, executeMacro, rollMode }`), so type it; `#getDefaultDialogData` at line 880 has the same gap if you're already in there.
- **src/managers/ItemActivationManager.test.ts:87** — this `vi.doMock` for `ItemActivationConfigDialog` is now dead: your comment three lines below explains that module mocks can't intercept these imports, which is exactly why you added the `testDependencies` injection. Delete it so the file doesn't look like it has two competing mocking strategies.

**What's good**

- Putting `skipRollDialog` on the shared `activation()` factory gets all four item types (including `monsterFeature`, which derives via `ReturnType<typeof activation>`) with no per-model edits, and `initial: false` on a `BooleanField` genuinely needs no migration.
- The five routing tests cover the branches that matter — flag on item, flag on spell (asserting `upcastResult` stays null, i.e. base tier), flag unset, cancel returning `{activation: null, rolls: null}`, and `fastForward` winning over the flag.

---

1. Localize the "Roll Dialog" section heading
2. Fold the single checkbox into an existing localized section
3. Let Alt force the dialog, or fix the hint wording
4. Type the `rollOptions` parameter on `#resolveDialogData`
5. Delete the now-dead `vi.doMock` for the config dialog

PR #902, opened 2026-08-05
fix(rules): make data-prep rules apply live and stop them double-applying

Fronix:
## Summary

Rules like Skill Bonus did not apply live, and a bonus written as `@strength` never applied at all. Two independent defects in the data-prep pipeline, plus a sheet-reactivity gap, all of which made a rule's effect depend on when you happened to reload.

## Changes

- **Data-prep rules no longer stack twice per prepare cycle.** `ready.ts` and the language-customization setting both re-prepared every character with a bare `prepareData()`, which reuses the already-derived `system` object and re-runs every rule hook. Rules that read-modify-write their target stacked twice: a +2 Skill Bonus read +2 while the world ran and +4 after a reload. Same for ability, save, wound, speed, hit-dice and healing-potion bonuses. Both call sites now `reset()`, and `NimbleBaseActor.prepareData` guards re-entry so a bare call from anywhere else re-initializes instead of compounding. Subclasses extend the cycle through `_onBeforePrepareData` / `_onAfterPrepareData` so an override cannot bypass the guard.
- **Stat references in rule formulas resolve.** Rule formulas evaluate against `getRollData()`, which reads ability, save and skill modifiers straight off `system` — but those were computed *after* the `prePrepareData` sweep, so every `@<ability>` / `@<skill>` reference saw the source value, `0`. The double `prepareData()` above used to hide this: the second pass read what the first had computed, so a stat-referencing bonus appeared to work after a reload and never while playing. The modifiers are now computed in `_prepareEarlyDerivedData` as well, before the sweep.
- **Actor sheets refresh on Active Effect changes.** The reactivity subscriber watched `updateActor` and item create/update/delete but nothing for Active Effects. Applying a condition re-runs data prep without touching the actor or any item, so a rule predicated on a `self:` domain tag resolved to its new value in `actor.system` while the open sheet kept rendering the old one until reopened.
- **The Rules Builder says why a rule is not running.** A rule whose source fails to build is dropped from the `RulesManager` with only a console warning, and one that builds with unresolved validation failures is force-disabled — either way its card looked normal. The reason is now recorded per rule id and rendered on the card, with nested field failures flattened in so the message names the offending field.
- **Ability and skill modifier caps removed.** Ability mods were clamped to 12 with no basis in the rules, and skill mods to 12 which is a real rule but the wrong thing to enforce silently. Neither is capped now — homebrew stays unblocked.

Predicate tags on ability keys are deliberately left late and unchanged: they must reflect post-rule values, so `LATE_PREDICATE_KEYS` stays accurate.

## Test Plan

Unit tests cover the prepare-cycle re-entry guard, the Active Effect ownership filter and the rule-failure reporting. The pipeline behaviour needs a live world.

In a v13 world with a linked PC:

1. Open the character sheet. On a feature the PC owns, open the Rules Builder and add a **Skill Bonus**: pick a skill, set the bonus to `2`. The skill's modifier on the still-open sheet changes immediately — no reload, no reopen.
2. Change the bonus to `4`. The sheet tracks it live.
3. **Reload the browser.** The modifier must be unchanged. Before this PR it doubled.
4. Set the bonus to `@strength`. The modifier must move by the character's STR mod, live and after a reload. Before this PR it did nothing.
5. Add a rule with a predicate of `self` → `concentrating`, then toggle the Concentration condition on the actor. The dependent number changes without reopening the sheet.
6. Delete the rule — the bonus clears live.
7. Give a rule a bad field (e.g. a Skill Bonus with an unknown skill via the raw-JSON editor). Its card shows a red "This rule is not running:" line naming the field.

Repeat step 1–3 for Stat Bonus, Save Bonus, Speed Bonus, Wounds and Armor Class to confirm the whole data-prep family.

## Checklist


trevlar (changes requested):
**🔧 Push back — the diagnosis and the fixes are right, but the ordering fix that headlines the PR has no test, and the new failure text bypasses `localize()`**

I traced both defects and I agree with your read. `skillBonus.prePrepareData` read-modify-writes `system.skills[x].bonus`, so a bare `prepareData()` over an already-derived `system` genuinely doubled it, and `getRollData()` at `character.ts:1030` exposes `@strength`/`@<skill>` straight off `abilities[k].mod`/`skills[k].mod` — which the old code didn't compute until *after* the `prePrepareData` sweep. Both fixes land where they should.

**Must fix before merge**

- **`src/documents/actor/character.ts:393`** — the early `_prepareAbilitySaveAndSkillModifiers()` call has no test. This is half the bug the PR is named for, and it is a pure *ordering* invariant: nothing fails if someone later moves this call back out of `_prepareEarlyDerivedData`, and the symptom (`@strength` silently resolving to 0) is invisible in CI. You already have the stub harness for this in `prepareDataAccumulatorReset.test.ts` — add a test that runs a prepare cycle with a rule whose `prePrepareData` reads `getRollData()` and asserts it sees the derived mod, not the source value. The repo standard is tests for logic changes, and this is the load-bearing one.
- **`src/managers/RulesManager.ts:68`** — `` `"${source.type}" is not a recognized rule type.` `` is rendered verbatim on the rule card next to the localized `NIMBLE.rulesBuilder.ruleNotRunning` prefix. Same at **`:40`** (`'The rule failed validation and has been disabled.'`) and **`:87`** (`'The rule source is malformed.'`). AGENTS.md: *"Use `localize()` for all user-facing strings."* Add `NIMBLE.rulesBuilder.*` keys for the three — the split-key pattern you already use for `errorSchemaLoadBefore`/`errorSchemaLoadAfter` handles the `${source.type}` interpolation. Foundry's own `toString()` output and `err.message` are upstream English and fine to pass through.

- **`src/documents/actor/base.svelte.ts:122`** — the item hooks three lines above your new effect hooks still match on `doc?.actor?.id === this.id`, which is exactly the unlinked-token cross-fire your `_ownsEffect` doc comment describes. You clearly know the hazard; switching these three to the same identity check while you're in here would close it.
- **`src/documents/actor/base.svelte.ts:320`** — `reset()` only re-derives when `game._documentsReady` is true (`ClientDocument#_initialize` gates `prepareData()` on it). A bare `prepareData()` on an already-initialized actor during `init`/`setup` would therefore reset `system` to source and never re-derive. Nothing in-system does that today, so it's theoretical — but `if (this.initialized && game._documentsReady)` makes the guard unconditionally safe for a module that does.
- **`src/documents/actor/character.ts:428`** — Under Skill Checks & Saves in the Core Rules doc it says (the max bonus a skill can ever have is +12). Also same book under Leveling Up there is a note that says "The maximum bonus a skill can have is +12." Removing the +12 skill cap is a rules change riding along in a bug-fix PR, and it silently changes the numbers of any existing character sitting at the cap.

**What's good**

- The `_onBeforePrepareData`/`_onAfterPrepareData` template split is the right call — it makes the re-entry guard structurally unbypassable rather than a convention, and it left `NimbleBaseActor` as the only `prepareData` override in the actor tree.
- `_ownsEffect` comparing by document identity instead of id, with the token-actor case called out in the comment *and* covered by a test, is exactly the level of care that class of bug needs.

---

1. Test the early ability/save/skill modifier pass.
2. Localize the three rule-failure reason strings.
3. Use identity, not id, in the item hooks.
4. Guard `reset()` on `game._documentsReady`.
5. Confirm and test the skill-cap removal.

Fronix:
Thanks — good catch on all four of the code items. Addressed in `3b3f7d8..c5bb870`. The skill cap I'm keeping, for reasons below.

**1. Test the early modifier pass** (`3b3f7d8`)

Fair, and it was the weakest part of the PR. `src/documents/actor/earlyModifierPass.test.ts` composes the real `_prepareEarlyDerivedData`, `_prepareAbilitySaveAndSkillModifiers` and `getRollData` onto a stub and asserts three things: the modifiers get computed, a rule reading `getRollData()` during `prePrepareData` sees the derived values, and the early pass runs before the sweep.

I mutation-checked it rather than trusting a green run — deleting the `_prepareAbilitySaveAndSkillModifiers()` call from `_prepareEarlyDerivedData` fails it with `expected +0 to be 5`, which is precisely the reported symptom. The Actor test double needed `prepareDerivedData` and `getRollData` added; both exist on core `Actor`.

**2. Localize the failure reasons** (`def0c18`)

Right, and there was a second problem underneath: the manager is constructed during data preparation, which can run before `i18nInit`, where `game.i18n.localize` hands back the raw key. So localizing at the call site would have been wrong too.

Failures are now stored unresolved as `{key, data}` and localized in `failureFor()`, which the Rules Builder only calls at render time. Three new keys under `NIMBLE.rulesBuilder.*`; `{type}` interpolates through `game.i18n.format`, so no split-key needed. Foundry's validation output and `err.message` still pass through as upstream English, as you suggested.

Verified live — the rendered banners read `This rule is not running: "notARuleType" is not a recognized rule type.` and `This rule is not running: SkillBonusRule validation errors: skills: 0: notaskill is not a valid choice`.

**3. Identity in the item hooks** (`74c9207`)

You're right that I should have carried it across in the same pass. Both now go through one helper, renamed `_ownsEffect` → `_ownsEmbeddedDocument` since walking the parent chain covers an owned item, an effect on the actor and an effect on one of its items alike.

This is the highest-risk change in the set — it's the path that makes sheets update at all — so I re-ran the full live regression through real Rules Builder clicks: all 10 checks pass, including live edit, no-double-on-reload, and the condition-gated refresh.

**4. `_documentsReady` guard** (`c5bb870`)

Confirmed against `client-document.mjs:59` and guarded, with a test that flips the flag. Agreed it's theoretical today, but it costs one condition to make the guard unconditionally safe.

**5. The skill cap — keeping it removed**

You're factually right that +12 is in the book, and the code comment says so. The system should not silently clamp an authored number. A homebrewer who writes a +15 and gets a +12 with no explanation hits exactly the "why can't I do X?" question we don't want to generate. It's consistent with how this system treats rules generally — report state, don't referee it — and hard gates stay reserved for action limits ("1/round"), not arithmetic ceilings.

If anyone previously has gone over the cap but were "saved" by the silent cap will have to just adjust their characters in my opinion. It's better we don't cap things unless it makes sense, even if the rules say it is, it's better that each user is left to handle that for themselves.

PR #903, opened 2026-08-05
chore: add a release note generator claude command

trevlar:
Ports the release-note generator command onto `dev`.

`.claude/commands/nimble-release-notes.md` was merged in #869, but that landed on `main` only — `dev` never received it. This branch is `dev` plus that single commit, and the file content is byte-identical to the copy already on `main` (verified with `git diff origin/main:… HEAD:…`).

Split out of the ancestry size-options branch, which had the commit sitting in its history and was carrying it into an unrelated review.

No code changes: one markdown command file, nothing imported or executed by the system build.


Fronix (approved):
Ah, my bad. Forgot about the dev switch ^^

PR #907, opened 2026-08-06
fix(levelUp): stop duplicate feature copies from crashing the level-up dialog

### LINKED ISSUE #906: Cannot level up Shepherd from 2 -> 3
### Foundry VTT Version

V13 Stable 9 Build 351

### Nimble System Version

0.8.8 latest

### Description

I'm unable to level up a shepherd from level 2 to 3. 

Image

The console returns the error circled in the above screenshot when I attempt.

It will not register clicking on anything in the level up popup while this is happening apart from the button to close it.

This is only happening with Shepherd from what I can see. I've done the same with Hunter, Cheat and Berserker with no problems. 

### Steps to Reproduce

Create a new player character.
Give them the Shepherd class.
Attempt to level them up to 3.

### Expected Behavior

I expect the flow to work as intended. 
- Click the button to begin levelling up.
- Make the selections presented in the popup.
- Save the changes.

### Actual Behavior

- Click the button to begin levelling up.
- I cannot click on any of the selections presented in the popup.
- I cannot save the changes.
- Unable to level up.

### Additional Context

Image

Image

Fronix:
## Summary

When two copies of the same class feature are visible to the level-up dialog (for example a world item alongside its compendium original), their identical grantSpells rules produced spell/school selection groups with duplicate ruleIds. The dialog renders those groups in keyed each blocks, so the duplicate key threw svelte's `each_key_duplicate` and froze the whole dialog. This is what made leveling a Shepherd from 2 to 3 impossible in #906: Master of Twilight is the only feature at that level that emits keyed selection groups, which is why other classes appeared unaffected.

`collectSpellGrants` now collapses repeated selection keys into their first occurrence, so no data condition can crash the dialog. Same-named or source-linked duplicates are already presented as a "keep one or all" comparison by #854; this covers copies that clustering cannot match (renamed or unlinked hand copies) and acts as a safety net for the general case.

## Changes

- `collectSpellGrants` tracks emitted selection keys and skips repeats for both `selectSchool` and `selectSpell` groups
- Added unit tests covering duplicated rules arrays for both selection modes

## Test Plan

1. Import "Master of Twilight" from the class features compendium into the world Items directory, then rename the copy (renaming breaks the duplicate-source clustering, simulating a hand-made copy)
2. Create a character with the Shepherd class and level it to 2
3. Click Level Up on the sheet
4. Before this fix: the dialog freezes with `Uncaught Error: https://svelte.dev/e/each_key_duplicate` in the console. After: the dialog renders fully, each utility spell choice appears once, and the level-up completes
5. `pnpm vitest run src/view/dialogs/spellGrantUtils.test.ts`

Verified live on Foundry v13 (build 351) via scripted real-UI clicks for both the renamed-copy case (crashed before, works after) and the same-name case (unchanged, still shows the #854 comparison UI).

## Checklist

## Related Issues

Fixes #906

### DIFF
--- a/src/view/dialogs/spellGrantUtils.test.ts
+++ b/src/view/dialogs/spellGrantUtils.test.ts
@@ -629,4 +629,63 @@ describe('collectSpellGrants', () => {
 			expect(result.spellSelections[0].ruleId).toBe('spell-choice-1-ice');
 		});
 	});
+
+	describe('duplicate rule ids', () => {
+		it('collapses selectSpell groups from duplicate copies of the same rules', () => {
+			const index = buildTestIndex();
+			const rules: RulesArray = [
+				{
+					id: 'spell-choice-1',
+					type: 'grantSpells',
+					schools: ['fire', 'ice'],
+					tiers: [0],
+					mode: 'selectSpell',
+					count: 1,
+					predicate: { level: { min: 4 } },
+				},
+			];
+
+			// The same feature present twice (e.g. a world copy alongside the pack
+			// original) supplies its rules array twice
+			const result = collectSpellGrants(
+				[rules, [...rules]],
+				index,
+				'mage',
+				4,
+				new Set(),
+				new Set(),
+			);
+
+			const ruleIds = result.spellSelections.map((s) => s.ruleId);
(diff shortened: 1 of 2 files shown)


trevlar (approved):
LGTM

PR #908, opened 2026-08-07
feat(backgrounds): automate Haunted Past fear advantage

### LINKED ISSUE #402: [Automation] Background - Haunted Past: savingThrowRollMode for fear
## Context
Background **Haunted Past** has no rules in its pack JSON.
Pack file: `packs/backgrounds/core/haunted-past.json`

## Rules from Rulebook
- "Advantage against fear."

Note: The rulebook says "advantage against fear" broadly, not specifically "saving throws vs fear." This likely applies to fear-related saves but could be interpreted more broadly.

## Implementation
Add `savingThrowRollMode` rule: advantage on fear-related saves.

## Rule Types
`savingThrowRollMode` (exists)

Part of #365

trevlar:
Closes #402

## What

Haunted Past grants "Advantage against fear" but shipped with `"rules": []`, so it had no mechanical effect. Automating it needed a rule type the system did not have, so this PR adds one — `situationalRollMode` — and uses it for the background.

## Why a new rule type

The existing roll mode rules (`savingThrowRollMode`, `skillRollMode`, `initiativeRollMode`) all bake their value into the actor's stored default roll mode, so they apply to *every* roll of that kind. Encoding "advantage against fear" that way would grant +1 on all WIL saves, which is broader than the background text — the system has no way to know what a given save is against.

`situationalRollMode` is offered rather than applied. It stores nothing on the actor; the check roll dialog lists every offered adjustment whose predicate passes and whose check type and target key match the roll being configured, and folds the checked ones into that roll only. Rolling a WIL save on a Haunted Past character shows an "Against fear" checkbox below the roll mode slider; leave it unchecked and the save is unchanged.

The rule's `predicate` still gates whether the option is offered at all, so state the system *does* know about (bloodied, unarmored, raging) belongs there rather than in the label. The two compose. Attack rolls are not covered — they go through the item activation dialog, where `conditionalBonus` already offers a per-attack choice.

## Changes

| Area | |
|---|---|
| `src/models/rules/situationalRollMode.ts` | the new rule: `value`, `checkType`, and a target list per check type (`saves` / `abilities` / `skills`, each accepting an `all` sentinel) |
| `src/view/dialogs/CheckRollDialog.utils.ts` | `getSituationalRollModeOptions` — the discovery helper the dialog uses to list the options for a given roll |
| `src/view/dialogs/CheckRollDialog.svelte` | renders one checkbox per offered option and folds the checked values into the roll's roll mode |
| `packs/backgrounds/core/haunted-past.json` | the rule (id `N43aM4mMNGpq5WN6`), with an `[A]` annotation following `fearless.json` |
| `src/migration/migrations/Migration044HauntedPastFearAdvantage.ts` | backfill for existing characters and world-level copies |
| `src/migration/compendiumSourceId.ts` | extracts `toSnapshotId`, which had been copied verbatim into migrations 035, 041 and now 044 — Rule of Three |
| `src/config.ts`, `src/config/registerRulesConfig.ts`, `public/lang/en.json` | rule registration, the `rollCheckTypes` list, and the English strings |
| `docs/system/rules.md`, `docs/documentation/reference/_partials/situationalRollMode.md` | when to reach for a default roll mode rule vs. a situational one, and the worked Haunted Past example |
| `.github/workflows/checks.yml` | the migration-count gate globbed `Migration*.ts` and counted a co-located `.test.ts` as a migration |

## Why a migration is needed

The check roll dialog reads the rules on the actor's own items, so a pack edit alone never reaches a character who already has the background embedded. `Migration044` adds the rule to the embedded background on each character, and to any world-level copy so characters built from it later carry it too — the same two places `Migration022CelestialSavingThrow` handles.

Nothing persisted is written. Unlike `savingThrowRollMode`, this rule does not feed `savingThrows.will.defaultRollMode`, so the stored roll mode is left exactly as the player configured it. A background already carrying an equivalent WIL rule — a GM who automated this themselves — is left alone.

## Tests

- **`situationalRollMode.test.ts`** (19) — schema shape, the `showWhen` pairing of check type to target list, the `all` sentinel, zero-value rules offering nothing, and `matchesRoll` across all four check types.
- **`CheckRollDialog.utils.test.ts`** (15) — target-key mapping per check type, the filter chain, the label fallback chain, and item-scoped option keys (rule ids are only unique within an item).
- **`CheckRollDialog.svelte.test.ts`** (7) — the dialog wiring: checking an option folds into the rendered formula *and* the submitted roll mode, unchecking drops it, multiple options sum, and the total clamps at the slider ceiling. Verified by mutation — submitting `selectedRollMode` instead of `effectiveRollMode`, building the formula from the wrong one, or dropping the clamp each fail a test.
- **`CharacterCreationDialog.test.ts`** — assertions driven from the real `haunted-past.json`, so a typo in `checkType`/`saves`/`value` fails a test instead of silently shipping a background that does nothing, plus a check that character creation leaves the default WIL roll mode at 0.

All gates green: 248 files / 3830 tests, Biome, dependency-cruiser, typecheck.

## Not included, deliberately

**Babele skeletons.** `docs/system/compendium-translation.md` says to regenerate after editing English pack sources, but `AGENTS.md` forbids agents generating or modifying non-English locale files, and running the regenerator produced ~700 lines of unrelated drift. Repo practice agrees — only #821 has ever touched `public/lang/babele/`, across 18+ pack commits since. For a maintainer who wants it now, the es/fr `nimble.nimble-backgrounds.json` entries each need:

```json
"rules": { "N43aM4mMNGpq5WN6": { "label": "Against fear" } }
```

**`[A]` vs `[M]` on the description.** Kept `[A]`. The convention is undocumented and already self-inconsistent elsewhere (`stout.json` marks hook-driven rules `[M]`). The rule is genuinely automated — the option appears on its own and applies without any manual arithmetic — and marking it `[M]` would invite players to apply it a second time by hand. Flagging the inconsistency rather than silently spreading it.


src/migration/migrations/Migration044HauntedPastFearAdvantage.test.ts
Fronix, line 1: Should be removed, migration tests are only useful for one run.

Fronix (dismissed):

Fronix (dismissed):
I reviewed this branch against `origin/dev`. The design is correct.

Issue #402 recommends the `savingThrowRollMode` rule. That rule writes to `savingThrows.will.defaultRollMode`. It then gives advantage on all WIL saves. The new `situationalRollMode` rule is the correct choice.

Correct these two problems before you merge.

**Migration045 has no effect.**

Migration044 already writes the icon. Migration045 stops if the icon has a value, so it never changes data. Delete it and its `index.ts` entry. Set `LATEST_SCHEMA_VERSION` to 44.

**The pack description shows an incorrect value.**

The description text is `+1 WIL save roll mode, offered when the save is against fear [A]`. Other backgrounds use the format `+N <term>` for a flat modifier. Three examples are `+1 Initiative [A]`, `+3 Max Wounds [A]`, and `+1 Naturecraft [A]`. The Core Rules give a different definition of advantage. The player rolls more dice and removes the lowest dice. The description also uses internal terms.

Replace the description text with `Advantage against fear [A]`.

These two problems are less important:

- Set the slider to 6, then select an option. The option adds no value to the roll. But the row shows the highlight and the text "Advantage × 1".
- The `registerRulesConfig.ts` file puts `situationalRollMode` after `skillBonus` in two maps. The lines are 89 and 137. The `en.json` file uses the correct sequence.

The `saves: ["will"]` value agrees with the rulebooks. All fear effects that use a save use a WIL save. Six examples are Fearsome, Horrifying Wail, Terrible Roar, Savage Screech, Cruelty's Edge, and Frightening. The other fear effects give no roll to the target. The `[A]` annotation is correct. Only the description text is incorrect.

One comment about the tests. The `CheckRollDialog.utils.test.ts` file replaces all rule methods with stubs. The `toHaveBeenCalledWith('savingThrow', 'will')` assertion shows only that the code sends these arguments. It does not show that the code filters the options. The `matchesRoll` stub always returns true. The `CheckRollDialog.svelte.test.ts` file gives the real coverage.

src/migration/migrations/Migration044HauntedPastFearAdvantage.test.ts
trevlar: Done in `bc81c1cca` (before the rebase onto `dev`) - the file is deleted. The migration itself is now `Migration045HauntedPastFearAdvantage`, since dev's merged #927 took 044.

trevlar:
## Fixed

- **Dead icon migration** — `Migration045HauntedPastFearIcon` and its `index.ts` entry are gone. Your diagnosis was right: the rule literal it followed already carried the icon, so the `if (rule.icon) continue` guard skipped every copy.
- **Font Awesome in game content** — `situationalRollMode` no longer has an `icon` field at all. The dialog renders the granting item's own `img`, so Haunted Past shows the background's spirit art (`src/models/rules/situationalRollMode.ts:135`, `src/view/dialogs/CheckRollDialog.svelte:106`).
- **Pack description used internal terms** — now `Advantage against fear [A]` (`packs/backgrounds/core/haunted-past.json`).
- **Option at the slider ceiling claimed a bonus it never applied** — a selected row now reports the adjustment actually applied and drops to a neutral accent when that is zero; unselected rows still preview what they offer (`src/view/dialogs/CheckRollDialog.svelte:75`).
- **Rule ordering disagreed with `en.json`** — `situationalRollMode` now precedes `skillBonus` in both maps (`src/config/registerRulesConfig.ts:88`, `:136`).
- **One-shot migration test** — deleted (`bc81c1cca`).

## Not fixed

- **`LATEST_SCHEMA_VERSION` is 45, not 44** — done differently on purpose. The branch is rebased onto current `dev`, where merged #927 occupies 044 with `Migration044MaxHpBonusDerived`, so the fear-advantage migration is 045. 45 migration files against version 45, so the CI count check holds.
- **Weak stubs in `CheckRollDialog.utils.test.ts`** — left as is. You are right that the `toHaveBeenCalledWith` assertions only prove argument passing, but the `filtering` block already covers exclusion directly for wrong type, zero adjustment, non-matching roll, and failing predicate.

## Approach

Removing the `icon` field is what removed the migration: with no field to stamp, the second migration had nothing left to do. Using the item's own image also means new situational rules need no icon authored and none to keep in sync.

Rather than blocking the checkbox at the slider's ceiling, the row now tells the truth about what it contributed. The roller can still record that the situation applied even when the roll mode cannot go higher.

Pushed as `cf29b76d8`.

Fronix (approved):
LGTM

PR #909, opened 2026-08-07
feat(saves): grant Survivalist advantage against poison saves (#401)

### LINKED ISSUE #401: [Automation] Background - Survivalist: add savingThrowRollMode for poison
## Context
Background **Survivalist** has a `maxHitDice` rule (+1 max Hit Die) but is missing the poison save advantage.
Pack file: `packs/backgrounds/core/survivalist.json`

## Rules from Rulebook
- "Advantage against poison saves. +1 max Hit Die."

## Implementation
Add `savingThrowRollMode` rule: advantage on poison saves.

## Rule Types
`savingThrowRollMode` (exists)

Part of #365

trevlar:
Closes #401

## What this adds

Survivalist's "Advantage against poison saves" now works, using the `situationalRollMode` rule that #908 built for Haunted Past's "Advantage against fear".

```json
{
  "type": "situationalRollMode",
  "label": "Against poison",
  "checkType": "savingThrow",
  "saves": ["strength"],
  "value": 1
}
```

Poison is a STR save, so the rule names `strength` exactly as Haunted Past names `will`. It is *situational* rather than a stored default because STR also covers forced movement, restraint and extreme temperatures, none of which Survivalist helps with. A `savingThrowRollMode` with `target: "strength"` would hand out advantage on all four.

`Migration047SurvivalistPoisonSave` mirrors `Migration045HauntedPastFearAdvantage`: it backfills embedded and world-level copies, matches on compendium source id with a name fallback, and matches by rule shape as well as id so a GM's hand-authored equivalent is not duplicated. `LATEST_SCHEMA_VERSION` → 47.

## Why this PR changed shape

This branch predates #908 and originally built a parallel path for the same concept: a `situation` free-text field on `savingThrowRollMode`, a collector, and a "Situational" reminder section on the saving throw config dialog. #908 then landed a dedicated `situationalRollMode` rule type with a richer roll dialog — predicates, check-type and target matching, per-roll toggles that move the slider and clamp correctly.

Two rule types for one concept is one too many, so the branch now uses theirs. Removed here:

- `savingThrowRollMode.situation` and its schema tests
- `situationalSaveRules.ts` and its collector tests
- the Situational section on the saving throw config dialog
- the `situation` guards in `resolveSavingThrowRollModes`, `ancestryBonusRequiresSaveChoice` and `StatAssignment.svelte`
- the localization behind all of it

The check roll dialog is otherwise unchanged from #908 and is now the only surface these appear on.

## Also in here

**`calculateDefaultRollModes` extracted** from `ActorSavingThrowConfigDialog.svelte` into `ActorSavingThrowConfigDialog.utils.ts` (net −114 lines in the component), with 17 tests. The component previously had no test file, so the "Reset to Class Defaults" calculation was untested; it now covers class defaults, `set` vs `adjust`, the `all` / `advantaged` / `disadvantaged` / `neutral` targets, priority ordering, and choice-based rules.

**Square option icons.** The situational rows in the roll dialog rendered their artwork in a round accent ring; they now match the feature cards on the character sheet's Features tab — square, unringed, dark-backed — so the same item reads the same in both places. Direction is still carried by the row tint and the value pill. This affects Haunted Past's row too.

## Testing

- `CharacterCreationDialog.test.ts` asserts the Survivalist pack data ships the expected rule, and that a Survivalist character's stored STR default stays at the class value rather than absorbing the advantage. Both sit beside the equivalent Haunted Past tests.
- `ActorSavingThrowConfigDialog.utils.test.ts` pins that a `situationalRollMode` rule never reaches the stored-default calculation.

`pnpm check` is green — Biome format, Svelte + Biome lint, dependency-cruiser, `tsc --noEmit`, vitest: **260 files, 4029 tests passing**. The single lint "info" is pre-existing in an untouched `NimbleBaseItem` test helper.

## Deliberately out of scope

- **Survivalist's `+1 max Hit Die [M]`** is tagged manual despite having a working `maxHitDice` rule. Pre-existing and unrelated; left alone. The new poison line is tagged `[A]`.


Fronix (commented):
Reviewed against `origin/dev` @ 94063654. tsc, biome format/lint, depcruise, 260 test files and `vite build` all clean locally. Not verified in a live Foundry session.

Using `situationalRollMode` instead of the `savingThrowRollMode` #401 asked for is the right call: `CoreRules-2.md:232` has STR covering forced movement, restraint, poison and temperature, so a stored roll mode would over-grant on three of four.

## Fix before merge

### Docs partial names the wrong item and label

`docs/documentation/reference/_partials/savingThrowRollMode.md:1-3`

`**Example: Celestial.**` with **Label** → `Celestial`. The rule ships on `packs/ancestryBonuses/core/exotic/highborn.json` with label `Highborn`, on an `ancestryBonus`. The Celestial ancestry's `rules` array holds only `grantProficiency`. `CoreRules-2.md:788` names the feature "Highborn. Your disadvantaged save is Neutral instead."

Use `**Example: Highborn.** The Celestial ancestry bonus that cancels...` with **Label** → `Highborn`. The field values (0 / `disadvantaged` / `set`) are correct.

### Paired em-dash renders broken

`docs/documentation/reference/_partials/savingThrowRollMode.md:10`

`generateReference.gen.ts:46` runs `body.replace(/ — /g, ': ')`. After `pnpm docs:generate`, `rules-bonuses.md:236` ships as:

> For an advantage that only applies in a circumstance the system cannot detect: "against poison", "against fear": use `situationalRollMode` instead

A trailing dash survives the transform (`diceConsumer.md`); a parenthetical pair does not. Use commas or parentheses.

### Migration047 silently skips characters

`Migration047SurvivalistPoisonSave.ts:76-84`

Any `situationalRollMode` with `checkType: 'savingThrow'` and `strength` in `saves` counts as already handled. Safe for Migration045 because WIL/fear is a tight pairing; STR is not. A GM who hand-authored "Against restraint" or "Against extreme cold" on Survivalist gets no poison rule and no signal. Add `label` or `value` to the signature, or match on id only.

## Should fix

### `resolveTargetSaves` duplicated verbatim

`ActorSavingThrowConfigDialog.utils.ts:49` and `CharacterCreationDialog.svelte.ts:27`

Same switch, same `-3..3` clamp, near-identical rule type. The JSDoc asserts the two stay semantically identical, with nothing testing it. On drift, "Reset to Class Defaults" and character creation write different stored defaults for the same character. Commit 2 is titled "extract roll-mode calculators, share pack source-id fold".

### `formatRollModeLabel.ts` promoted with one caller

`AGENTS.md` Code Promotion Rules: local, same file, only one caller in this file; never skip a level. There is one call site. The `localize()` fix works in-component. `RollModeConfig.svelte:4` still defines a second function of the same name one directory away.

## Nits

`CheckRollDialog.svelte:254` hardcodes `rgba(0, 0, 0, 0.7)` against STYLE_GUIDE §"Use Semantic Color Variables". Fourth copy in the tree, and the six-declaration block is verbatim `.nimble-feature-card__img`.

`src/config.ts:120-123` comments on keys that are not there. Filing them as `NIMBLE.rollMode.*` rather than `NIMBLE.saveConfig.*` removes the need.

`collectRollModeRules`'s `if (!item.rules) continue`, and the test "ignores rules of other types and items carrying none", cover states production data cannot reach (`AGENTS.md` YAGNI). The `target ?? 'all'` and `mode` branch flip are not in this bucket: they match the schema `initial` values and fix a real divergence from `resolveSavingThrowRollModes`.

trevlar:
## Fixed

- **Rule reference named the wrong item** - the worked example is now `**Example: Highborn.**`, the Celestial ancestry bonus, with the label matching the rule the pack ships (`docs/documentation/reference/_partials/savingThrowRollMode.md:1`).
- **Paired dashes broke the generated page** - the parenthetical is now a parenthesis pair, so `rules-bonuses.md:232` reads correctly after `pnpm docs:generate` (`docs/documentation/reference/_partials/savingThrowRollMode.md:10`).
- **Migration047 skipped hand-authored STR rules** - the already-present check now also matches the rule label, so "Against restraint" or "Against extreme cold" no longer counts as the poison rule and the character still gets it (`src/migration/migrations/Migration047SurvivalistPoisonSave.ts:88`).
- **`resolveTargetSaves` duplicated** - both entry points now call one `calculateSavingThrowRollModes`, so the clamp, the `adjust` branch and the category targets exist once (`src/utils/calculateSavingThrowRollModes.ts:66`, `src/view/dialogs/ActorSavingThrowConfigDialog.utils.ts:19`, `src/documents/dialogs/CharacterCreationDialog.svelte.ts:19`).
- **`formatRollModeLabel` promoted with one caller** - demoted back into `ActorSavingThrowConfigDialog.svelte:12` and the module deleted.
- **Comment on absent config keys** - the labels moved to their own `NIMBLE.rollMode.*` namespace, so nothing is missing from `saveConfig` and the comment is gone (`public/lang/en.json:254`, `src/config.ts:111`).
- **Hardcoded artwork backing** - new `--nimble-artwork-backdrop` token, applied to all four copies (`src/scss/base/_variables.scss:65`).
- **Unreachable branch and test** - the `!item.rules` guard is gone and the test now covers only other rule types, which production data does reach (`src/view/dialogs/ActorSavingThrowConfigDialog.utils.ts:23`).

## Not fixed

- **The six-declaration icon block shared with `.nimble-feature-card__img`** - two copies, under the Rule of Three, so only the color is shared.

## Approach

The shared calculator lives in `src/utils/` because its two callers sit in different features. The creation dialog holds the ancestry save choice on the form rather than on the rule, so it stamps `selectedSave` on before handing rules over; everything else is identical between the two. Its semantics tests moved with it to `src/utils/calculateSavingThrowRollModes.test.ts`, leaving the dialog utils test covering what is actually its own: folding rules off items and ignoring the types it should not read.

Pushed as d2d89035f.

Fronix (approved):

PR #910, opened 2026-08-07
feat(spells): roll Shadow Trap's damage from the chat card (#370)

### LINKED ISSUE #370: [spells] Shadowtrap
Show a roll damage button when added to chat.

trevlar:
Closes #370.

## What

Shadow Trap's 3d12 lands on *the next creature to move adjacent*, which the activation cannot know about, so the spell shipped with no damage node at all rather than putting a number on the card for something that had not happened yet. Its card now carries a **Roll Damage** button.

## How

Damage nodes gain a `deferredRoll` flag rather than getting a new node type, so everything downstream — Apply Damage, resistances, reductions, tooltips — keeps working unchanged.

- `ItemActivationManager` skips deferred nodes: they take no roll, and they do not claim the primary damage slot the first *rolled* node is owed.
- `findNodesByContexts` surfaces a deferred node itself, since it has no `damageOutcome` child to stand in for it.
- Pressing Roll Damage evaluates the formula and writes it back onto the node; Apply Damage then picks it up as an ordinary packet. Until then `canApplyAllDamage()` already reports nothing to apply, so that button stays disabled on its own.
- The item sheet gets a `Roll Damage From Card` checkbox on top-level damage nodes, so this is authorable rather than JSON-only.

A `DamageRoll`, so the node's `canCrit` and `canMiss` mean something: Shadow Trap is Single Target, so the Core Rules give it a primary die that misses on a 1 and crits on a 12, and only `DamageRoll` reads those two flags. The card restates its own `isCritical` and `isMiss` from the result, because the activation posted before there was a roll to read them from. The roll runs on the clicking client rather than being routed to the primary GM the way the incoming-reaction resolvers are, because a chat message is updatable by its author or a GM and the button is offered to nobody else. Players who can't write to the card see nothing until it is rolled — a `0 Necrotic` placeholder would read as damage that already resolved.

**Migration 045** adds the node to copies of the spell already on actors, in world items, or in duplicated world compendiums, skipping any that already carry damage so a GM's own automation survives.

## Tests

- `buildDeferredDamagePatch` — attaches the roll, appends to `rolls`, refuses an already-rolled node / a non-deferred node / an unknown id.
- `findNodesByContexts` — deferred nodes surface on their own; ordinary root damage still surfaces only through its outcome child. Verified red before the change, green after.
- `ItemActivationManager` — a deferred node produces no roll, and a following damage node still claims the primary slot. Verified red before the change, green after.
- `rollDeferredDamage` — writes the roll onto the node, appends to `rolls`, lets a GM roll a card they did not author, refuses a bystander, refuses a second click, leaves the rest of the card alone.
- `applyAllDamage` — nothing to apply while deferred, applies 21 once rolled.

## Gates

`pnpm check` green: Biome format (no fixes), Svelte + Biome lint (1 pre-existing info in an unrelated test file), dependency-cruiser clean, `tsc --noEmit` clean, 3572 tests passing. Migration count 45 == `LATEST_SCHEMA_VERSION`. `pnpm build:compendia` builds all 14 packs.

## Notes for review

- **English strings added** (`NIMBLE.chat.rollDamage`, `NIMBLE.activationEffects.deferredRoll`) — flagging per AGENTS.md. No other locale files touched.
- **Deliberately out of scope:** Shadow Trap's upcast (*"+1 size category, +1d12 damage when they escape"*) is not automated, and cannot be: the engine has no operation for advancing a size category and no event for a creature escaping a condition. The wording is clear enough, there is simply nothing to hang it on, so it stays as description text.


Fronix (dismissed):
Correct these three problems before you merge this branch.

**The spell must be able to crit and to miss.**

The pack file sets `canCrit` to false and `canMiss` to false. The Core Rules do not permit this.

> Any spell or ability that can harm an enemy counts as an attack. [...] if you roll a 1, you miss, and the attack has no effect. For attacks with multiple dice, the leftmost die is called the Primary Die.

> Exploding Critical Hits. Rolling the max on a Primary Die is a critical hit (crit). [...] Crits also ignore monster armor.

There is one exception, and Shadow Trap does not meet it.

> Multi-Target/Area of Effect (AoE). Multi-target/AoE attacks follow special rules to streamline play: they do not miss or crit, and a single roll is applied to all targets.

Shadow Trap is a Single Target spell. Unspeakable Word shows that a Single Target spell must state the exception in its own entry.

> REACH: 8. DAMAGE: d66 (with advantage, ignoring armor, does not miss or crit) on a failed INT save.

The pack files follow this rule. I counted the root damage nodes in `packs/spells/`. 15 nodes permit a crit and a miss. 14 nodes prevent both. One node permits only a crit. The rulebook gives the AoE tag to 13 of these 14 spells. Only Condemn and Shadow Trap are Single Target spells.

You cannot correct this with the two flags. Only the `DamageRoll` class reads `canCrit` and `canMiss`. Refer to `src/dice/DamageRoll.ts:233`. A plain `Roll` object cannot hold these two values. Migration044 also copies the two flags to all existing worlds.

Use a `DamageRoll` object. If the current values are correct, quote the rule that permits them.

**The card shows an empty block of roll details.**

`DamageRoll.svelte:56` calculates `showRollDetails` from `rollOptions.primaryDieValue != '0'`. A plain `Roll` object writes `options: {}`. The two comparisons are then true. `RollSummary.svelte:44` draws the block, but the two values are empty. The player clicks Roll Damage and sees this text:

```
Primary Die Value:
Primary Die Modifier:
```

Change the condition. Test if the two values exist.

**The homebrew page is now incorrect.**

The user documentation must match the code. `docs/documentation/homebrew/activations.md:41` contains the words "Two checkboxes refine it: Ignore Armor and Only Damage Hostile Actors". This branch adds a third checkbox. Update this page.

These four problems are less important:

- Dice So Nice does not show the dice. Dice So Nice uses the `createChatMessage` hook. This code writes the roll with `update()`. Call `game.dice3d?.showForRoll()` after the update.
- `types/effectTree.d.ts:70` contains the name Shadow Trap. A comment in the engine must describe only the mechanism. Three test files contain the same name.
- The comment at `chatMessage.ts:1011` tells you that two clicks add two rolls to the array. This is incorrect. The `update()` method replaces the full `rolls` array. The second update deletes the first roll.
- Two commits change Commander pack files. Move these two commits to a different branch.

Your description tells us that the upcast text is not clear. I do not agree. The Core Rules use a condition for Shadow Trap and no condition for Greater Shadow.

> Shadow Trap. UPCAST: +1 size category, +1d12 damage when they escape.

> Greater Shadow. UPCAST: +1d12 damage, +1 shadow minion on explosion.

The engine cannot do this. The engine has no `addSizeCategory` operation. The engine has no event for an escape. Do not automate the upcast. Change only this explanation in the description.

trevlar:
## Fixed

- **Spell could neither crit nor miss** — both flags flipped to `true` in `packs/spells/core/necrotic/shadow-trap.json` and in the migration, and the card now builds a `DamageRoll` rather than a plain `Roll`, so they are actually read (`chatMessage.ts:1034`). It forwards the node's `canCrit`/`canMiss` and defaults both to `true` when the node states neither.
- **Card outcome never reflected the deferred roll** — the activation posted before there was a roll, so `system.isCritical`/`isMiss` stayed false forever. `rollDeferredDamage` restates both from the result, coerced to booleans since `DamageRoll` leaves them undefined when a flag is off (`chatMessage.ts:1058`).
- **Empty block of roll details** — the condition tests that the two values are present and non-zero instead of comparing against the string `'0'` (`DamageRoll.svelte:37`).
- **Homebrew page listed two checkboxes** — now three, with a paragraph on what Roll Damage From Card does and the note that Can Miss and Can Crit still apply (`docs/documentation/homebrew/activations.md:41`).
- **Dice So Nice showed no dice** — `showForRoll` is called after the update, carrying the message's whisper/blind/speaker so a hidden roll stays hidden (`src/utils/showDiceAnimation.ts`).
- **Spell name in engine code** — dropped from the `deferredRoll` doc comment and from the three test files; each describes the mechanism instead (`types/effectTree.d.ts:70`).
- **Wrong comment about two clicks appending two rolls** — `update()` does rewrite `rolls` wholesale. The comment says that now, and the patch routes through `replaceDamageRollInRollsSource` so the roll replaces the card's damage roll instead of stacking beside it (`buildDeferredDamagePatch.ts:66`).
- **Commander pack files** — that commit is off this branch; it is parked on `chore/commander-pack-trailing-newlines` locally and will go up separately.
- **Upcast explanation in the description** — rewritten. You are right that the rulebook is not ambiguous; the blocker is that the engine has no operation for advancing a size category and no event for an escape, so there is nothing to hang it on.

## Not fixed

- **Actor-level crit suppression on the deferred roll** — minion/flunky/AoE/proficiency suppression lives in `ItemActivationManager` and reads the item, which the card does not hold. Duplicating it here would be a second place for the rule to drift. No shipped spell with a deferred node is affected today; if a monster feature ever gets one, the resolution belongs in a shared helper both callers use.

## Approach

Making it a `DamageRoll` has one deliberate side effect worth checking: `findPrimaryDamageNode` matches on `roll.class === 'DamageRoll'`, so once the trap is rolled it becomes the card's primary damage node and a Death Blow spend folds into it. That reads as correct for a spell whose only damage is the trap, and an unrolled node still does not match.

Crit and miss detection happens in `DamageRoll._evaluate`, which the shared test Roll mock never runs, so the outcome tests state the outcome via a prototype spy. What is asserted from the real class is the work that does run under the mock: `3d12` coming back as `1d12x + 2d12` proves the roll treated the damage as an attack with a primary die.

Pushed as `fa289917b`. `pnpm check` green: format clean, lint clean (the one info is the pre-existing `useLiteralKeys` in `shouldSuppressActivationCard.test.ts`), dependency-cruiser clean, `tsc --noEmit` clean, 3845 tests passing.

Fronix (approved):
Some merge conflicts but otherwise LGTM

PR #911, opened 2026-08-07
feat(conditions): let GMs add custom conditions from world settings

### LINKED ISSUE #484: Ability to add custom/new statuses
Useful for Patreon content released with new statuses.

trevlar:
## Summary

Lets GMs define their own conditions from world settings, so statuses from content the system doesn't ship with (e.g. Patreon releases) can be used without waiting on a system update.

A new GM-restricted **Custom Conditions** submenu lives in Settings → System → *Manage Conditions*. Each row is a name, an auto-derived snake_case id, a description, and a file-picker icon. Saved conditions merge into `CONFIG.NIMBLE`'s condition dictionaries and then show up everywhere built-ins do: the Token HUD, the actor conditions tabs, the canvas conditions panel, the `[[/condition]]` enricher, and the `applyCondition` / `conditionImmunity` rule dropdowns.

This mirrors the existing custom-spell-schools feature, since it's the same problem shape — a world-scoped `Array` setting, a snapshot of the built-ins so re-merging is idempotent, a validating reader, and in-place mutation of the config dictionaries so held references stay live.

## Changes

- **`src/settings/customConditionSettings.ts`** (new) — data layer: setting key, `CustomCondition` shape, id sanitizer, validating reader that drops malformed / duplicate / built-in-colliding entries, and an idempotent merge into `conditions` / `conditionDescriptions` / `conditionDefaultImages`.
- **`src/settings/registerCustomConditionSettings.ts`** (new) — registers the setting plus a GM-restricted `registerMenu`. Its `onChange` re-merges, re-runs `ConditionManager.initialize()` / `configureStatusEffects()`, and fires a `nimble.conditionsChanged` hook that already-rendered condition lists listen for, so the token HUD and open conditions panels land edits without a reload.
- **`src/view/settings/CustomConditionsEditor.svelte`** (+ `.types.ts`, `CustomConditionsEditorState.svelte.ts`) (new) — the editor. Ids auto-fill from the name until edited by hand; save is disabled while any id is empty, reserved, or duplicated.
- **`src/managers/ConditionManager.ts`** — `initialize()` now clears its map and registers records **synchronously**. See below.
- **`src/nimble.ts`** — registration runs from the `i18nInit` hook, after `prelocalize` and before Foundry initializes documents. `setup` is too late: every item's `RulesManager` is built in between, so a rule naming a custom condition would fail its `choices` validation and be silently dropped.
- **`public/lang/en.json`** — new `NIMBLE.settings.customConditions.*` block. **English source strings added by AI — please review the wording.** No other locale files touched.
- 19 new tests across three files.

### The one non-obvious change

`ConditionManager.initialize()` used `forEach(async …)` and `set()` its map record *after* awaiting `enrichHTML`, so records landed asynchronously — and it never cleared the map. Both had to change for the settings `onChange` path to work: calling `initialize()` then `configureStatusEffects()` back-to-back would otherwise publish an **empty** `CONFIG.statusEffects`, and a removed custom condition would linger forever. Records are now registered synchronously with the raw enricher text as a placeholder, upgraded in place once the text editor resolves. 5 of the 6 new `ConditionManager` tests fail against the old implementation and pass against the new one.

## Test Plan

Automated: `pnpm check` — 228 files / 3572 tests, exits 0.

Manual, in Foundry as a GM:

1. **Game Settings → System Settings → Custom Conditions → Manage Conditions.** Empty state should read "No custom conditions yet."
2. Click **Add Condition**, type a name like `Hexed`. The id field should auto-fill to `hexed`. Add a description and pick an icon. **Save Conditions.**
3. **Without reloading**, right-click a token → the Token HUD status grid should now include *Hexed* with your icon. Click it; the condition applies and the token shows the icon.
4. Open the token's actor sheet → **Conditions** tab. *Hexed* should be listed, and its tooltip should show your description.
5. Type `[[/condition condition=hexed]]` into a journal or chat message — it should enrich to a *Hexed* button with the description in its tooltip.
6. Open any feature/spell item → **Effects** config → the condition dropdown should include *Hexed*. (Item sheets already open when you saved need reopening — see below.)
7. Reload the world. *Hexed* should still be there, and any token you applied it to should still have it.
8. Go back into the editor and **remove** *Hexed*, then save. It should disappear from the Token HUD and condition lists without a reload.
9. **Validation:** add a row and try id `prone` → "That id is reserved by a built-in condition."; two rows with the same id → "Duplicate id."; a blank id → "Id is required." Save should be disabled in each case.
10. **As a player** (non-GM): the Custom Conditions button should not appear in settings, but custom conditions should still render correctly on sheets and the HUD.

Known limitation: sheets and rule dropdowns that were already open when you saved won't refresh their condition list until reopened. Rule `choices` are read at render, and re-rendering every open sheet class for a rare GM action didn't seem worth it. Open conditions panels and the token HUD do update immediately.

Out of scope (happy to follow up if wanted):
- **Overriding built-in conditions** (rename / re-icon / re-describe) — built-in ids are reserved. `languageSettings.ts` has the `{ overrides, custom }` pattern if we want this later.
- **Alias / linked / stackable / overlay / trigger relationships** for custom conditions — a custom condition is a plain, non-stackable, non-overlay status. Exposing those graphs would add four more controls per row for mechanics the system automates only for specific built-ins.

## Checklist

- [ ] Tested in Foundry with no console errors — **not done; needs a human pass** (see Test Plan)
- [ ] **If yes:** I have reviewed all AI-generated code against the repo's [STYLE_GUIDE.md](docs/STYLE_GUIDE.md) and [AGENTS.md](AGENTS.md) — author review pending

## Related Issues

Closes #484


Fronix (dismissed):
Nice feature, and the editor itself is clean. The blocker is where the merge into `CONFIG` is wired: as written, custom conditions do not survive a reload once they are referenced from a rule, so the headline interaction with `conditionImmunity` / `applyCondition` does not actually hold up in a real world.

## Blocking

**1. Custom conditions are merged into `CONFIG.NIMBLE` too late (`src/settings/registerCustomConditionSettings.ts:82`)**

`registerSystemSettings()` runs from the `setup` handler, but Foundry's boot order is `init` -> `i18nInit` -> `initializeDocuments()` -> `Hooks.callAll("setup")`. That means every Item's `RulesManager` is built while `CONFIG.NIMBLE.conditions` still holds only the built-ins.

Consequence: a `conditionImmunity`, `applyCondition` or `markTarget` rule whose `condition` is a custom id fails `StringField` choices validation during data prep, `RulesManager` catches it, and the rule is silently never `set()`. Characters happen to recover because `ready.ts` calls `actor.reset()` on every character for an unrelated reason; monsters, NPCs, unlinked token actors and world Items stay broken until something else re-preps them.

Fix: do the merge in the existing `i18nInit` handler (after `prelocalize`, before document init) rather than in `setup`.

**2. Custom descriptions are injected into tooltip HTML unescaped (`src/view/dataPreparationHelpers/prepareActorConditions.ts:112`)**

`buildConditionTooltip` deliberately runs name, duration, source and the condition tag through `escapeHtml`, then interpolates `${description}` raw, which was fine when descriptions were trusted i18n HTML. They are now free-form GM text that reaches every player's client on every condition tooltip. A stray `<` truncates the tooltip; an `` executes.

`src/enrichers/enrichCondition.ts:16` has the same issue for the condition *name*, and `:21` for the description, with no escaping at all. The name in particular has no reason to allow markup.

**3. All-digit ids corrupt `CONFIG.statusEffects` (`src/settings/customConditionSettings.ts:52`)**

`sanitizeConditionId` permits a purely numeric id. V14's `CONFIG.statusEffects` proxy `set` trap writes `statuses[status.id] = status` against the raw target array, so an id of `13` writes an array *index*, not a string key. With an id larger than the entry count the array goes sparse and `NimbleTokenHUD.svelte`'s `{#each statusEffects as effect}` throws on `effect.img`. An id of `length` produces a `RangeError` inside the trap and takes out `configureStatusEffects()` outright.

Fix: reject all-digit ids and `length` in `sanitizeConditionId`. The editor's validation (`CustomConditionsEditorState.svelte.ts:28-37`) only covers empty / reserved / duplicate today.

## Non-blocking, but worth doing here

**4. Saved ids stay editable, and delete has no usage check (`CustomConditionsEditorState.svelte.ts:65`)**

Rows loaded from the setting come back with `idEdited: true`, so a saved id can be renamed freely. Renaming or deleting orphans every live ActiveEffect carrying that status *and* every stored rule referencing it, with no warning. The orphans are then hard to clean up: `ConditionManager.getMetadata` skips statuses it cannot resolve, so they vanish from the sheet's Conditions tab, and the token HUD's Clear All builds its id list from `CONFIG.statusEffects`, so it skips them too. The GM has to go find them in Temporary/Passive Effects.

Minimum viable version: lock the id once a row has been saved, and warn on delete when actors currently carry the condition. A real rename migration can be a follow-up.

**5. Rules Builder shows raw ids for conditions (`src/models/rules/conditionImmunity.ts:12`)**

`choices: () => Object.keys(CONFIG.NIMBLE.conditions)` hands `resolveChoices()` an array, which it maps to `[v, v]`, so "Soul Burned" renders as `soul_burned` in the `conditionImmunity` tag group and the `applyCondition` select. `markTarget.ts:35` already passes the object form and gets proper labels. Matching it here also fixes the pre-existing `blinded` / `lastStand` labels.

**6. Index-keyed `{#each}` over `bind:value` rows (`CustomConditionsEditor.svelte:24`)**

Keying on `index` makes the key a no-op, so `removeRow(index)` reuses block N's DOM for what used to be row N+1. A focused, dirty description textarea can flush its pending value into the row that shifted up. Give each row a `foundry.utils.randomID()` uid at creation and load, and key on that.

**7. `captureBuiltInConditions()` can cache an empty snapshot permanently (`customConditionSettings.ts:34`)**

If it ever runs before `CONFIG.NIMBLE` is populated, `{ ...config.conditions }` yields `{}` with no error, and the `if (builtInConditions) return` guard then blocks any recapture. The next merge wipes all built-ins out of `CONFIG`. Currently safe only because the `setup` registration happens to be the first caller, but `getBuiltInConditionIds()` is exported. Bail without caching when `config?.conditions` is missing.

**8. `onChange` does not re-render anything (`registerCustomConditionSettings.ts:57`)**

The PR description says it "re-renders actor sheets so edits land without a reload", but the handler only does merge + `initialize()` + `configureStatusEffects()`. `prepareActorConditions` is `$derived` over `effectVersion`, not over `CONFIG`, so an open sheet, Rules Builder or conditions panel keeps the stale list. Either add the re-render or correct the description; as written, step 4 of the test plan will read as a failure.

## Notes, no action needed for this PR

- The feature is close to a verbatim clone of custom spell schools (`sanitizeConditionId` is byte-identical to `sanitizeSpellSchoolKey`, the registration file differs only by noun, and `.condition-card` mirrors `.school-card`). Not worth blocking on, but every fix above now has a twin that will drift. A shared `createCustomRegistrySetting()` plus one parameterised editor is the eventual shape.
- The new `--nimble-validation-error-*` tokens duplicate literals still hard-coded in `CustomSpellSchoolsEditor.svelte` (lines 177, 284-285, 295), so the dark-mode contrast fix only lands for conditions. Worth migrating both call sites whenever the above consolidation happens.
- `picker.browse()` in `CustomConditionsEditorState.svelte.ts:82` is fire-and-forget; a rejection is a silent unhandled promise with no feedback to the GM.

trevlar:
## Fixed

- **Merge landed after documents were built** — registration now runs from the `i18nInit` hook (after `prelocalize`, before `initializeDocuments()`) instead of `registerSystemSettings()`, so a rule naming a custom condition passes its `choices` validation on the first data prep (`src/nimble.ts:85`).
- **GM description injected raw into tooltip HTML** — descriptions are escaped and paragraph-wrapped on the way into `CONFIG.conditionDescriptions`, so both the tooltip and the enricher get safe markup from one choke point (`src/settings/customConditionSettings.ts:81`). The stored value stays plain text, so the editor round-trips it unchanged.
- **Condition name injected raw into the enricher** — the heading and the button label share one escaped value (`src/enrichers/enrichCondition.ts:13`); `createEnricherButton` writes the label through `innerHTML`, so it needed the same treatment.
- **All-digit and array-property ids corrupt `CONFIG.statusEffects`** — `isUnsafeConditionId` rejects them at the reader and in the editor, with its own error message (`src/settings/customConditionSettings.ts:68`). It covers `length` and every `Array.prototype` name, not just digits: `push` would shadow the method `configureStatusEffects()` republishes with.
- **Saved ids stayed editable, delete had no usage check** — rows loaded from the setting carry `persisted: true`; their id input is read-only, and removing one that actors are currently carrying goes through a confirmation naming the count (`CustomConditionsEditorState.svelte.ts:88`). Unlinked token actors are counted too.
- **Raw ids in the Rules Builder** — `conditionImmunity` passes the object form, matching `markTarget`; this also fixes the pre-existing `blinded` / `lastStand` labels (`src/models/rules/conditionImmunity.ts:12`).
- **Index-keyed `{#each}` over `bind:value` rows** — each row gets a `randomID()` uid at creation and load, and the block keys on that (`CustomConditionsEditor.svelte:24`).
- **Empty built-in snapshot cached permanently** — `captureBuiltInConditions()` bails without caching when `config?.conditions` is missing, and the merge bails with it rather than wiping CONFIG (`customConditionSettings.ts:37`).
- **`onChange` re-rendered nothing** — it now fires `nimble.conditionsChanged`, which `ActorConditionsList` listens for and bumps `effectVersion` on, so open conditions panels re-derive (`registerCustomConditionSettings.ts:68`). The PR description is corrected to match what actually refreshes.
- **`picker.browse()` was fire-and-forget** — a rejection now logs and notifies the GM (`CustomConditionsEditorState.svelte.ts:136`).

## Not fixed

- **Consolidating the custom-conditions and custom-spell-schools twins** — every fix above has a twin in the spell schools files that will now drift, including the `i18nInit` timing one. That is a shared `createCustomRegistrySetting()` plus one parameterised editor, which is a bigger change than this PR should carry. Same for migrating `CustomSpellSchoolsEditor.svelte`'s hard-coded colours onto the new `--nimble-validation-error-*` tokens.
- **A real rename migration for ids** — locking the id is the minimum viable version you described; rewriting live effects and stored rules on rename is its own piece of work.

## Approach

Escaping happens once, when custom conditions are merged into CONFIG, rather than at each of the three interpolation sites. That keeps built-in descriptions (trusted i18n HTML) untouched and means anything reading `CONFIG.conditionDescriptions` later is safe by construction. The trade-off is that custom descriptions are plain text with paragraph/break handling instead of HTML, which matches the plain `<textarea>` the editor already had; docs updated to say so.

Registration moved to `src/nimble.ts` as a second `Hooks.once('i18nInit', ...)` rather than into `i18nInit()` itself. The test harness calls `i18nInit()` directly during global setup, so folding it in there would have pulled the settings module and a Svelte component into every suite's bootstrap.

`isUnsafeConditionId` uses `id in Array.prototype` rather than a hardcoded list, which covers `length` and the mutation methods in one check.

Live refresh went through a hook rather than re-rendering sheets: the condition list is `$derived`, so a sheet `render()` would not have re-run it anyway. Rule dropdowns read `choices` at render time and still need a reopen; that is now stated accurately in the description instead of overclaimed.

`pnpm check` passes: 234 files / 3620 tests. 11 new tests across the four affected suites.

Pushed as `a71b265`.

Fronix (commented):
Review of the custom conditions work. All 39 new tests pass. Findings below, most significant first.

Three of them land in files this PR does not touch, so they cannot be inline comments. They are all cases where a change made here has a matching site that was missed:

**1. `src/view/chat/components/ConditionNode.svelte:29` - the same raw interpolation fixed in `enrichCondition.ts`, missed here.**

`getConditionTooltip` builds `<h3 ...>${label}</h3>` with the unescaped condition name and assigns it to `data-tooltip`. A GM condition named `Hex <3 Bane` produces an `<h3>` that never closes, so Foundry's `cleanHTML` swallows the tag span and the description, and every player hovering that chip on the chat card gets a truncated tooltip. v14 runs `cleanHTML` over tooltip content, so this is markup corruption rather than script execution; the `enrichCondition.ts` escape added in this PR was the security-relevant one, since that label lands in `button.innerHTML`. Fix: `foundry.utils.escapeHTML(label)`, matching `src/enrichers/enrichCondition.ts:12`.

**2. `src/models/rules/applyCondition.ts:40` - still shows raw ids in the rule GMs actually use.**

`choices: () => Object.keys(CONFIG.NIMBLE.conditions)` returns bare ids. Create `Soul Burned` (id `soul_burned`), open a feature's Rules Builder and pick Apply Condition: the dropdown renders `soul_burned`, because `SchemaFieldRenderer` maps an array form to `[v, v]`. Condition Immunity, changed in this same PR with the comment "The object form makes the tag group render 'Soul Burned' instead of the raw `soul_burned` id", and `markTarget.ts:35` both render the label. The same one-line change applies here.

**3. `src/view/pixi/NimbleTokenHUD.svelte:37` - an open Token HUD never picks up a condition edit.**

`const statusEffects = CONFIG.statusEffects;` is a non-reactive const, the component registers no `CONDITIONS_CHANGED_HOOK` listener, and `configureStatusEffects()` mutates the array in place (`length = 0` then `push`), so the array identity never changes and the `{#each}` has nothing to invalidate. Keep a token selected while saving a new condition and the HUD grid stays stale until the token is deselected and reselected. `docs/documentation/gm/settings.md`, added in this PR, states "The token panel and any open conditions list pick up changes right away." Fix: mirror `ActorConditionsList.svelte` with a `$state` version counter bumped from `Hooks.on(CONDITIONS_CHANGED_HOOK, ...)`.

src/view/settings/CustomConditionsEditorState.svelte.ts
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+
+	return actors;
+}
+
+function countActorsWithCondition(conditionId: string): number {
+	let count = 0;
+	for (const actor of collectActors()) {
+		if (actor.statuses?.has(conditionId)) count += 1;
```
Fronix: `countActorsWithCondition` reads `actor.statuses`, which only lists statuses from **active** effects, so the in-use warning silently does not fire in the cases where it matters most.

`Actor#applyActiveEffects` does `if (!effect.active) continue;` before `this.statuses.add(statusId)` (v14 `client/documents/actor.mjs:243-251`). An actor carrying the custom condition via a disabled effect, or via an effect granted by an unequipped or otherwise suppressed item, never appears in `actor.statuses`. The count comes back 0, `removeRow` skips the confirm, the GM saves, and the effect is orphaned with no route to clear it from the conditions list. That is exactly the outcome `removeInUseWarning` exists to prevent.

The codebase already documents this pitfall at `src/utils/applyConditionToActor.ts:57`.

Fix: iterate `actor.effects` / `allApplicableEffects()` and test `effect.statuses.has(id)`.

src/settings/customConditionSettings.ts
```diff
@@ -0,0 +1,159 @@
...
+		conditions[custom.id] = custom.name;
+		descriptions[custom.id] = toDescriptionHtml(custom.description);
+		images[custom.id] = custom.img;
+	}
+
+	const config = CONFIG.NIMBLE as unknown as ConditionConfig;
+
+	// Mutated in place, not reassigned: anything holding a reference to one of these dictionaries
+	// from init would otherwise keep the stale copy.
+	replaceEntries(config.conditions, conditions);
```
Fronix: `replaceEntries` deletes every key currently in the three CONFIG dictionaries and rebuilds them from a snapshot taken at `i18nInit`, so anything added after that point is destroyed.

If a module (or future system code running at `setup` / `ready`) sets `CONFIG.NIMBLE.conditions.hexed` plus the matching description and image entries, the next `mergeCustomConditionsIntoConfig()` wipes it. That call happens the first time the GM saves the Manage Conditions editor, even for an unrelated row. The follow-up `initialize()` / `configureStatusEffects()` then drops `hexed` from `CONFIG.statusEffects`, and every actor carrying it vanishes from the Conditions tab and the Token HUD.

Merging only the stored custom ids (add and delete by id) instead of a wholesale rebuild would leave foreign entries intact.

src/settings/registerCustomConditionSettings.ts
```diff
@@ -0,0 +1,84 @@
...
+			config: false,
+			type: Array,
+			default: [],
+			onChange: () => {
+				mergeCustomConditionsIntoConfig();
+				// `CONFIG.statusEffects` is a snapshot, not a live view of the config, so the manager
+				// has to rebuild its records and republish it.
+				game.nimble.conditions.initialize();
+				game.nimble.conditions.configureStatusEffects();
+				// CONFIG is not reactive, so an already-rendered condition list only re-derives if
```
Fronix: The setting's `onChange` has no error isolation, so any throw aborts before `Hooks.callAll(CONDITIONS_CHANGED_HOOK)` and leaves every open conditions list silently stale.

`onChange` runs inside `Setting#_onUpdate` (v14 `client/documents/setting.mjs:47-52`). `game.nimble.conditions` is dereferenced unguarded, and `configureStatusEffects()` throws "Conditions are not ready yet." when `#ready` is false. Any throw propagates out of the document update handler: the hook never fires, open panels keep the old list with no notification, and the error surfaces as an unrelated document-update failure.

Fix: wrap the body in try/catch, log with the `Nimble |` prefix, and fire the hook in a `finally`.

src/settings/customConditionSettings.ts
```diff
@@ -0,0 +1,159 @@
...
+ */
+let builtInConditions: ConditionConfig | null = null;
+
+function captureBuiltInConditions(): void {
+	if (builtInConditions) return;
+
+	const config = CONFIG.NIMBLE as unknown as ConditionConfig | undefined;
+	// An empty snapshot would be cached forever and the next merge would wipe the built-ins out
+	// of CONFIG, so bail without caching rather than capturing an unpopulated config.
+	if (!config?.conditions) return;
```
Fronix: `captureBuiltInConditions` guards only on `config.conditions`, so a partially-populated CONFIG gets cached permanently and the merge then throws with no recovery.

If `CONFIG.NIMBLE.conditionDescriptions` or `conditionDefaultImages` is absent when the first merge runs (a registration-order change, or a partial-config path in tests), the guard still passes, `{ ...undefined }` caches `{}` for that dictionary forever, and `replaceEntries(config.conditionDescriptions, ...)` throws `TypeError: Cannot convert undefined or null to object` at `Object.keys(target)`. The cache is written once, so no later call can repair it.

Fix: guard all three keys, and null-check inside `replaceEntries`.

src/settings/registerCustomConditionSettings.ts
```diff
@@ -0,0 +1,84 @@
...
+import localize from '#utils/localize.js';
+import CustomConditionsEditor from '#view/settings/CustomConditionsEditor.svelte';
+import {
+	CONDITIONS_CHANGED_HOOK,
+	CUSTOM_CONDITIONS_SETTING_KEY,
+	mergeCustomConditionsIntoConfig,
+} from './customConditionSettings.js';
+
+const EDITOR_ICON = 'fa-solid fa-biohazard';
+
```
Fronix, line 13: The module-level `openEditor` singleton is never cleared, and it duplicates a registry `GenericDialog` already owns.

`openEditor = this` is set in `render()` and nothing ever nulls it. After the GM closes the editor, the `CustomConditionsMenu` instance, its mounted Svelte tree and the `rows` state stay reachable for the rest of the session, and each subsequent open replaces rather than releases the previous one.

`GenericDialog` already implements this via `uniqueId` plus `#openDialogs` and `getOrCreate` / `isOpen` / `closeById` (`src/documents/dialogs/GenericDialog.svelte.ts:24-110`). This is now the second verbatim copy of the class-plus-global pattern; see `registerSpellSchoolSettings.ts:13,30-38`.

Fix: add a public accessor on `GenericDialog` for the existing instance and drop both copies, or at minimum clear `openEditor` in `close()`.

src/settings/customConditionSettings.ts
```diff
@@ -0,0 +1,159 @@
...
+		.replace(/^_+|_+$/g, '');
+}
+
+/**
+ * Ids that break `CONFIG.statusEffects`. Its V14 proxy mirrors each entry onto the backing array
+ * under `array[status.id]`, so an all-digit id writes an array index (going sparse, or throwing a
+ * RangeError at `length`) and an `Array.prototype` name shadows the method used to republish it.
+ */
+export function isUnsafeConditionId(id: string): boolean {
+	return /^\d+$/.test(id) || id in Array.prototype;
```
Fronix, line 74: `id in Array.prototype` walks the whole prototype chain, so ordinary English words are rejected as condition ids with a misleading error.

"Flat", "Find", "Values", "Every", "Sorted" (`sort`), "Joined" (`join`) and "Sliced" (`slice`) all trip it, as do Foundry's own `Array.prototype` extensions (`equals`, `partition`, `findSplice`). The GM sees "That id cannot be used. Try adding a letter." and a disabled Save, with nothing explaining why.

Only the ids the system actually indexes or calls on the backing array are dangerous: `length`, `push`, `at`, `find`, `findIndex`, `map`, plus all-digit ids. An explicit denylist of those plus `constructor` and `__proto__` keeps the guard and drops the false positives.

src/managers/ConditionManager.ts
```diff
@@ -29,53 +33,53 @@ export class ConditionManager {
...
-					`[[/condition condition=${id}]]`,
-				);
+				data._id = String(id).padEnd(16, '0');
 			}
 
+			// Registered synchronously, with the raw enricher text as a placeholder, so that
+			// `configureStatusEffects()` can run immediately after `initialize()`.
 			this.#conditions.set(id, data);
-		});
+			this.#enrich(data);
```
Fronix: `initialize()` fires one untracked `#enrich` promise per condition, and it is now re-run on every setting change on every client.

Each GM save triggers `onChange` and therefore `initialize()` on every connected client, spawning roughly 35 unawaited `enrichHTML` calls, each of which re-enters the `[[/condition]]` enricher. A GM iterating in the editor multiplies this, and a second `initialize()` before the first batch settles leaves the earlier promises resolving into `Condition` records already dropped from `#conditions` and from `CONFIG.statusEffects`, so the work is discarded.

Nothing in `src/` reads `Condition.enriched` (no production consumer turns up), so the cheapest fix is to enrich lazily on first read. Failing that, a generation counter so stale batches short-circuit.

src/view/settings/CustomConditionsEditorState.svelte.ts
```diff
@@ -0,0 +1,189 @@
...
+
+		const cleaned = rows
+			.map((row) => {
+				const id = sanitizeConditionId(row.id);
+				const name = row.name.trim() || id.charAt(0).toUpperCase() + id.slice(1);
+				const description = row.description.trim();
+				const img = row.img.trim() || DEFAULT_CUSTOM_CONDITION_ICON;
+				return { id, name, description, img };
+			})
+			.filter((row) => row.id);
```
Fronix: `.filter((row) => row.id)` in `save()` is dead code.

`save()` returns early on `hasErrors`, and `rowErrors` already maps any row whose sanitized id is empty to `t('errorEmptyId')`, which makes `hasErrors` true. No row reaching this filter can have an empty id. As written it reads as a second validation layer and suggests empty ids are silently dropped rather than blocking the save.

src/config/registerConditionsConfig.ts
```diff
@@ -1,9 +1,13 @@
...
- * Canonical status-effect IDs for system-managed conditions. The keys here are the
- * source of truth — the dictionaries below are typed against this so adding/renaming
- * an ID in one place forces an update in the others.
+ * Canonical status-effect IDs for the conditions that ship with the system. The keys here are
+ * the source of truth for those: the dictionaries below are typed against this, so adding or
+ * renaming an ID in one place forces an update in the others.
+ *
+ * At runtime the dictionaries are open-ended. GMs can merge arbitrary IDs into them through the
+ * custom conditions setting (`src/settings/customConditionSettings.ts`), so treat a condition ID
+ * read from CONFIG as a plain string: only these built-ins are guaranteed to be present.
```
Fronix: This comment now declares the dictionaries runtime-open-ended, but the declared types stay `Record<StatusEffectId, string>`, so call sites have to cast around them.

`conditionImmunity.ts:14` needs `as Record<string, string>` purely to widen a type this comment says is no longer accurate, and any future caller indexing with a GM-supplied id hits the same wall.

The fix belongs at the declaration: give the dictionaries an index signature (`Record<StatusEffectId, string> & Record<string, string>`) so `STATUS_EFFECT_IDS` keeps enforcing the built-in set while arbitrary string reads type-check without casts.

src/view/settings/CustomConditionsEditorState.svelte.ts
trevlar: Fixed in `13d1e06` — the count walks `allApplicableEffects()` (falling back to `effects`) and tests `effect.statuses.has(id)`, so a disabled effect or one on an unequipped item is counted. New test covers the suppressed-effect case, plus one asserting an unused condition still deletes without a prompt.

src/settings/customConditionSettings.ts
trevlar: Fixed in `13d1e06` — the merge now tracks the ids it wrote (`mergedCustomIds`) and only deletes those, so a foreign `hexed` set at `setup`/`ready` survives. Rebuilding from a snapshot is gone entirely, which also retired `replaceEntries`. Test asserts a module-added entry across all three dictionaries survives a later save.

src/settings/registerCustomConditionSettings.ts
trevlar: Fixed in `13d1e06` — body is wrapped in try/catch logging with the `Nimble |` prefix, `game.nimble?.conditions?.` is optional-chained, and `Hooks.callAll` moved to `finally`. Two tests: a throwing `initialize()` and a missing `game.nimble` both still fire the hook.

src/settings/customConditionSettings.ts
trevlar: Fixed in `13d1e06` — one `getConditionConfig()` helper guards all three dictionaries and returns null otherwise; both the capture and the merge bail on it. Since the merge no longer rebuilds from a snapshot, only the built-in *ids* are cached now, so there is nothing left to null-check inside. Test deletes `conditionDescriptions` and asserts the merge neither throws nor writes.

src/settings/registerCustomConditionSettings.ts
trevlar: Fixed in `13d1e06` — added `GenericDialog.getOpen(uniqueId)` over the existing `#openDialogs` registry (`isOpen` now delegates to it), and dropped both module globals. `close()`/`submit()` already unregister, so nothing outlives the dialog. Took the drop-both-copies route rather than the minimum, since the leak was identical in `registerSpellSchoolSettings.ts`.

src/managers/ConditionManager.ts
trevlar: Fixed in `13d1e06` — dropped the async upgrade rather than making it lazy. Since nothing reads it, a getter that side-effects an enrichment on read seemed worse than no enrichment at all; the field keeps the enricher source and is renamed `enricherText` so the name is honest. That removes the promise storm and the stale-batch window in one go. Test asserts a double `initialize()` calls `enrichHTML` zero times.

src/view/settings/CustomConditionsEditorState.svelte.ts
trevlar: Fixed in `13d1e06` — filter removed.

src/config/registerConditionsConfig.ts
trevlar: Fixed in `13d1e06` — added `ConditionDictionary = Record<StatusEffectId, string> & Record<string, string>` and typed `conditions` / `conditionDefaultImages` with it, so `STATUS_EFFECT_IDS` still forces the built-in set while arbitrary reads type-check. The `conditionImmunity.ts:14` cast is gone. `conditionDescriptions` stays `Record<string, string>`: `charged`, `distracted` and `hampered` have no description, so it never satisfied the required half.

src/settings/customConditionSettings.ts
trevlar: Pushed back on this one — kept `id in Array.prototype` and fixed the error message instead (`13d1e06`).

The denylist reopens more than it closes. `CONFIG.statusEffects` is a real array and the proxy writes `array[id]`, so an id of `filter`, `sort`, `forEach`, `some`, `slice`, `reduce` or `includes` shadows that method for every consumer — core v14 and any module, not just our call sites. Enumerating what is dangerous means enumerating what every current and future caller happens to invoke; the list you named leaves all seven of those open, and the failure surfaces as a `TypeError: not a function` in unrelated code with nothing pointing back to a condition name.

Against that, the false positives are thinner than the examples suggest: `sanitizeConditionId` maps "Sorted" to `sorted`, "Joined" to `joined` and "Sliced" to `sliced`, none of which trip the guard. What actually collides is the bare single words — `find`, `flat`, `every`, `values` — and the cost of rejecting one is a keystroke.

So the part I took is the misleading error. `errorInvalidId` now names the collision and what it would break rather than "Try adding a letter":

> "find" cannot be used: ids that are all digits, or that match a JavaScript array property such as "length" or "find", would corrupt the token HUD. Try a longer or more specific id.

Happy to narrow it if you still want that, but it felt like trading a loud, recoverable rejection for a silent, hard-to-trace break.

trevlar:
## Fixed

- **Foreign CONFIG entries destroyed on save** — the merge tracks the ids it wrote and deletes only those, instead of clearing every key and rebuilding from the built-in snapshot (`customConditionSettings.ts:126`).
- **In-use count blind to suppressed effects** — walks `allApplicableEffects()` and tests `effect.statuses.has(id)`, so a disabled effect or one on an unequipped item counts (`CustomConditionsEditorState.svelte.ts:47`).
- **`onChange` could swallow the change hook** — wrapped in try/catch with the `Nimble |` prefix, `game.nimble` optional-chained, `Hooks.callAll` in `finally` (`registerCustomConditionSettings.ts:59`).
- **Partial CONFIG cached permanently** — one `getConditionConfig()` guards all three dictionaries; capture and merge both bail on it, and only built-in ids are cached now (`customConditionSettings.ts:27`).
- **Leaked editor singletons** — `GenericDialog.getOpen(uniqueId)` reads the existing `#openDialogs` registry, and both module globals are gone, spell schools included (`GenericDialog.svelte.ts:94`).
- **Enrichment storm on every save** — dropped the async upgrade rather than making it lazy; the field keeps the enricher source and is renamed `enricherText` (`ConditionManager.ts:19`).
- **Dead id filter in `save()`** — removed (`CustomConditionsEditorState.svelte.ts:161`).
- **Dictionaries needed casts to read** — `ConditionDictionary = Record<StatusEffectId, string> & Record<string, string>`; the `conditionImmunity` cast is gone (`registerConditionsConfig.ts:50`).
- **Raw name in the chat chip tooltip** — `escapeHTML` on the label, matching `enrichCondition.ts` (`ConditionNode.svelte:26`).
- **Raw ids in the Apply Condition select** — passes the object form like `conditionImmunity` and `markTarget` (`applyCondition.ts:40`).
- **Token HUD stale while open** — `$state` version counter bumped from `CONDITIONS_CHANGED_HOOK`, and the grid derives a fresh array so the in-place rebuild invalidates it (`NimbleTokenHUD.svelte:39`).

## Not fixed

- **Narrowing `isUnsafeConditionId` to a denylist** — pushed back inline; kept the prototype-chain guard and rewrote the error message instead. Reasoning on the thread.
- **Consolidating the conditions / spell-schools twins** — still its own change, as last round. The `openEditor` fix above is the one piece that landed in both, since the leak was verbatim identical.

## Approach

The wholesale rebuild and the partial-config guard turned out to be one fix: once the merge adds and deletes by id, the built-in *values* never need snapshotting, so only the id list is cached and `replaceEntries` is gone. That removes the class of bug rather than the instance.

On `Condition.enriched` — you are right that nothing reads it, which is why I removed the enrichment rather than deferring it. A getter that kicks off a side-effecting enrichment on property read is the kind of thing AGENTS.md's KISS rule is aimed at, and lazily enriching for zero consumers is work either way. The field now holds the enricher source under an honest name; anything that wants markup can enrich it.

`pnpm check` passes: 234 files / 3628 tests, 17 new across six suites. `conditionDescriptions` is the one dictionary that stayed `Record<string, string>` — `charged`, `distracted` and `hampered` have no description, so it cannot satisfy the required half of `ConditionDictionary`.

Pushed as `13d1e06`.

Fronix (approved):

PR #912, opened 2026-08-07
feat(ancestry): choose which kind of people a multi-variant ancestry is (#159)

### LINKED ISSUE #159: On character creation, add feature for selecting ancestry specifics when choosing multi-option such as oozeling/construct,

trevlar:
Closes #159

Three shipped ancestries bundle two kinds of people under one entry — Dryad/Shroomling, Minotaur/Beastfolk, Oozeling/Construct — and the rules treat the two as the same ancestry with free flavour ("Flavor Is Free"). Character creation had no way to say which one a character is, so every such character read as "Dryad/Shroomling" on their sheet, in chat cards, and in the PDF export.

## What changed

**Ancestries declare their names, nothing is inferred.** A new `system.variants` array holds the names an ancestry covers. A slash in a name means whatever the GM meant by it — so the three pack entries list their names explicitly, and homebrew opts in the same way rather than by how it happens to be named.

**GM authoring** — the ancestry sheet's Config tab gets a "Variant Names" control beside the size options: a text field adds a name, badges remove them, and a summary line states what the list means for character creation. Re-adding a name the list already carries keeps the text in the field with a note, rather than vanishing as though it had been added.

**Player choice** — two or more names open **Step 2c. Ancestry Variant**: radio rows that pick exactly one with nothing pre-selected, gating the ancestry-options stage the way the size choice does. Size and the enhanced save shift to 2d and 2e. A lone name is the ancestry's own, so it asks nothing. Switching ancestry drops the pick, since variants belong to the ancestry that offers them.

**On the finished character** — the embedded ancestry takes the chosen name, so the sheet, chat cards, and the PDF export all read "Shroomling". Only a name that ancestry actually offers is honoured; a stale or hand-built submission can't rename the item to anything it likes.

## The identifier catch

`NimbleBaseItem.prepareBaseData()` derives `system.identifier` from the item's name, unconditionally, and an ancestry identifier is what keys the GM's language grants and alias names (`character.ts` `_applyConfiguredLanguageGrants`). Renaming the embedded item to "Shroomling" would therefore have silently dropped any `dryad-shroomling` grants a GM had configured.

So the rename records the identifier the ancestry had beforehand, and `NimbleAncestryItem` now lets a declared identifier outrank the derived one. A renamed Shroomling still answers to `dryad-shroomling`. Scoped to ancestries deliberately: 79 pack items store a non-empty identifier (classes, subclasses), so changing this in the base class would be a much wider behaviour change. A side effect is that the ancestry sheet's existing Identifier input now actually sticks, which it did not before.

## Also

`AncestrySizeSelection.svelte` hosted three choices, so it becomes `AncestryOptionsSelection.svelte`, and the variant and size radios share one local snippet instead of a second copy of the same markup and ~60 lines of SCSS.

## Testing

`pnpm check` exits 0 — Biome format, ESLint + Biome lint, dependency-cruiser (no violations), `tsc --noEmit`, 3588 tests across 227 files. Migration count check passes (41 == 41; the new field has an `initial` and rewrites no world data). `pnpm build:compendia` builds all 14 packs clean.

New coverage:

- `ancestryVariants.test.ts` — the pure helpers, including blanks, case-insensitive repeats, and authored order.
- Three state-gate tests — gates until a variant is chosen, skips for a single kind of people, drops the pick when the ancestry changes.
- Six component tests — the new variant radios *and* the size radios, which had no coverage before this refactor touched their markup.
- Five submit tests — renames to the chosen variant, pins the identifier, leaves things alone with no variant chosen, ignores a variant the ancestry doesn't offer, respects an already-declared identifier.
- Two item tests — a declared identifier survives data prep; one is still derived from the name when absent.

## Out of scope (deliberate)

- **No migration for world copies.** Character creation reads compendium ancestries, so the pack edits cover the shipped case; a GM's world copy gets variants through the new sheet control.
- **Descriptions untouched.** "Dryads and Shroomlings embody…" still names both. Rewriting rules prose per variant is guesswork worth leaving to a human.
- **Variant names aren't Babele-translatable.** The skeleton generator extracts `name`/`description`/`rules` only; adding `variants` would mean regenerating non-English locale files, which `AGENTS.md` forbids agents from touching.

## Note for review

The English strings in `public/lang/en.json` are agent-written and want a human read — in particular the step renumbering (2c/2d/2e) and the sheet hint and summary wording.


Fronix (approved):
LGTM

---

## Optional fixes

**One test expects an identifier that Foundry cannot produce.**

`CharacterCreationDialog.test.ts:395` expects `dryad-shroomling`. The real `slugify` method deletes each character that is not a letter or a number, so `Dryad/Shroomling` gives `dryadshroomling`. The polyfill at `tests/setup.ts:26` first changes `/` to a space, then to `-`. The polyfill is wrong, but `AGENTS.md` tells you not to change `tests/setup.ts`.

Six other places already use `dryadshroomling`, including `languageSettings.test.ts:111`, which holds the key that the GM stores. Change the literal to `dryadshroomling`.

This test is the only guard on the language grant. If the identifier changes, `_applyConfiguredLanguageGrants` stops to grant Sylvan Cant to a Shroomling, and this test still passes.

**Six comments name a specific ancestry.**

A comment in `src/` must describe only the mechanism. These lines contain the names Dryad, Shroomling, or Invoker of Chaos:

- `src/documents/item/ancestry.ts:80` and `:84`
- `src/utils/ancestryVariants.ts:4` and `:54`
- `src/models/item/AncestryDataModel.ts:16`
- `src/documents/item/item.d.ts:77`
- `src/documents/dialogs/CharacterCreationDialog.svelte.ts:264`

**The new component has no types.**

`AncestryOptionsSelection.svelte` is a new file with 467 lines. It starts with a plain `<script>` tag and declares seven props inline. `AGENTS.md` tells you to use `lang="ts"` and to put prop types in `types/components/*.d.ts`. Your `VariantOptionsInput.svelte` does both.

**The documentation needs two changes.**

`docs/documentation/characters/creation.md:23` adds two em-dashes. The file on `dev` has none. Em dashes do not look good, so do not add them.

`docs/documentation/homebrew/character-options.md:25` says that the Config tab has "an Identifier field and an Exotic Ancestry checkbox". The tab now also has Size Options and Variant Names.

These three problems are less important:

- Arrow keys no longer select a size. The old code moved the selection with the focus, and a comment gave the reason. The new `handleRadioKeydown` moves only the focus. The variant group has a reason for this change. The size group does not.
- `ancestryVariants.ts:5` quotes "Flavor Is Free" as the reason that a Dryad and a Shroomling are the same kind of people. That sidebar permits you to use the mechanics of one ancestry for a different ancestry. It is a different rule.
- The ancestry sheet now uses the Identifier field. The class sheet and the spell sheet still ignore it. Six subclasses store an identifier that the system does not read.

PR #913, opened 2026-08-07
feat(spells): let a dropped spell become a spell scroll

### LINKED ISSUE #395: Allow creation of spell scroll
We need some way to allow spells scroll to be created when you drag a spell into a character sheet. Shouldn't be automatic but be one of the options.

- Add spell to spell list
- Add spell as spell scroll to inventory

trevlar:
Closes #395.

Dragging a spell onto a character sheet had exactly one outcome and asked nothing: it landed on the spell list, the sheet force-switched to Spells whatever tab you were on, and no notification fired. The only route to a scroll was dragging one of ten generic `Spell Scroll - Tier N` blanks and retyping the spell by hand, which lost the spell entirely.

Every spell drop now opens a chooser, and every scroll-template drop opens a picker. The question is always asked — the open tab carries no meaning.

**Design doc, with screenshots of the built dialog in both themes:** https://claude.ai/code/artifact/44cdf760-ee29-40c6-8b9f-c01ebc79f5d2

## What it does

**Chooser** — spell list vs scroll, defaulted to the spell list so Enter reproduces the old behaviour. States the four differences that matter: mana equals the spell's tier, a listed spell can be upcast to your highest unlocked tier while a scroll is fixed at its inscribed one, and whether the Arcana check applies. Warns when the actor has no mana, where a listed spell would sit uncastable.

**Picker** — the tier's spells, filtered by school using the icons the Spells tab already uses, searchable, with rows that expand to show the spell's description. Secret spells are offered to a GM only, never to a player.

**Scrolls cast the spell they carry.** The activation is copied, so a scroll costs the spell's actions, no mana, cannot be upcast, and is consumed. A DC 10 Arcana check gates the effect only when the wielder knows no spell of that school, per the rulebook exemption — cantrips count.

**Scrolls share a slot.** `objectSizeType: 'smallSized'`, so all of them together cost one inventory slot rather than one each. Migration 042 retargets copies of the pack blanks.

**The drop-landing flash follows the actual destination** instead of always requesting the Spells tab, and falls back to the absorbing row when a stacking create folds into an existing scroll.

## Decisions behind it

Ten mechanics were settled with @trevlar before implementation and are recorded in the design doc with their code and rulebook references. The load-bearing ones:

- Mana equals tier (cantrips free); a listed spell does not use Arcana, a scroll does.
- A listed spell can be upcast to `highestUnlockedSpellTier`; a scroll never can.
- The Arcana check is conditional on not knowing the school.
- Scrolls share one inventory slot.
- Secret spells are GM-only in both directions.
- The fixed scroll rules are always written; the spell's own description is a world setting.

## Nothing is spent until the player commits

Worth review attention, because the first implementation got this wrong in four ways:

- A scroll with nothing to roll and no check due asks outright, naming the scroll and saying it will be consumed. `ItemActivationManager` skips its dialog when there is nothing to roll and the item is not a spell, so without this a scroll was consumed on a single click.
- A required check goes through the normal skill-check flow and gets the Configure Arcana Skill Check dialog, with a hint naming the school and the stake. Closing it leaves the scroll in the pack.
- Closing the activation dialog leaves the scroll alone. It is spent on a completed roll or a failed check, never on a cancel.

## Style notes

The dialog reuses the house patterns rather than inventing its own:

- The choice cards mirror `.rest-type-card` from `FieldRestDialog` — hidden native input, whole label as the control, selection dot top-right. Five of the system's six radio components hide the input; a visible one was the outlier.
- The accent uses `--nimble-action-info-*` rather than the literal amber `FieldRestDialog` hardcodes, because those tokens carry `.theme-dark` values and the literals do not. Every colour in the component resolves through a token, so it ships **zero** dark-specific rules and both themes were correct first time.
- One full-width footer action naming its own outcome — "Add to spell list" / "Add to inventory" / "Inscribe and add to inventory" — matching Level Up, Field Rest, Safe Rest and seven others. Dynamic labels are themselves the house pattern.

## Verification

`pnpm check` green: 226 files, 3579 tests. 24 new unit tests cover the tier→price table, the utility-spell case, activation copying, tier pinning, and description composition.

Walked through in a running Foundry v13 world in both themes: chooser and picker, the default spell-list path, scroll creation, tab switch and landing flash, slot arithmetic (two scrolls added, slots unchanged), secret-spell gating, the Arcana check skipped for a known school and rolled for an unknown one, a failed check wasting the scroll, duplicate stacking, and all three cancel paths.

**No schema changes.** Ten compendium template JSONs edited plus one migration.

## Not in this PR

Expandable inventory rows and inline weapon damage (T4 in the doc). They improve every item and touch only `PlayerCharacterInventoryTab.svelte`, so they are separable and shouldn't gate this.

## For a reviewer

English strings only, flagged for human review; no other locale touched. Two follow-ups worth their own tickets: `FieldRestDialog` and `HitPointSelection` both hardcode selected-state colours with no dark values, and the shipped secondary navigation hardcodes its filter divider the same way.


Fronix (dismissed):
## Request changes

Nice feature, and the happy path (character drops a spell, picks it from the chooser, casts it) reads well. I checked the implementation against the scroll rules in CoreRules-2.md before writing this up, so the notes below separate genuine rules deviations from code issues.

### Where the implementation matches the rules

Worth stating so none of this gets "fixed" by accident:

- Flat DC 10, not scaled by tier. That is the scroll rule; the tier-scaled DC belongs to wands (`DC 10 + spell tier`), which are a separate item with a separate exemption.
- `knowsSpellSchool` implements the exemption exactly as written ("if they do not already know a spell from that school"), with cantrips counting as spells of their school.
- A failed check consumes the scroll, matching "on a failure, it is wasted".
- Action cost carries over, since the spell's whole `activation` block is cloned. The rules require the spell's normal action cost.
- No mana is spent: `manaToSpend` only exists on the upcast path, which is gated on `type === 'spell'`.

### Blocking: rules deviations

**1. NPCs get a free pass on the Arcana check** (`src/documents/item/object.ts:222`)

`if (!actor?.rollSkillCheck) return 'passed'` is not a neutral fallback, it changes the rules outcome. The rulebook makes exactly one exemption from the DC 10 check: already knowing a spell from that school. It does not exempt anyone else, and "does not require magical ability, anyone who can read the language can utilize it" is about who may *attempt* a scroll, not about who skips the check. As written, an NPC with no spells of that school auto-succeeds where an identical PC would roll, so NPCs are strictly better at scrolls than heroes.

`rollSkillCheck` is defined only on `NimbleCharacter` (`src/documents/actor/character.ts:1098`); `NimbleNPC` has nothing equivalent. So a GM handing an NPC a Scroll of Fireball gets no roll, no chat message, and an unconditional cast. AGENTS.md is explicit that an unsupported path should throw rather than no-op, and this is the rules-critical branch. Either give NPCs a real check or throw until they have one.

**2. Schoolless spells can never reach the exemption** (`src/utils/spellLabels.ts:15`, `knowsSpellSchool`)

`SpellDataModel.school` initialises to `''`, so schoolless spells are real documents (the new `knowsSpellSchool` tests rely on it). `knowsSpellSchool` returns false for `''`, so those scrolls always demand the check with no way to ever satisfy it, and `getSpellSchoolLabel('')` returns `''`, so the player is told "DC 10, Modrisse knows no  spells" with a double space and no noun. The rulebook assumes every spell has a school, so this is undefined territory rather than a contradiction, but the current behaviour picks the harshest reading and then fails to explain itself. Suggest deciding it deliberately: either treat a schoolless spell as exempt, or keep the check and give the string a fallback noun.

**3. The "cannot be upcast" claim is presented as a rule and is not one**

The doc comment on `activate` and the `noUpcast` locale string both assert that a scroll cannot be upcast. The rulebook never says this. It says upcasting means spending additional mana, up to the tier you have unlocked (CoreRules-2.md:410 and 1358), and scroll casting spends no mana, so the conclusion does follow, but it is an inference across two rules rather than a quoted one. Please reword to state the mechanism ("a scroll spends no mana, so there is nothing to upcast with") rather than asserting a rule that is not in the book, so nobody later cites our string back at us.

### Blocking: correctness

**4. A scroll can be consumed with no prompt at all** (`src/documents/item/object.ts:151`)

`#hasActivationRolls()` is used to predict "an activation dialog will appear, so skip the confirm dialog", but `ItemActivationManager` also skips its dialog when `activation.skipRollDialog` is set or Alt is held. `createScrollFromSpell` deep-clones the whole activation block, `skipRollDialog` included (`src/models/item/common.ts:67`). So for a spell authored with `skipRollDialog: true` and a damage node: confirm dialog skipped, `#resolveDialogData` computes `skipDialog = skipRollDialog ? !altPressed : altPressed` -> true, card posts, `#consumeScroll()` deletes the item. Same outcome for any scroll if the player happens to be holding Alt. The one click that was supposed to always raise a confirmable prompt spends the scroll outright.

**5. Hidden Arcana rolls are broadcast to everyone** (`src/documents/item/object.ts:235`)

`rollSkillCheck` returns `{ roll, rollData }`, and `rollData.visibilityMode` is `'blindroll'` when the GM ticks the hide-roll box in the Configure dialog. `#rollScrollArcanaCheck` destructures only `roll` and calls `ChatMessage.create(...)` with no `ChatMessage.applyRollMode(...)`. Every sibling path (`rollSkillCheckToChat`, `rollAbilityCheckToChat`, `rollSavingThrowToChat`) applies it. Result: the hidden roll, its total, and the pass/fail line all go public.

**6. Template detection reads a key Foundry never wrote** (`src/utils/getSpellScrollTemplateTier.ts:54`)

`getItemSourceId` resolves `item.sourceId ?? item._stats?.compendiumSource ?? item.flags?.core?.source` (`src/utils/itemSourceRules.ts:10`). The real legacy key is `flags.core.sourceId`, which is what `MigrationBase#getSourceId` and `Migration042` both read. So a `Spell Scroll - Tier 3` stamped by an older Foundry gets migrated correctly but is not recognised on drop: no picker opens, the raw blank is created. The new unit test asserts `flags: { core: { source: ... } }`, so the gap ships behind a green test.

### Should fix before merge

**7. Em-dashes get baked into stored item content** (`public/lang/en.json:3231` and eight others)

New em-dashes at 3231, 3232, 3243, 3255, 3256, 3257, 3258, 3268, 3269, plus `arcanaCheckHint`. Beyond the house style rule, `buildScrollRulesHtml()` calls `localize()` at creation time and writes the result into `system.description.public`, so `noMagicalAbility` and `noUpcast` become permanent content on every scroll ever created. A later locale cleanup will not touch documents already in a world. Colons or parentheses work fine here ("No magical ability needed: anyone who can read it can use it."). Worth fixing in the same pass as item 3, since `noUpcast` needs rewording anyway.

**8. Concentration is dropped when inscribing** (`src/utils/createScrollFromSpell.ts:119`)

Only `system.activation` is copied; `system.properties.selected` is not. `ObjectDataModel` allows `'concentration'` (`src/models/item/ObjectDataModel.ts:79`), `NimbleSpellItem` stamps it onto the chat card (`src/documents/item/spell.ts:155`), and the status effect is tracked on the actor (`base.svelte.ts:449`). Inscribe a concentration spell and cast it: no marker on the card, nothing flags the actor, the constraint quietly disappears. A scroll casts the spell as the spell is written, so it should carry the spell's own properties.

**9. A tier with no candidate spells is a dead end** (`src/documents/sheets/PlayerCharacterSheet.svelte.ts:189`)

Once `getSpellScrollTemplateTier` matches, the picker is the only outcome: `if (!result?.spellUuid) return null` and `_onDropItem` creates nothing. If the world has no tier-8 spell documents, `buildCandidates` returns `[]`, the dialog renders `noCandidates`, and `isSubmitDisabled` stays true, so closing the window is the only exit and nothing is created. The old behaviour (drop the blank, fill it in by hand) is gone with no replacement, so there is no way to place an uninscribed scroll, shop stock, or a scroll of a spell that has no Item document. This one is a product call as much as a bug: if blanks are meant to be placeable, the chooser needs a "keep it blank" option.

### Non-blocking cleanups

**10. The batch-drop path is unreachable** (`src/documents/sheets/PlayerCharacterSheet.svelte.ts:208`)

`const itemData = item.toObject()` is never an array, so `const items = Array.isArray(itemData) ? itemData : [itemData]` always has length 1. That makes `items.length === 1` always true, `spells.length !== items.length` always false, and `batchCount` always 1. The `batchCount > 1` branch in `SpellScrollDialog.svelte` and the `batchChooserTitle` locale entry can never render, so the documented "GM drags a folder onto an NPC" case does not exist. Either wire up a real multi-item drop or drop the scaffolding.

**11. `#hasActivationRolls()` duplicates manager-private logic** (`src/documents/item/object.ts:181`)

`flattenEffectsTree(effects).some(node => node.type === 'damage' || node.type === 'healing')` appears verbatim at `src/managers/ItemActivationManager.ts:828`. In the manager it is one input among several; here it is treated as the whole answer. Adding `savingThrow` to that list later would silently reintroduce item 4. Better for the manager to expose the decision (something like `willPromptForActivation()`) than for the item to re-derive it.

**12. The picker eagerly loads and enriches every candidate** (`src/view/dialogs/openSpellScrollDialog.ts:78`)

`buildCandidates` indexes every Item pack, then for each entry of the tier does `await fromUuid(...)` (a per-document compendium fetch, not an index read) plus `await enrichSpellText(...)` (which resolves `@UUID` links). For the cantrip template that is every cantrip and utility spell in every installed module, to render collapsed rows that only need `activationSummary`. Adding `system.activation.cost` to the index fields removes the `fromUuid` entirely, and description can be enriched lazily in `toggleExpanded`.

**13. No-mana warning asserts something it does not check** (`src/view/dialogs/SpellScrollDialog.svelte:61`)

The warning is gated on `hasMana` alone, but the text also claims "and no unlocked spell tiers". `highestUnlockedSpellTier` is a separate field a GM can set directly in PlayerCharacterSettingsTab. An actor with `mana.max = 0` and `highestUnlockedSpellTier = 3` sees the warning directly under an Upcasting line reading "Up to Tier 3, your highest unlocked". Note that neither field actually bears on whether the scroll works: a scroll needs no mana and no unlocked tier, so this card is explaining why upcasting is unavailable, and the copy should say only that.

**14. Global `dl dt` reset for one dialog's styling** (`src/scss/base/_base.scss:9`)

`.nimble-sheet dl dt { text-shadow: none; }` hits every definition list under any Nimble sheet: actor sheets, item sheets, existing dialogs. The only `<dl>` this PR adds is `.nimble-spell-scroll-dialog__facts`. Scoping it to the component's own `<style>` block (as every other component here does) fixes the same problem without the blast radius.

trevlar:
## Fixed

- **NPCs skipping the Arcana check** — an actor with no `rollSkillCheck` now throws instead of resolving `passed`, so the only exemption left is the rulebook's (`src/documents/item/object.ts:207`). Worth stating plainly: NPCs have no inventory in this system, so nothing can hand one a scroll to begin with. The branch is unreachable, and now says so.
- **Schoolless spells** — `createScrollFromSpell` refuses to inscribe one, and the drop handler reports it and lets the spell be learned normally instead (`src/utils/createScrollFromSpell.ts:124`, `src/documents/sheets/PlayerCharacterSheet.svelte.ts:222`). A spell without a school is broken data, not a case to render a fallback noun for.
- **"Cannot be upcast" asserted as a rule** — the string now states the mechanism: upcasting spends extra mana, a scroll spends none (`public/lang/en.json:3243`).
- **Scroll consumable with no prompt** — the `#hasActivationRolls()` prediction is gone. When no Arcana check applies the confirmation always runs, and the prompt is the commit point: past it the scroll is spent whether or not the activation dialog produces a card (`src/documents/item/object.ts:148`). That closes the `skipRollDialog` and held-Alt cases in one place instead of widening the predicate.
- **Hidden Arcana rolls going public** — the pass/fail message is built, passed through `ChatMessage.applyMode(chatData, toMessageMode(rollData.visibilityMode))`, then created (`src/documents/item/object.ts:240`).
- **Template detection missing old blanks** — `flags.core.sourceId` is read alongside `getItemSourceId`, with the modern compendium source still winning (`src/utils/getSpellScrollTemplateTier.ts:58`).
- **Em-dashes baked into stored scroll content** — every new string in the block is rewritten; the rules lines that get written into `system.description.public` are clean (`public/lang/en.json:3236`).
- **Concentration dropped when inscribing** — the spell's `properties.selected` transfers, filtered to what keeps its meaning on an object. `range`/`reach` stay behind: on `ObjectDataModel` those are weapon reach backed by their own min/max fields, not the spell's range (`src/utils/createScrollFromSpell.ts:29`).
- **Unreachable batch-drop scaffolding** — removed. `#resolveSpellScrollDrop` takes the one item Foundry resolved, and `batchCount` plus `batchChooserTitle` are gone (`src/documents/sheets/PlayerCharacterSheet.svelte.ts:192`).
- **`#hasActivationRolls()` duplicating manager-private logic** — dissolved with the predicate itself. Nothing re-derives the manager's decision now.
- **Picker loading every candidate** — rows are built from the pack index alone (`system.activation.cost` added to the indexed fields), and a description is fetched only when its row is expanded (`src/view/dialogs/openSpellScrollDialog.ts:71`).
- **No-mana warning claiming more than it checks** — the copy now says only what `hasMana` proves (`public/lang/en.json:3258`).
- **Global `dl dt` reset** — moved into the dialog's own scoped block (`src/view/dialogs/SpellScrollDialog.svelte:200`).

## Not fixed

- **"A tier with no candidate spells is a dead end"** — rejected on its premise. Nimble ships world items covering every spell school, and there is always a spell at a given tier; a world where that is not true is one where someone has broken their own system install. No "keep it blank" escape hatch, per YAGNI.

## Approach

The prompt is now the commit point, which is the single idea behind the consumption change: nothing is spent before it, everything after it is. A failed check wastes the scroll, a passed one casts the spell and spends it, and cancelling the prompt leaves it untouched. That is simpler to reason about than predicting which downstream dialog might appear, and it is why item 11 needed no separate fix.

`src/documents/item/object.test.ts` is new and covers the activation flow that had none: both prompt paths, cancel/pass/fail, the roll mode, stack decrement, automation off, and the throw.

Lazy descriptions are fetched through a loader passed in as dialog data rather than imported, so the dialog does not reach back into the module that opens it.

Pushed as `dab544a`. Thanks @Fronix — the roll-mode and `flags.core.sourceId` catches were both real and neither had a failing test to show for it.

Fronix (changes requested):
Looks good just have a few small things

1. The "Add to spell list" warning looks a bit wonky, icon doesn't fit properly and the yellow color doesn't really hit the contrast rules.
Image

2. If you have DSN installed the resolution of the DC 10 check is resolved before the dice has been rolled. So you get the "Activate spell"-popup before the dice has even been rolled, it should wait for DSN to resolve the dice roll then show the popup (if the DC was met).

trevlar:
## Fixed

- **Activation prompt opening over rolling dice** — the Arcana check now awaits `game.dice3d.waitFor3DAnimationByMessageID(message.id)` after posting the pass/fail card, so the outcome, the consumed scroll, and the activation dialog all land after the throw. DSN's own API resolves immediately when the module is absent, disabled, or set to `immediatelyDisplayChatMessages`, so nothing waits when there is nothing to wait for (`src/documents/item/object.ts:264`).
- **Unreadable warning hint on a dark sheet** — `.nimble-hint--warning` now carries dark-theme colors of its own, the way `--success` already does (`src/scss/components/_hint.scss:24`). Measured off your screenshot the old body text sat at 3.2:1; it is 5.7:1 on the selected card and 7.3:1 on an unselected one now.
- **Icon not fitting its disc** — two causes. `--nimble-warning-icon-background-color: var(--nimble-light-text-color)` is declared on `:root`, so it substitutes there and keeps the light-theme cream no matter the theme, which is the pale blob in the shot; the dark override drops the disc instead. And the glyph was `fa-triangle-exclamation` in a `border-radius: 50%` disc, so the scroll dialog now uses `fa-circle-exclamation` like every other warning hint in the app (`src/view/dialogs/SpellScrollDialog.svelte:54`).

## Not fixed

- **The same `:root` token indirection in `CharacterStatConfigDialog` and `PredicateBuilder`** — both read `--nimble-warning-*` directly and inherit the same light-theme disc on a dark sheet. Left alone deliberately: fixing it means retuning the shared tokens for two dialogs this PR does not touch.

## Approach

The contrast fix is scoped to the hint variant rather than the shared `--nimble-warning-*` tokens, since those feed two other dialogs and a token retune would change surfaces outside this PR. `_hint.scss` already models this: `--success` defines its own colors with a `.theme-dark &` override, and `--warning` now matches.

`object.test.ts` covers the wait with a deferred animation promise: it asserts the base activation has not run while the promise is pending, then runs once it resolves. It fails if the `await` is dropped.

Verified the hint by rendering the real markup and tokens against Foundry's FontAwesome and sampling the pixels, since the numbers depend on the translucent fill compositing over the selected card.

Pushed as `493130f`.

Fronix (approved):

PR #914, opened 2026-08-09
feat(dice): distinct Dice So Nice color for the primary die with per-user customization

Fronix:
## Summary

When Dice So Nice is active, the primary die of a damage roll now renders with a distinct appearance (crimson body, gold numerals and edge by default) so it stands out from the rest of the dice pool, and each user can customize or disable the styling for their own rolls.

image
DSN not enabled/installed
<br />
image

Rolling
image

## Changes

- New `src/dice/diceSoNiceIntegration.ts`: registers a system colorset on the `diceSoNiceReady` hook and builds the per-term appearance override from the rolling user's preferences. Without the module active, the hook never fires and the term options are inert.
- `DamageRoll._extractPrimaryDie` (both branches) and the `c`/`cv` die modifier handlers tag their dice with `options.appearance`, which Dice So Nice reads per term. Term options serialize with the roll, so the roller's colors show on every client. Vicious explosion dice inherit the styling because their results live on the primary term.
- New `src/settings/diceSoNiceSettings.ts`: three user-scoped settings (v14 `scope: 'user'`, stored per account) with an enable toggle plus `ColorField`-backed color pickers for the die body and numerals. A custom body color also recolors the edge to match; invalid values fall back to the defaults.
- Unit tests for the appearance builder (defaults, disabled, custom colors, short hex expansion, Color instance stringification, invalid input fallback).

## Test Plan

1. Install and enable Dice So Nice in a v14 world, then roll damage from a weapon or attack feature. The leftmost primary die should be crimson with gold numerals while the remaining dice use your normal DSN theme. Crit explosion dice (including vicious chains) keep the primary color.
2. Open Configure Settings, Nimble tab: change "3D Dice: Primary Die Color" and "3D Dice: Primary Die Number Color" via the color pickers, roll again, and confirm the new colors. Confirm both settings render as color pickers, not text inputs.
3. Toggle "3D Dice: Distinct Primary Die" off and confirm the primary die reverts to your regular DSN dice theme.
4. Roll as a second user with different colors and confirm each player's rolls carry their own colors on both screens.
5. Modifier-mode formulas (e.g. `4d4cv` via the dice testbench) should color every crit-capable die.
6. `pnpm test` covers the appearance-building logic in `src/dice/diceSoNiceIntegration.test.ts`.

## Checklist

## Related Issues

None.

### DIFF
--- a/docs/documentation/gm/settings.md
+++ b/docs/documentation/gm/settings.md
@@ -51,6 +51,15 @@ Want the system to carry as much as possible? Leave every Automation toggle **on
 Prefer to stay hands-on? Open Configure Automation and turn **off** Apply Conditions and Effects from Rules (conditions become one-click suggestions on chat cards instead), plus any other family you'd rather track yourself, such as Resource Spending or Action Tracking. Leave **off** Auto-Add Character To Combat On Initiative Roll and Auto-Track Token Adjacency. Consider turning **on** Hide Rolls by Default on your own client so GM rolls stay behind the screen. The system still rolls the dice and does the math. It just asks before changing anything.
 :::
 
+## 3D dice
+
+If your table uses the **Dice So Nice** module, the **Configure 3D Dice** button in the system settings tab opens a window where each player chooses how the primary die of their damage rolls looks. The primary die is drawn in its own colors so it reads apart from the rest of the pool, and the colors travel with the roll: everyone at the table sees the roller's choice, not their own.
+
+- **Distinct Primary Die** is on by default. Turn it off to have the primary die use your normal Dice So Nice theme.
+- **Primary Die Color** and **Primary Die Number Color** set the body and the numerals. **Reset Colors** puts both back to the crimson and gold defaults.
+
+All three are per-user, so every player sets their own. Without Dice So Nice installed and enabled, the button is greyed out: these settings only change 3D dice, and there is nothing for them to affect.
+
 ## The combat tracker has its own settings
 
 The combat tracker at the top of the screen isn't configured from the settings menu. Instead, click the gear button on the tracker itself to open the **Combat Tracker Settings** window. Size, colors, hit point bar display, and what players are allowed to see all live there, and every option is listed in the [Settings Reference](../reference/settings.md).
--- a/docs/system/dice-engine.md
+++ b/docs/system/dice-engine.md
@@ -144,7 +144,7 @@ Content authors can leave `weaponType` blank (nothing breaks) or set it for weap
 ## What the engine does NOT do
 
 - **It doesn't render the chat card.** That lives in `src/view/chat/components/` and consumes the finalized roll.
-- **It doesn't integrate with Dice So Nice.** DSN visualizes whatever dice are on the roll; the engine has already decided the outcome before DSN sees anything. One current quirk: the vicious explosion path manually rerolls the primary die to avoid DSN preempting the crit chain. Don't touch that without talking to whoever owns DSN integration.
+- **It doesn't take results from Dice So Nice.** DSN visualizes whatever dice are on the roll; the engine has already decided the outcome before DSN sees anything. The one thing the engine hands DSN is cosmetic: `_extractPrimaryDie` and the `c`/`cv` modifier handlers tag the primary term with an `appearance` option so the primary die renders in the roller's own colors (`src/dice/diceSoNiceIntegration.ts`). Those options are inert without the module. One current quirk: the vicious explosion path manually rerolls the primary die to avoid DSN preempting the crit chain. Don't touch that without talking to whoever owns DSN integration.
 - **It doesn't do targeting.** Target selection is upstream in the activation flow; the engine just receives a target count.
 - **It doesn't validate rule compliance.** If you pass it a mixed-type primary pool (`d6 + d8`), it will roll it. The rules say you shouldn't, but the engine doesn't police that — that's on the content layer.
 
@@ -155,6 +155,7 @@ Content authors can leave `weaponType` blank (nothing breaks) or set it for weap
 | The roll formula preprocessing | `src/dice/DamageRoll.ts` (`_preProcessFormula`, `_applyRollMode`) |
 | The primary die logic | `src/dice/terms/PrimaryDie.ts` |
 | Tie-aware adv/dis handlers | `src/dice/nimbleDieModifiers.ts` |
+| Dice So Nice primary die styling | `src/dice/diceSoNiceIntegration.ts`, `src/settings/registerDiceSoNiceSettings.ts` |
 | Die modifier metadata (`c`, `cv`, `v`, `n`) | `src/dice/nimbleDieModifiers.ts` (`getNimbleMods()`, `NIMBLE_MODS`) |
 | Where custom modifiers get registered | `src/hooks/init.ts` (calls `registerNimbleDieModifiers()`) |
 | Roll construction from an item | `src/managers/ItemActivationManager.ts` |
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -2768,16 +2768,30 @@
 				"name": "Auto-Add Character To Combat On Initiative Roll",
 				"hint": "When enabled, rolling initiative from a character sheet adds that character to the current scene combat before rolling, if they are not already in combat."
 			},
+			"diceSoNiceMenu": {
+				"name": "3D Dice",
+				"manageButton": "Configure 3D Dice",
+				"hint": "Choose how the primary die of your damage rolls looks in Dice So Nice. Each player sets this for themselves.",
+				"title": "3D Dice Settings",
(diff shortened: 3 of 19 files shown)


trevlar (dismissed):
**🔧 Request changes: The integration works, but the three DSN settings should be grouped and the dice engine docs need to be updated**

I tested this in Foundry with the latest version of Dice So Nice, and it works. Choosing custom body and numeral colors changes the dice as expected. You can consider the unchecked “Tested in Foundry” item covered; there’s no need to verify it again.

I also checked the Dice So Nice source at tag `6.2.9`, since parts of its documentation are out of date:

* `DiceFactory.getAppearanceForDice` calls `mergeObject(appearance, dicenotation.options.appearance)` after resolving the colorset. That means the flat `{colorset, background, foreground, edge}` structure is correct, and the custom colors properly override the colorset defaults.

* `Dice3D.addColorset(colorset, "default")` only registers the colorset. `"preferred"` is the option that would override player themes, so `"default"` is the right choice here.

* `DamageRoll.fromData` restores `roll.terms` from the serialized data, so the roller’s appearance settings do carry over to other clients as described.

**Must fix before merge**

* **`src/settings/diceSoNiceSettings.ts:71`**
  My main concern is that the three `3D Dice: ...` settings are added directly to Nimble’s flat settings list. They should be grouped behind either a “Dice So Nice” divider or header, or preferably their own settings dialog.

  We already have a pattern for this: register the three settings with `config: false`, create a `GenericDialog` subclass, and expose it through `game.settings.registerMenu(SYSTEM_ID, 'diceSoNiceMenu', {...})`.

  See `src/settings/registerAutomationSettings.ts:108` and `src/view/settings/AutomationSettingsDialog.svelte`, or `registerCustomConditionSettings.ts:81`. The one difference is that this menu should use `restricted: false`, since these are user-scoped settings that every player needs to access.

* **`docs/system/dice-engine.md:147`**
  This still says, “It doesn’t integrate with Dice So Nice,” which is no longer true. Both `DamageRoll` and `nimbleDieModifiers` now import the DSN integration.

  Please update that bullet and add `src/dice/diceSoNiceIntegration.ts` to the “Where things live” table. The document specifically asks us to update it when the implementation changes, so we should keep it accurate as part of this PR.

**Worth fixing, but non-blocking**

* **`src/dice/diceSoNiceIntegration.ts:88`**
  Add `visibility: 'hidden'`. `Utils.js:302` builds the player-facing theme list by filtering `COLORSETS` for entries where `visibility !== 'hidden'`, and `addColorset` defaults to `'visible'`.

  Without this, “Nimble Primary Die” will appear in every player’s DSN appearance dropdown under a new “Nimble” group as a selectable theme for the entire dice pool.

* **`src/dice/diceSoNiceIntegration.ts:86`**
  `outline: '#310606'` is designed for the default crimson body, but `PrimaryDieAppearance` does not include an `outline` field. If the player changes the body color, the background and edge update while the numeral outline remains dark red. The outline should be derived from the custom color for the same reason as the edge.

* **`src/dice/nimbleDieModifiers.ts:86`**
  The `appearance` is attached to the Die term, so modifier mode renders all four dice in `4d4cv` using the primary colors. Nothing visually distinguishes the primary die from the rest of the pool.

  The test plan says this is intentional, which is fine if that is the desired behavior. It is still worth confirming whether the styling should instead be applied only to `roll.primaryDie`.

* **`src/dice/diceSoNiceIntegration.ts:79`**
  `addColorset` is asynchronous because it awaits `FontConfig.loadFont` internally, but its promise is neither awaited nor caught. Make the hook callback async and await the call so failures surface properly instead of becoming unhandled rejections.

* **`src/dice/DamageRoll.ts:344`**
  One minor robustness concern: in DSN’s resolution order, `appearance.colorset` has lower priority than `options.colorset` and damage-type detection through `options.type` or `options.flavor`.

  The multi-die branch sets `flavor: 'Primary Die'`, while the single-die branch does not set a flavor. If a GM happens to map that flavor, it could silently override the primary styling. That is unlikely, but the inconsistency between the two branches is worth reviewing.

**What looks good**

* Applying the appearance to the term options when the roll is created is the right approach. It allows the roller’s per-user settings to travel to every client, and it survives serialization correctly.

* Namespacing the colorset as `${SYSTEM_ID}-primary` prevents collisions with the development rolling release.

* The appearance-builder tests cover the important cases: disabled settings, defaults, custom values, short hex colors, `Color` stringification, and invalid input.

**Requested changes**

1. Move the three DSN settings into a submenu dialog.
2. Update the outdated Dice So Nice statement in `dice-engine.md`.
3. Set the colorset’s visibility to `'hidden'`.
4. Update the outline along with the background and edge when custom colors are used.
5. Confirm whether modifier mode should style every crit die or only the primary die.
6. Await or catch the `addColorset` promise.
7. Make the flavor behavior consistent between the two primary-die branches.

Fronix:
Some points:

- would be better to have it come up in a separate dialog. Like a button that opens up dice so nice settings for Nimble.
- The button should be grayed out/disabled if DSN is not installed/enabled and there should be some kind of info that they need to enable/install DSN first

trevlar (dismissed):
**🔧 Push back — the integration itself checks out, but nothing tests that the colours reach the dice**

**Must fix before merge**

- **src/dice/nimbleDieModifiers.ts:90** — `applyPrimaryDieAppearance` is new and completely untested. The one test file you added covers `getPrimaryDieDiceOptions`, which only builds an object and never touches a Die term, so nothing proves the appearance actually lands on a die. `nimbleDieModifiers.test.ts` already has a `createMockDie` helper, so four small tests cover it: `nimbleCrit` sets `options.appearance.colorset`; `nimbleCritVicious` does the same; an appearance already on the term is preserved; a disabled preference leaves `options` untouched.

- **src/dice/DamageRoll.ts:344** — the legacy path has the same gap. `_extractPrimaryDie` builds the `PrimaryDie` term in two places: one for a multi-die pool like `2d6` (:344), and one for a single-die pool like `1d8` (:362). You added the appearance to both, and no test checks either. `DamageRoll.test.ts` already rolls formulas that go down both branches, so each needs one new assertion: the Primary Die term it produced has `options.appearance` set.

- **src/dice/diceSoNiceIntegration.ts:121** — nothing tests the colorset payload passed to `addColorset`, and it has to match the settings defaults. `getPrimaryDieDiceOptions` deliberately emits *only* `colorset` when the user is on defaults. So if `DEFAULT_PRIMARY_DIE_COLOR` and the colorset's `background` ever drift apart, the default Primary Die renders the wrong colour and no test fails. Invoke the `diceSoNiceReady` callback with a stub `dice3d` and assert name, background, foreground and visibility.

- **src/settings/diceSoNiceSettings.ts:69** — `setPrimaryDiePreferences` writes all three settings on every save, even the untouched ones. The sibling `AutomationSettingsDialog` diffs against an `initialValues` snapshot first and writes only what changed. Matching that turns the common "I only changed the body colour" save into one write instead of three.

**What's good**

- I checked the DSN assumptions against installed Dice So Nice 6.2.9 and every one holds: `addColorset(..., 'default')` really is a no-op (only `'preferred'` mutates user state), `options.appearance` is merged *over* the resolved colorset in `getAppearanceForDice`, `dsnDamageTypeManaged` does short-circuit flavor-based damage-type detection, and `visibility: 'hidden'` only hides the set from the theme picker. The comments explaining each are accurate, which is rare for a third-party integration.

- Cross-client behaviour is right for a non-obvious reason: `DamageRoll.fromData` re-runs the constructor, and so `getPrimaryDieDiceOptions()` on the *viewing* client, but then overwrites `terms` with the deserialized ones. The roller's colours win everywhere.

- `pnpm docs:generate` produces a clean **3D Dice** section with the right defaults, and the `type.initial` fallback in `describeSettingDefault` is the correct fix for `ColorField`-typed settings.

---

1. Test that the `c` and `cv` handlers tag the die
2. Test that both `_extractPrimaryDie` branches tag the term
3. Test the registered colorset matches the defaults
4. Write only the settings that actually changed

Fronix:
**1 and 2 are fair, both done** in `8f57ffc6`. `nimbleDieModifiers.test.ts` gets four tests (`c` tags the die, `cv` tags the die, an existing appearance is preserved, a disabled preference leaves `options` alone). `DamageRoll.test.ts` asserts the appearance on both `_extractPrimaryDie` branches, plus that the leftover damage die in `2d6` stays untagged. I mutation-checked them: removing either piece of the source fails the new tests.

**3 I would skip.** The colorset payload does not repeat the colour literals, it passes the same constants `getPrimaryDieDiceOptions` compares against (`diceSoNiceIntegration.ts:126-127`), so the drift cannot happen. A test there asserts a constant equals itself.

**4 I would skip.** `AutomationSettingsDialog` diffs because it has fifteen toggles. This one has three, and the dialog is opened rarely. Not worth an `initialValues` snapshot to keep in sync.

PR #918, opened 2026-08-28
fix(rules): move per-level maxHpBonus with the class level

### LINKED ISSUE #499: maxHPBonus Rule is not recalculated on Level-Up
Currently it is possible to create a rule like this:

```
{
  "type": "maxHpBonus",
  "disabled": false,
  "id": "iU7DaavgKe68RVMf",
  "identifier": "some-feature",
  "label": "Some Feature",
  "predicate": {},
  "priority": 1,
  "value": 2,
  "perLevel": true
}
```

It works as intended until you Level up. Then the "Bonus From Rules in HP Overview" does show the correct value, but the static bonus is wrong. It does not fix itself even after deleting+adding the item again.

Image

Screenshot was made directly after leveling up a PC with said bonus above.

eastagiletracker:
This PR proposes a fix for the per-level `maxHpBonus` rule, so a rule carried by a feature moves with the character's class level instead of staying frozen at the level its item was added on. We include this PR work along with a full history of your repo at https://eastagiletracker.com/projects/469. You can sign in with your GitHub ID to claim ownership of the project.

## Summary

A `maxHpBonus` rule with `perLevel: true` folds `value * classLevel` into the actor's stored `system.attributes.hp.bonus` once, in `preCreate`, and takes the same amount back out in `afterDelete`. Nothing re-derived it in between. The only level-change handler was `MaxHpBonusRule#preUpdate`, and rule `preUpdate` is dispatched by `NimbleBaseItem#_preUpdate` on the item being updated — so it fired only when the rule happened to sit on the class item itself. The rule in #499 lives on a feature, so leveling up never moved the stored bonus. That is why the HP Overview's "from rules" total, which `EditHitPointsDialog.svelte` recomputes live off `system.rules`, read correctly while the stored bonus did not.

It also explains the "does not fix itself even after deleting+adding the item again" part of that report. With `value: 2, perLevel: true` added at level 2, `preCreate` stores 4; leveling to 3 stores nothing, so 4 stands where 6 is due; deleting the item runs `afterDelete`, which subtracts `2 * @level` at the *current* level — 6 — leaving −2; re-adding puts 6 back and lands on 4 again. The stored bonus stays exactly one level's worth short, permanently. Separately, on the class item where `preUpdate` did run, it added a flat `this.value` whenever `system.classLevel` appeared in the change, unsigned and unscaled: `NimbleCharacter#revertLevelUp` writes `classLevel - 1`, so reverting a level *added* HP instead of giving it back, and the reset to level 1 in `validateLevelHistory` moved the bonus by one level rather than the full difference.

## Changes

- `src/models/rules/maxHpBonus.ts` — add `getMaxHpBonusPerLevel(actor)`, totalling the per-level `maxHpBonus` rules across every item on the actor. It counts disabled rules and skips invalid ones, matching what `preCreate` and `afterDelete` actually wrote, so the stored total stays symmetric with them.
- `src/models/rules/maxHpBonus.ts` — remove `MaxHpBonusRule#preUpdate`. It only ever covered class-hosted rules, and leaving it in would double-count them.
- `src/documents/item/class.ts` — in `NimbleClassItem#_preUpdate`, the one hook that is told the class level moved, fold `perLevelTotal * (newLevel - oldLevel)` into the `actorUpdates` object the method already builds. Being signed, it gives the bonus back on a revert and moves the full amount on a multi-level jump.

The new import runs one way only, from the document into the rule module; the rule module still imports no document, and `pnpm circular-deps` reports no violations.

## Test Plan

`src/documents/item/class.test.ts` is new and drives the real `_preUpdate` hook. With only that file applied on top of `dev`:

```
$ npx vitest run src/documents/item/class.test.ts
 FAIL  src/documents/item/class.test.ts > raises the stored HP bonus for a rule carried by a non-class item
 FAIL  src/documents/item/class.test.ts > lowers the stored HP bonus when a level up is reverted
 FAIL  src/documents/item/class.test.ts > scales by the full difference when the level moves more than one step
 FAIL  src/documents/item/class.test.ts > totals every per-level rule on the actor, including the class item's own
 Test Files  1 failed (1)
      Tests  4 failed | 3 passed (7)
```

The three already passing are the controls that must stay green: a flat `perLevel: false` rule, a rule of another type, and an update where the level does not change. With the fix, all seven pass. `src/models/rules/maxHpBonus.test.ts` adds five cases for the helper itself (totalling across items, flat and other-type rules excluded, disabled counted, invalid skipped).

Whole-suite comparison: `pnpm test` on `dev` is 245 files / 3784 tests passing; on this branch it is 246 / 3796 — the twelve added tests, no new failures. `pnpm type-check` is clean, `biome check` is clean on the touched files, and the repo's own pre-push hook (biome check, type-check, full suite) passes.

`tests/integration` needs a licensed Foundry instance, so I could not run it here. Its `maxHpBonus` case exercises a flat `perLevel: false` add and delete, which this change does not touch. To replay the fix by hand in a world: give a level 2 character a feature carrying `{"type": "maxHpBonus", "value": 2, "perLevel": true}`, open Edit Hit Points and note Bonus HP reads 4; level to 3 and it should now read 6, matching the "from rules" total beside it; revert the level up and it should return to 4.

## Related Issues

Fixes #499

## How this was managed

This work was tracked as a single story on a live board imported from this repository's own issues and pull requests — 837 stories in all: [maxHPBonus Rule is not recalculated on Level-Up](https://eastagiletracker.com/projects/469/stories/419217), on the [Nimble for FoundryVTT board](https://eastagiletracker.com/projects/469).

board

If you'd rather not receive contributions like this, reply `no-more-prs` on this pull request and we won't open any further ones on your repositories.

---

**Lawrence W. Sinclair**
CEO / East Agile
[linkedin.com/in/lwsinclair/](https://linkedin.com/in/lwsinclair/)
[eastagile.com](https://eastagile.com)

### DIFF
--- a/src/documents/item/class.test.ts
+++ b/src/documents/item/class.test.ts
@@ -1,0 +1,153 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { NimbleClassItem } from './class.js';
+
+/**
+ * `NimbleClassItem._preUpdate` ends in `super._preUpdate(...)`. The shared `Item` mock has no
+ * such method, so stand one up for the duration of this file rather than reaching into the
+ * shared mocks.
+ */
+const itemPrototype = (globalThis as unknown as { Item: { prototype: Record<string, unknown> } })
+	.Item.prototype;
+
+beforeEach(() => {
+	itemPrototype._preUpdate = vi.fn(async () => undefined);
+});
+
+afterEach(() => {
+	delete itemPrototype._preUpdate;
+	vi.restoreAllMocks();
+});
+
+interface MaxHpBonusSource {
+	value: number;
+	perLevel?: boolean;
+	disabled?: boolean;
+	type?: string;
+}
+
+/** An item carrying rules, in the `Map` shape `NimbleBaseItem#rules` exposes. */
+function createRuleHost(rules: MaxHpBonusSource[]) {
+	return {
+		rules: new Map(
+			rules.map((rule, index) => [
+				`rule-${index}`,
+				{ type: rule.type ?? 'maxHpBonus', value: rule.value, perLevel: rule.perLevel ?? false },
+			]),
+		),
+	};
+}
+
+function createActor(hpBonus: number, items: ReturnType<typeof createRuleHost>[]) {
+	return {
+		classes: {},
+		items,
+		system: {
+			attributes: { hp: { bonus: hpBonus }, hitDice: {} },
+			classData: { levels: [], startingClass: 'shepherd' },
+		},
+		update: vi.fn(async (_data: Record<string, unknown>) => undefined),
+	};
+}
+
+function createClassItem(actor: unknown, classLevel: number) {
+	return {
+		identifier: 'shepherd',
+		isEmbedded: true,
+		parent: actor,
+		rules: new Map(),
+		system: { classLevel, hitDieSize: 8 },
+	};
+}
+
+async function runClassPreUpdate(classItem: unknown, changed: Record<string, unknown>) {
+	return NimbleClassItem.prototype._preUpdate.call(
+		classItem as NimbleClassItem,
+		changed,
+		{} as never,
+		{} as never,
+	);
+}
+
+/** The single `actor.update()` payload the hook builds. */
+function updatePayload(actor: ReturnType<typeof createActor>): Record<string, unknown> {
+	return actor.update.mock.calls.at(-1)?.[0] ?? {};
+}
+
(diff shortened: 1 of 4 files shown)


Fronix:
So the fix is solid and appreciated. However, due to [Foundry's AI policy](https://foundryvtt.com/article/ai-policy/) we cannot accept contributions that are fully autonomous PRs, a person has to be the one opening the PR.

You are more than welcome to reopen the PR with your own account.

Fronix:
no-more-prs

Issue #640, opened 2026-04-08
[Automation] Shadowmancer - Pilfered Power: limited-use spellcasting resource


Fronix:
## Context
Shadowmancer Level 2 feature **Pilfered Power**. This is the class's primary tiered-spellcasting resource and replaces a conventional mana pool. It uses a DEX-capped counter with a punishing overdraw penalty.

Pack file: `packs/classFeatures/core/shadowmancer/shadowmancer-progression/pilfered-power.json`

## Rules from Rulebook

> "Pilfered Power. You may steal power from your patron to cast tiered spells, always casting them at the highest tier you have unlocked. You can do this DEX times before your patron takes notice. Each time you exceed this limit, your patron damages you for half your max HP as recompense. This limit resets when you commune with your patron during a Safe Rest."

## Implementation
- Counter resource with maximum = DEX, recharges on Safe Rest
- Spell casts always execute at the Shadowmancer's highest unlocked tier
- On cast: decrement counter; if already at 0, allow cast but apply half-max-HP damage automatically
- Integrate with Shadowmancer tier progression (L2 tier 1, L5 tier 2, L8 tier 3, L10 tier 4, L13 tier 5, L16 tier 6, L19 tier 7 — infer from pack progression)
- Hook for features that refund/bypass Pilfered Power: Hungering Shadows, Heart of Burning Fire, We'll ALL Burn!, Shadowfrost, Shadow Exploit, My Blood My Power

## Rule Types
- `resourceCounter` (DEX-scaled, Safe Rest recharge, overdraw penalty)
- `triggerOnEvent` (on-cast hook, on-overdraw hook)

## Blocked By
- #576 resourceCounter
- #584 triggerOnEvent

⚠️ **This is a flagship feature.** Multiple invocations and all three subclasses modify or refund this resource — must land before those can proceed.

Part of #365

PR #919, opened 2026-08-30
fix(mana): seed current mana when max mana first becomes non-zero

Fronix:
## Summary

A character gets no starting mana when their class first grants it. The maximum
goes above zero at level 2, the current value stays at zero, and nothing fills it
until the first Safe Rest. This fills it the moment the maximum first appears.

## Changes

- New mana seeding hooks. The pre-update hooks record the derived max mana before
  a change, and the post-update hooks compare and fill when it first goes above zero.
- Only fills a pool that is empty and whose max was zero before this change. A partly
  spent pool is left alone. Note the predicate is "max was zero", not "never been
  filled": a character who levels to 2, spends mana, levels down, and levels up again
  is seeded a second time, because levelling down returns the max to zero without
  clamping the current value. Low impact, and left as is rather than adding a
  persistent flag.
- The stashed max is keyed by actor uuid. One options object is shared by every document
  in a batched update, and the backend runs the whole pre-hook loop before any
  post-hook, so a single slot left every actor reading the last one's value. Uuid rather
  than id, because an unlinked token's synthetic actor carries the base actor's id.
- Level down clamps the current value. Max mana falls back to zero with the level and
  the current value did not follow it, which the seed makes reachable: a pool filled to
  2 of 2 kept its 2 while the max returned to 0, and readouts show that as a 2/0 chip
  since they only bail when both are zero.
- Only the client that made the change writes. A non-owner never attempts a write
  the server would reject.
- Registered in `nimble.ts` next to the charge system hooks.

This is broader than the title says. The hooks watch max mana, not levelling, so any
change that raises the max from zero on an empty pool seeds it. In practice that means
the Configure Mana dialog too: raising Base Mana from zero fills the pool. That seems
right, and it is covered by a test.

## Test Plan

1. Make a level 1 character of a class that gains mana at level 2 (Mage, Songweaver,
   Shepherd, Stormshifter, or Oathsworn).
2. Level them to 2. The mana bar should be full, not 0 out of max.
3. Spend some mana, then level again. The current value should not move.
4. Have a player who does not own the character watch their console while the GM
   levels them. There should be no permission errors.

Covered by 15 unit tests over the seeding function, plus a live integration test at
`tests/integration/manaSeeding.test.ts` that levels a real Mage up and back down inside
a running world.

Also run in a live Foundry v14 world (14.365), driving the real controls:

- A level 1 Mage levelled to 2 through the sheet's Level Up button and the level-up
  window (Take Average, spend the skill point, Submit) ended on 2 of 2 mana.
- Raising Base Mana from 0 to 7 in the Configure Mana dialog filled the pool to 7 of 7.
- A batched `Actor.updateDocuments` over two actors seeded only the one that gained a
  pool. There is no UI that performs a batched multi-actor update, so this path was
  exercised through the API rather than the interface.
- No console or page errors throughout.

Both fixes were run against a build without them, to confirm the tests catch what they
claim. The batch case left the actor gaining a pool at 0 of 5, because the other actor's
previous max was the one every post-hook read. The level-down case stranded 2 of 0.

## Checklist

## Related Issues

### DIFF
--- a/src/hooks/manaSeeding.test.ts
+++ b/src/hooks/manaSeeding.test.ts
@@ -223,4 +223,53 @@ describe('registerManaSeedingHooks', () => {
 
 		expect(actor.update).toHaveBeenCalledTimes(1);
 	});
+
+	it('clamps the current value down when max mana falls away entirely', async () => {
+		const callbacks = await registerHooks();
+		const actor = createManaActor({ current: 2, max: 2 });
+
+		await updateActor(callbacks, actor, { newMax: 0 });
+
+		expect(actor.update).toHaveBeenCalledWith({ 'system.resources.mana.current': 0 });
+	});
+
+	it('clamps the current value down to a max that fell but stayed positive', async () => {
+		const callbacks = await registerHooks();
+		const actor = createManaActor({ current: 5, max: 5 });
+
+		await updateActor(callbacks, actor, { newMax: 3 });
+
+		expect(actor.update).toHaveBeenCalledWith({ 'system.resources.mana.current': 3 });
+	});
+
+	it('leaves a current value that still fits inside a reduced max alone', async () => {
+		const callbacks = await registerHooks();
+		const actor = createManaActor({ current: 1, max: 5 });
+
+		await updateActor(callbacks, actor, { newMax: 3 });
+
+		expect(actor.update).not.toHaveBeenCalled();
+	});
+
+	it('does not re-clamp from the write the clamp itself performs', async () => {
+		const callbacks = await registerHooks();
+		const actor = createManaActor({ current: 2, max: 2 });
+
+		await updateActor(callbacks, actor, { newMax: 0 });
+		expect(actor.update).toHaveBeenCalledTimes(1);
+
+		actor.system.resources.mana.current = 0;
+		await updateActor(callbacks, actor, { newMax: 0 });
+
+		expect(actor.update).toHaveBeenCalledTimes(1);
+	});
+
+	it('does not clamp when the update was initiated by a different user', async () => {
+		const callbacks = await registerHooks();
+		const actor = createManaActor({ current: 2, max: 2 });
+
+		await updateActor(callbacks, actor, { newMax: 0, userId: 'someone-else' });
+
+		expect(actor.update).not.toHaveBeenCalled();
+	});
 });
--- a/src/hooks/manaSeeding.test.ts
+++ b/src/hooks/manaSeeding.test.ts
@@ -13,6 +13,7 @@ function globals() {
 
 type MockManaActor = {
 	id: string;
+	uuid: string;
 	type: string;
 	system: { resources: { mana: { current: number; max: number } } };
 	update: ReturnType<typeof vi.fn>;
@@ -25,10 +26,13 @@ function createManaActor({
 	max = 0,
 	type = 'character',
 	id = '',
+	uuid = '',
 } = {}): MockManaActor {
 	nextActorId += 1;
+	const actorId = id || `actor-${nextActorId}`;
 	return {
-		id: id || `actor-${nextActorId}`,
+		id: actorId,
+		uuid: uuid || `Actor.${actorId}`,
 		type,
 		system: { resources: { mana: { current, max } } },
 		update: vi.fn().mockResolvedValue(undefined),
@@ -209,6 +213,51 @@ describe('registerManaSeedingHooks', () => {
 
 		expect(gainsAPool.update).toHaveBeenCalledWith({ 'system.resources.mana.current': 5 });
 		expect(alreadyHasOne.update).not.toHaveBeenCalled();
+	});
+
+	// An unlinked token's synthetic actor carries the base actor's id, so two of
+	// them in one batch would collide on an id-keyed stash. Their uuids differ.
+	it('keeps unlinked token actors sharing a base id separate', async () => {
+		const callbacks = await registerHooks();
+		const gainsAPool = createManaActor({
+			current: 0,
+			max: 0,
+			id: 'base-actor',
+			uuid: 'Scene.s1.Token.t1.Actor.base-actor',
+		});
+		const alreadyHasOne = createManaActor({
+			current: 0,
+			max: 4,
+			id: 'base-actor',
+			uuid: 'Scene.s1.Token.t2.Actor.base-actor',
+		});
+		const options = {};
+
+		callbacks.get('preUpdateActor')?.(gainsAPool, {}, options, 'user-1');
+		callbacks.get('preUpdateActor')?.(alreadyHasOne, {}, options, 'user-1');
+
+		gainsAPool.system.resources.mana.max = 5;
+		alreadyHasOne.system.resources.mana.max = 7;
+
+		callbacks.get('updateActor')?.(gainsAPool, {}, options, 'user-1');
+		callbacks.get('updateActor')?.(alreadyHasOne, {}, options, 'user-1');
+		await flushAsync();
+
+		expect(gainsAPool.update).toHaveBeenCalledWith({ 'system.resources.mana.current': 5 });
+		expect(alreadyHasOne.update).not.toHaveBeenCalled();
+	});
+
+	// Nothing stashed means the pre-hook never ran for this actor, so there is
+	// no before-value to compare against and the post-hook must do nothing.
+	it('does nothing when the post-hook runs without a stashed previous max', async () => {
+		const callbacks = await registerHooks();
+		const actor = createManaActor({ current: 0, max: 0 });
+
+		actor.system.resources.mana.max = 5;
+		callbacks.get('updateActor')?.(actor, {}, {}, 'user-1');
+		await flushAsync();
+
+		expect(actor.update).not.toHaveBeenCalled();
 	});
 
 	it('does not re-seed from the write the seed itself performs', async () => {
--- a/src/hooks/manaSeeding.test.ts
+++ b/src/hooks/manaSeeding.test.ts
@@ -48,23 +48,6 @@ async function registerHooks(): Promise<HookCallbacks> {
 	return callbacks;
 }
 
-async function updateClassItem(
-	callbacks: HookCallbacks,
-	actor: MockManaActor,
-	{
-		newMax,
-		itemType = 'class',
-		userId = 'user-1',
-	}: { newMax: number; itemType?: string; userId?: string },
-): Promise<void> {
-	const item = { type: itemType, actor };
-	const options = {};
-	callbacks.get('preUpdateItem')?.(item, {}, options, userId);
-	actor.system.resources.mana.max = newMax;
-	callbacks.get('updateItem')?.(item, {}, options, userId);
-	await flushAsync();
-}
-
 async function updateActor(
 	callbacks: HookCallbacks,
 	actor: MockManaActor,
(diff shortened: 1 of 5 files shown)


src/hooks/manaSeeding.test.ts
```diff
@@ -0,0 +1,181 @@
...
+	{
+		newMax,
+		itemType = 'class',
+		userId = 'user-1',
+	}: { newMax: number; itemType?: string; userId?: string },
+): Promise<void> {
+	const item = { type: itemType, actor };
+	const options = {};
+	callbacks.get('preUpdateItem')?.(item, {}, options, userId);
+	actor.system.resources.mana.max = newMax;
```
trevlar: The tests manually update `actor.system.resources.mana.max` between the pre- and post-update hooks, so they verify the intended design rather than Foundry’s actual behavior.

The fix depends on two runtime details that a plain-object mock can’t confirm: that Foundry recalculates derived data before the `updateActor` hook runs, and that your stashed data survives the socket round trip through `options`.

The first is probably fine since `combatantHooks/combatantHealthStateSync.ts:107` already reads derived HP from an `updateActor` hook, but it still needs to be verified. Run the test plan in a Foundry world and check off the Foundry test. If it passes, this should be ready to merge.

src/hooks/manaSeeding.test.ts
```diff
@@ -0,0 +1,181 @@
...
+}
+
+async function updateActor(
+	callbacks: HookCallbacks,
+	actor: MockManaActor,
+	{ newMax, userId = 'user-1' }: { newMax: number; userId?: string },
+): Promise<void> {
+	const options = {};
+	callbacks.get('preUpdateActor')?.(actor, {}, options, userId);
+	actor.system.resources.mana.max = newMax;
```
trevlar, line 58: The PR description says the pool is only filled if it “has never been filled,” but the code only knows that its max was 0 before the current operation.

For example, a character can level up to 2, spend the resource, level down, and then level up again to refill it for free. `revertLastLevelUp` (`character.ts:1735`) reduces `classData.levels`, returning max to `baseMax` of 0 without clamping the current value. The next level-up then sees both the previous max and current value at 0 and fills the pool again.

This is low-impact, but either update the PR wording to reflect the actual behavior or store a flag indicating whether the pool has already been initialized.

src/hooks/manaSeeding.ts
```diff
@@ -0,0 +1,92 @@
+import { SYSTEM_ID } from '#system';
+
+const PREVIOUS_MANA_MAX_PATH = `${SYSTEM_ID}.previousManaMax`;
```
trevlar, line 3: The stashed value isn’t scoped to an actor. A batched update affecting multiple characters, such as a module calling `Actor.updateDocuments`, can share the same `options` object. The last actor’s max would then overwrite the value used for every actor in the batch.

Key the stashed value by actor ID to prevent that.

src/hooks/manaSeeding.ts
```diff
@@ -0,0 +1,92 @@
...
+
+	const snapshot = getManaSnapshot(actor);
+	if (!snapshot || !options || typeof options !== 'object') return;
+
+	const previousMax = foundry.utils.getProperty(options, PREVIOUS_MANA_MAX_PATH);
+	if (typeof previousMax !== 'number' || previousMax > 0) return;
+
+	// Seed only on the transition from "no pool" to "has a pool", and never
+	// touch a value that already exists.
+	if (snapshot.max <= 0 || snapshot.current !== 0) return;
```
trevlar: This guard applies to more than level-ups. The character sheet’s max mana field (`PlayerCharacterSheet.state.svelte.ts:299`, `updateMaxMana`) updates only `baseMax`, so increasing max mana from 0 also fills the current mana pool.

That behavior seems reasonable, but it’s broader than the PR title suggests and isn’t covered by a test. Please mention it in the PR description and add a test case for it.

trevlar (changes requested):
**🔧 Push back — the design holds up on a code trace, but the one assumption it rests on is the one thing your tests can't check**

I traced the level 1 → 2 path against `character.ts` and the seeding logic is correct: `_prepareMaxMana` (`character.ts:487`) returns `baseMax` while `levels.character === 1`, and `levels` comes from `system.classData.levels`, which only moves in the *second* write of `triggerLevelUp` (`character.ts:1463`). So max stays 0 through `updateItem` and the seed has to land on `updateActor`. That's exactly what your hooks do. Multiclass is covered too — `NimbleClassItem._preCreate` pushes to `classData.levels` before the item exists, so max only jumps on `createItem`, which you handle.

**What's good**
- The `previousMax > 0` guard doubles as the recursion brake — the seed's own `actor.update` re-enters the hook and bails, and you have a test pinning that.
- Correct read on the ownership problem: gating on `userId` rather than `isOwner` means exactly one client writes, and it's the one that already had permission.

---

1. Run the test plan in a live world, tick the box
2. Reconcile "never been filled" wording with what the code checks
3. Scope the stashed max by actor id
4. Document the sheet max-mana auto-fill side effect

trevlar (changes requested):
**🔧 Push back — the seed strands mana on level down, which is now visible where it wasn't before**

The core fix is right and the investigation behind it is better than most. One follow-on state needs closing before merge.

**Must fix before merge**

- **src/documents/actor/character.ts:1735** — `triggerLevelDown` never clamps `system.resources.mana.current`, so a level-2 caster seeded to `2/2` who levels back down keeps `current: 2` while derived max returns to `0`. That state was unreachable before your change (current was always 0), so this PR is what makes it visible: `getActorManaValueAndMax` only bails when *both* are `<= 0` (`src/utils/actorResources.ts:71`), so the combat tracker renders a `2/0` mana chip on a level-1 character (`src/view/ui/ctTopTracker/resources.utils.ts:320`). Add the same clamp `configureMana` already does at `character.ts:772` — push `'system.resources.mana.current'` into `actorUpdates` when current exceeds the post-level-down max. Worth a test pinning `2/0` can't happen. Note this doesn't change the level-down/level-up refill you documented and accepted; it just stops the stranded display.

- **src/hooks/manaSeeding.ts:28** — you key the stash by `typedActor.id` specifically to disambiguate actors inside one batched update, but unlinked token actors share the base actor's `id`. `typedActor.uuid` is unique per synthetic actor and is a drop-in swap that makes the key actually do what the comment above it promises.
- **src/hooks/manaSeeding.test.ts:225** — the `typeof previousMax !== 'number'` guard at `manaSeeding.ts:71` is the one branch in the new logic with no test. A post-hook firing without its pre-hook (another module short-circuiting, a hook registered mid-operation) should be a silent no-op — cheap to pin.

**What's good**

- Keying the stash by actor id, then proving it with the batched-update test at `manaSeeding.test.ts:194`, is the kind of thing that normally ships broken. Note `base.svelte.ts:1137` has the same single-slot shape for `previousHp` and presumably the same latent bug — separate PR, but you found it.
- The `game.user?.id !== userId` gate is the right call over a GM-only or owner check: no client attempts a write the server would reject, and exactly one client writes.

---

1. Clamp current mana on level down; add a test
2. Key the stash by `uuid`, not `id`
3. Test the missing-stash no-op branch

trevlar (changes requested):
**🔧 Push back — the seed is solid, but the down-clamp is parked in the one place it can't be tested and only covers one of three paths**

**Must fix before merge**

- **`src/documents/actor/character.ts:1746`** — the clamp lives inside `revertLastLevelUp`, so it only covers that one caller. Two other paths drop max mana the same way and still strand a value above it:
  - `validateLevelHistory` (`character.ts:1220`) slices `system.classData.levels` back to one entry when history is inconsistent. `levels.character` goes to 1, `_prepareMaxMana` returns `baseMax` (0), and `current` is never touched.
  - `NimbleClassItem._onDelete` (`src/documents/item/class.ts:312`) filters the identifier out of `classData.levels`. Same drop to `baseMax`, same untouched `current`.

  Both were technically broken before, but your seed is what makes them *visible*: a pool that used to sit at 0/0 until the first Safe Rest now sits at 2/2, so reverting through either of these leaves the same `2/0` chip your PR body calls out for level-down. Move the clamp into `manaSeeding.ts` as the mirror of the seed — the hook already sees every actor update with the pre-update max stashed, so `current > max` is one more branch in `seedManaIfNewlyAvailable`, and it covers all three sites at once.

  It also fixes the coverage hole: right now the clamp's only test is `tests/integration/manaSeeding.test.ts:92`, and `tests/integration/README.md` says that suite is local-only and never runs in CI. New logic in `character.ts` with nothing in `pnpm check` guarding it. In the hook it's a two-line unit test alongside the 15 you already wrote.

- **`src/hooks/manaSeeding.ts:91`** — I don't think the four class-item registrations can ever fire a seed. `_prepareMaxMana` (`character.ts:486-488`) derives max from the *actor's* `classData.levels`, not the item's `classLevel`, and `NimbleClassItem._preCreate` (`class.ts:170`) already awaits `actor.update({'system.classData.levels': ...})` before the item lands — so `preUpdateActor`/`updateActor` always sees the 0→N transition first, and by the time `createItem` runs, `current !== 0` blocks it. On the update side, a `classLevel` bump only moves max when `levels.character > 1`, which means `previousMax > 0` and the guard rejects it. Your unit tests set `actor.system.resources.mana.max` by hand between the two callbacks, so they pin the predicate but say nothing about whether the item hooks are reachable. AGENTS.md's YAGNI rule is explicit about paths with no current caller — worth confirming against a live multiclass before keeping them.

**What's good**

- The per-uuid stash and the batched-update test are the right call, and the comment explaining *why* uuid rather than id (synthetic token actors sharing a base id) is the kind of comment worth keeping.
- Verifying both fixes against a build without them, and reporting exactly how each failed (0 of 5, 2 of 0), is the part of the test plan that actually earns trust.

---

1. Move the down-clamp into the mana seeding hook.
2. Cover `validateLevelHistory` and class deletion with it.
3. Add a unit test for the clamp that runs in CI.
4. Confirm the class-item hooks are reachable, or drop them.

Fronix:
## 1 and 2. The clamp moved into the hook and covers all three paths

I reproduced both stranding paths on a live V14 world, running the branch as it stood. Starting from a level 2 Mage seeded to 2/2:

```
[history reset] wasReset = true | after = 2 / 0 | charLevel = 1
[class delete]  after = 2 / 0
```

Same 2/0 you called, from both sites.

The clamp is now a second branch in the update hook next to the seed, and the copy in `revertLastLevelUp` is gone. Every path that moves max mana writes `system.classData.levels` through an actor update, so that one branch catches level down, the history reset and class deletion together.

## 3. The clamp is under `pnpm test` now

No CI cover, as you said. `run-tests.yml` runs `pnpm test`, which is the `vite.config.mts` project matching `src/**/*.{test,spec}.{js,ts}`. The integration config says local only, in those words.

Five new unit tests in `src/hooks/manaSeeding.test.ts`, 17 in the file now. I wanted them to mean something rather than just pass, so I ran them against the pre-fix hook. The three positive ones go red:

```
× clamps the current value down when max mana falls away entirely
× clamps the current value down to a max that fell but stayed positive
× does not re-clamp from the write the clamp itself performs
```

The live suite grew two tests as well, one for the history reset and one for the class deletion.

## 4. The class item hooks were unreachable, so they are gone

This is where I spent the most time, because I thought I had a counterexample. A martial first multiclass looked like it would slip past the actor hooks. `_preCreate` pushes to `classData.levels` and awaits `actor.update`, but at that moment `classes` still holds only the Berserker, whose mana formula is empty, so max is still 0 at `updateActor`. Max would only rise once the Mage item existed, which makes `createItem` the one hook that could seed. Foundry's ordering backs that up. `parent?.reset()` runs before the `create<Type>` hook fires, so derived data is fresh by then.

Line 123 of `NimbleClassItem._preCreate` kills it:

```ts
if (Object.keys(actor.classes).length) return false;
```

It refuses a second class outright. Importing Mage onto a level 1 Berserker in a live world:

```
[before] berserker mana = 0 / 0 | charLevel = 1
[after]  mana = 0 / 0 | charLevel = 1 | classes = [ 'berserker' ]
```

That closes the only route I could find. `levels.character` is just `classData.levels.length`, so max mana can only cross zero on an actor update. I dropped all four item registrations and the two tests that only looked like they covered them.

One thing to flag for later. Those hooks would be needed again if multiclassing is ever unblocked, for the reason above. I put that in the commit message instead of leaving a stub behind.

trevlar (approved):

PR #920, opened 2026-08-30
fix(spells): show spell costs and make the cast dialog obey the automation setting

Fronix:
## Summary

Three small fixes to spell costs. Two view components read a field the spell
schema does not have, so no cost has ever rendered. The cast dialog blocks on
insufficient mana whether or not resource spending automation is on, while the
rest of the cast path obeys the setting. A new test catches a renumbered
migration that would otherwise never run.

## Changes

- The spell list and the level-up card read `system.tier` instead of `system.manaCost`,
  which does not exist. Costs now render.
- The cast dialog obeys the resource spending automation setting. With it off,
  nothing is blocked and nothing is deducted. Costs stay visible either way.
- Tier bounds are still enforced whatever the setting says. Turning automation off
  does not let you cast above your unlocked tier.
- New test asserting `LATEST_SCHEMA_VERSION` equals the highest registered migration
  version, and that no two migrations share a version.

## Test Plan

1. Open a Mage's sheet and go to Heroic Actions, then Cast a Spell. Each tiered
   spell should show its cost in mana. Cantrips should show none.
2. Level up a caster. The spell cards in the level-up dialog should show costs too.
3. With resource spending automation on, try to cast a spell you cannot afford.
   The dialog should refuse it.
4. Turn resource spending automation off and try again. The dialog should let it
   through and deduct nothing.
5. With automation off, try to cast above your highest unlocked tier. It should
   still be refused.

## Checklist

## Related Issues

## Summary by CodeRabbit

* **New Features**
  * Spell upcasting now respects the Resource Spending automation setting.
  * When automation is disabled, spells can be upcast to the highest unlocked tier without current mana limiting the selection or blocking the cast.
  * An actor with no spell tier ladder, which is every NPC and monster, casts at the tier of the spell itself and cannot upcast.
  * Spell mana costs are now calculated from spell tier, with cantrips costing zero.

* **Documentation**
  * Updated guidance for Resource Spending automation and spell upcasting.


src/view/dialogs/components/levelUpHelper/LevelUpSpellCard.svelte.ts
trevlar, line 1: This is the same `manaCost` → `tier` fix that now has test coverage on the `CastSpellActionPanel` side, but there’s nothing covering it here. The bug is exactly the kind that can fail silently: a field that doesn’t exist gets read, the UI falls back to 0 or nothing, and nobody notices.

src/view/dialogs/components/levelUpHelper/LevelUpSpellCard.svelte.ts
```diff
@@ -66,8 +61,8 @@ function extractDisplayData(system: SpellSystemDataWithMana): SpellDisplayData {
 		});
 	}
 
-	// Mana cost
-	const manaCost = system.manaCost ?? 0;
+	// Tiered spells cost their tier in mana; cantrips are free.
+	const manaCost = system.tier ?? 0;
```
trevlar: ou fixed `manaCost` → `tier` in two places, but only one of them has test coverage. This one can fail silently in exactly the same way as the original bug: read a field that doesn’t exist, end up with 0, and ship it. `LevelUpFeatureOptionPicker.svelte.test.ts` in the same directory already mocks `fromUuid` around a rune-based state module. I’d follow that pattern and assert that `createLevelUpSpellCardState` returns `manaCost: 3` for a tier-3 spell and `0` for a cantrip.

src/view/sheets/components/CastSpellActionPanel.svelte.test.ts
trevlar: `createSpellItem` still sets `manaCost: 0` alongside `tier: 1`. Nothing reads `manaCost` anymore, and the fixture now contradicts what the test actually renders: 1 mana for that spell. I’d just remove the stale field.

src/view/dialogs/SpellUpcastDialog.svelte
```diff
@@ -33,12 +34,16 @@
...
+	// mana neither bounds the slider nor blocks the cast.
+	const enforceManaCost = isResourceSpendingAutomationEnabled();
+
 	// Compute upcast constraints (safe for NPCs/Monsters that lack resources)
 	const baseMana = $derived(spell.tier);
 	const resources = $derived(actor?.system?.resources);
 	const currentMana = $derived(resources?.mana?.current ?? 0);
 	const maxTier = $derived(resources?.highestUnlockedSpellTier ?? 9);
-	const maxMana = $derived(Math.min(currentMana, maxTier));
+	const maxMana = $derived(enforceManaCost ? Math.min(currentMana, maxTier) : maxTier);
```
trevlar: With the toggle off, the `?? 9` fallback for `maxTier` becomes active for actors with no `resources` at all. Only `CharacterDataModel` defines `mana` and `highestUnlockedSpellTier`, so an NPC or monster with a tiered spell can now get an upcast slider all the way to tier 9. Previously, `Math.min(currentMana, maxTier)` effectively kept that at 0. If that behavior isn’t intentional, I’d fall back to `spell.tier` instead of 9 when `resources` is missing.

trevlar (changes requested):
**🔧 Push back: the level-up half of the fix is still untested**

The diagnosis is correct, and I verified it. `SpellDataModel` doesn’t have a `manaCost` field, only `tier`, so both view components were reading `undefined` and rendering nothing. Deriving the cost from `tier` is also consistent with what `SpellScrollDialog.state.svelte.ts` and `spell.ts` already do.

**What looks good**

* The UI design is intuitive and easy to use.
* Keeping the tier ceiling enforced when automation is off makes sense, and the settings docs now explain that clearly. Tier gating is a capability limit, not part of resource spending.
* `migrationVersion.test.ts` checks the highest registered *version*, which the existing CI file-count check can’t catch if a migration gets renumbered. Those checks are complementary, not duplicates.

**Requested changes**

1. Add coverage for the level-up card mana cost fix.
2. Remove the stale `manaCost` field from the spell fixture.
3. Bound NPC upcasting when the actor has no resources.

src/view/dialogs/SpellUpcastDialog.svelte
```diff
@@ -33,12 +34,17 @@
...
+
 	const baseMana = $derived(spell.tier);
 	const resources = $derived(actor?.system?.resources);
 	const currentMana = $derived(resources?.mana?.current ?? 0);
-	const maxTier = $derived(resources?.highestUnlockedSpellTier ?? 9);
-	const maxMana = $derived(Math.min(currentMana, maxTier));
+	// An actor with no resources node (an NPC or monster) has no tier ladder to
+	// climb, so it casts at the spell's own tier.
+	const maxTier = $derived(resources ? (resources.highestUnlockedSpellTier ?? 9) : spell.tier);
+	const maxMana = $derived(enforceManaCost ? Math.min(currentMana, maxTier) : maxTier);
```
trevlar: The new `maxMana` / `maxTier` branches have no tests. `applyUpcastDeltas.test.ts` covers the validator, but the dialog re-implements the same rule for the slider bound and the submit guard at line 306, and nothing exercises it. Pull the bound into a pure helper (`computeUpcastManaBounds({ spellTier, resources, enforceManaCost })` under `src/utils/spell/`) and test it, or drive the component through a harness the way `tests/harnesses/SpellScrollDialogStateHarness.svelte` does for the sibling dialog.

src/view/dialogs/SpellUpcastDialog.svelte
```diff
@@ -33,12 +34,17 @@
...
+	const enforceManaCost = isResourceSpendingAutomationEnabled();
+
 	const baseMana = $derived(spell.tier);
 	const resources = $derived(actor?.system?.resources);
 	const currentMana = $derived(resources?.mana?.current ?? 0);
-	const maxTier = $derived(resources?.highestUnlockedSpellTier ?? 9);
-	const maxMana = $derived(Math.min(currentMana, maxTier));
+	// An actor with no resources node (an NPC or monster) has no tier ladder to
+	// climb, so it casts at the spell's own tier.
+	const maxTier = $derived(resources ? (resources.highestUnlockedSpellTier ?? 9) : spell.tier);
```
trevlar: For an actor with no `resources` node, `maxTier` is now `spell.tier`, so `maxMana === baseMana` and `canUpcast` is false. NPCs and monsters cannot upcast at all, automation on or off. That matches `dev` today, so it is not a regression, but your PR body says NPCs "can upcast according to their spell tier," and the code says the opposite. Confirm which you meant — note `ItemActivationManager.ts:154` still falls back to `9` for the same actors, so the two layers disagree.

src/view/sheets/components/CastSpellActionPanel.svelte.ts
```diff
@@ -91,8 +91,9 @@ export function createSpellPanelState(
 		return null;
 	}
 
+	// Tiered spells cost their tier in mana; cantrips are free.
 	function getSpellManaCost(spell: Item): number {
-		return getSystemData(spell).manaCost ?? 0;
+		return getSystemData(spell).tier ?? 0;
```
trevlar: tier-as-mana-cost is now written out in three files across three features (here, `LevelUpSpellCard.svelte.ts:65`, `SpellScrollDialog.state.svelte.ts:66`), with the same comment copied and pasted twice. AGENTS.md's Rule of Three says that this is the point to extract a shared `getSpellManaCost(system)`.

src/view/dialogs/components/levelUpHelper/LevelUpSpellCard.svelte.test.ts
```diff
@@ -0,0 +1,54 @@
+import { render, waitFor } from '@testing-library/svelte';
+import { afterEach, describe, expect, it } from 'vitest';
+
+import type { SpellIndexEntry } from '#utils/getSpells.js';
+import LevelUpSpellCardStateHarness from '../../../../../tests/harnesses/LevelUpSpellCardStateHarness.svelte';
+
+/** Back `fromUuid` with one in-memory spell document. Returns a restore function. */
+function stubSpellDocument(uuid: string, system: Record<string, unknown>): () => void {
```
trevlar: `stubSpellDocument` hand-rolls the save/restore of `globalThis.fromUuid`. The repo uses `vi.stubGlobal('fromUuid', ...)` for this (`ConditionNode.test.ts:69`, `combatManaRules.test.ts:198`), which cleans itself up.

trevlar (changes requested):
**🔧 Push back — the dialog logic this PR is named for is the one part with no test**

The core fix is right. `manaCost` genuinely does not exist on `SpellDataModel`, `tier` is already the cost of record everywhere else (`spell.ts:70` deducts it, `applyUpcastDeltas.ts:82` uses it as `baseMana`, `SpellScrollDialog.state.svelte.ts:66` labels it), and gating on `isResourceSpendingAutomationEnabled()` matches the gate `spell.ts:68` already applies to the deduction. I reverted `LevelUpSpellCard.svelte.ts:65` to `system.manaCost` and your new test failed, so that one is genuinely pinned.

**What's good**

- The `enforceManaCost` default of `true` keeps every existing caller of `validateAndComputeUpcast` on the old behavior, and the three new validator tests pin exactly the right seam: mana skipped, tier ceiling and base cost still enforced.

---

1. Test the dialog's new mana bounds.
2. Confirm NPCs are meant to lose upcasting.
3. Extract a shared `getSpellManaCost` helper.
4. Use `vi.stubGlobal` for `fromUuid`.

Fronix:
Round 2 addressed.

- The dialog and the manager now take their bounds from one helper, `computeUpcastBounds` (`src/utils/spell/computeUpcastBounds.ts`). No `resources` means `maxTier = spell.tier` in both places, and the tier 9 fallback is gone from both. The PR body said the opposite and is corrected: an actor with no ladder casts at the spell's own tier and cannot upcast.
- The helper is unit tested: automation on with low mana, automation off, no resources, a null ladder, and a cantrip. The slider bound and the submit guard read the same values.
- `getSpellManaCost` moved to `src/utils/spell/getSpellManaCost.ts` and is used at all three sites; its tests moved next to it and the duplicated comment is gone.
- `vi.stubGlobal` in the level-up card test, and the migration version test now asserts every version is a number.
- Rebased onto dev, so #933 and #946 are in.

One thing I left alone: with Resource Spending on, an actor with no `resources` has 0 current mana, so the dialog's existing "not enough mana" guard still blocks a base tier cast for it. That predates this PR and is out of scope here.

src/migration/migrationVersion.test.ts
```diff
@@ -0,0 +1,20 @@
+import { describe, expect, it } from 'vitest';
+import { MigrationList } from './MigrationList.js';
+import { MigrationRunnerBase } from './MigrationRunnerBase.js';
+
+// The version gate reads the hand-maintained constant, not the computed list
+// version. If a migration is renumbered without updating the constant, the
+// migration silently never runs. This test turns that silence into a failure.
+describe('migration schema version', () => {
```
trevlar: This file duplicates `src/migration/MigrationList.test.ts`, which is already on `dev` and asserts both of these things: `LATEST_SCHEMA_VERSION === MigrationList.latestVersion`, and that no two migrations share a version. Both suites run and pass side by side right now, and nothing in #921, #922, #925 or #926 removes either one. Delete this file. If the `typeof version === 'number'` assertion is worth keeping, move it into the existing one.

src/view/dialogs/SpellUpcastDialog.svelte
trevlar: You're editing this handler anyway, so it's a good moment to drop the `console.log`. Its sibling at `src/documents/dialogs/SpellUpcastDialog.svelte.ts:74` logs every submission too, and #922 leaves both in place.

trevlar (dismissed):
**🔧 Push back — one duplicate test file; the rest is the stack's job**

**Must fix before merge**

- **src/migration/migrationVersion.test.ts:8** — This file duplicates `src/migration/MigrationList.test.ts`, which is already on `dev` and asserts both of these things: `LATEST_SCHEMA_VERSION === MigrationList.latestVersion`, and that no two migrations share a version. Both suites run and pass side by side right now, and nothing in #921, #922, #925 or #926 removes either one. Delete this file. If the `typeof version === 'number'` assertion is worth keeping, move it into the existing one.

- **src/view/dialogs/SpellUpcastDialog.svelte:292** — You're editing this handler anyway, so it's a good moment to drop the `console.log`. Its sibling at `src/documents/dialogs/SpellUpcastDialog.svelte.ts:74` logs every submission too, and #922 leaves both in place.

**What's good**

- `computeUpcastBounds` is the right extraction, and it shows: #922 extends it with `flatCost` and `pinnedCastTier` by adding two lines to one function instead of patching the dialog and the activation manager separately.
- Killing the phantom `system.manaCost` field and routing all three cost sites through the tier is the correct base for the class-declared costs that land in #922.

---

1. Delete the duplicate migration test file
2. Drop the two SpellUpcastDialog console.logs

Fronix:
Both are done.

1. `src/migration/migrationVersion.test.ts` is deleted (c5091821). You are right that `MigrationList.test.ts` on dev already asserts both things. I did not move the `typeof version === 'number'` assertion across: a migration that forgets its version inherits `undefined`, which makes `MigrationList.latestVersion` come out `NaN`, so the existing equality test already fails for that case.
2. Both `[SpellUpcastDialog]` logs are gone (44789a29): the cast button one in the Svelte component and the `submitActivation` one in `SpellUpcastDialog.svelte.ts`. I checked the top of the stack, neither returns in a later branch.

#921, #922, #925 and #926 are rebased on top. `pnpm check` passes on all five branches, and each one merges cleanly with dev.

PR #921, opened 2026-08-30
feat(spells): derive unlocked spell tier from a class's own spell grants

Fronix:
## Summary

One global level table decided which spell tiers every character had
unlocked, and it disagreed with every class in the system, including at the first
rung. It granted tier 1 at level 1 where every class grants it at level 2 or 3.
The table is deleted and the tier now comes from each class's own authored grants.

**This changes which tiers an existing character can reach at some levels.** Please
check a caster of a class you know before approving.

## Changes

- Delete the global spell tier table. The unlocked tier is the highest tier granted
  by a class or subclass feature whose level threshold the character has reached.
- A grant with no level threshold is skipped rather than read as level zero. Two
  features in the packs are affected.
- Whether a threshold is met is decided by the rule's own predicate through the
  rules engine, so a grant gated on more than a level is honoured in full.
- Spellcaster status no longer means holding mana. The spell scroll dialog used to
  warn that a character "has no mana, so this spell would sit on the list
  uncastable", which is wrong for a class that pays another way.
- The manual tier override on the sheet used to freeze at the first computed value.
  Clearing it now returns to derivation.

## Test Plan

1. Open a Mage at level 4, 6, 8 and 10. The unlocked tier should follow the class's
   printed ladder, not the old table.
2. Do the same for a Shadowmancer and a Stormshifter. Subclass grants should count.
3. On a character sheet, set the unlocked tier by hand in Settings. It should hold.
   Clear it. It should go back to deriving.
4. Drop a spell scroll on a caster. No "no mana" warning.
5. Drop a spell scroll on a Berserker. It should still warn, because they cannot
   cast tiered spells.

## Known limits, recorded rather than hidden

- Commander Spellblade unlocks tiers on a schedule independent of what it is granted,
  so derivation is wrong for it. Spellblade spellcasting is not implemented today, so
  this is not a regression, but it is not solved either.
- The Shadowmancer pack grants tier 3 at level 7. Issue #640's table says level 8.
  Both come from the same damaged text extract. Someone with the printed book should
  settle it. It is a one line content fix either way.

## Checklist

## Related Issues

Players and GMs now receive spell tiers from authored spell grants on class, subclass, and feature items. A numeric Highest Unlocked Spell Tier override remains active. A null override resumes automatic derivation. Scrolls now warn when their tier exceeds the character’s unlocked tier, including for casters without mana. Non-casters also receive the warning.

Homebrewers can define spell-tier progression with Grant Spells rules and level thresholds. Mana formulas now create mana pools without making a character a caster. Documentation explains spell grants, tier overrides, reset behavior, and scroll warnings.

Engine changes:
- `getHighestSpellTier` evaluates enabled, applicable grant rules through the predicate engine.
- Thresholdless, invalid, out-of-range, and non-integer tier grants are ignored.
- Nested `$and` and `$or` predicates support level-threshold checks.
- Highest tier selection is limited to integer tiers from 0 through 9.
- Spellcaster status no longer depends on mana ownership.
- Resetting the tier stores `null` and enables re-derivation during later preparation cycles.
- Scroll warning state uses `showTierWarning` and the `tierLockedWarning` localization key.
- Added unit, integration, actor preparation, settings, and scroll warning tests.
- No migrations were added or renumbered.
- No compendium pack content changed. Tests use authored feature-pack grants.
- No new settings were added or changed.
- No new rule types were added. Existing Grant Spells rules now unlock tiers from authored thresholds.
- The `noManaWarning` key was replaced by the `tierLockedWarning` key in `public/lang/en.json`.

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -3479,7 +3479,7 @@
 				"addToSpellListHintCantrip": "Learn it. Cantrips are free to cast.",
 				"addAsScroll": "Add as a spell scroll",
 				"addAsScrollHint": "Single use. No mana, no magical ability needed.",
-				"noManaWarning": "{name} has no mana, so this spell would sit on the list uncastable.",
+				"notACasterWarning": "{name} cannot cast spells of this tier, so it would sit on the list uncastable.",
 				"labelTier": "Tier",
 				"labelValue": "Value",
 				"labelCasting": "Casting",
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -3479,7 +3479,7 @@
 				"addToSpellListHintCantrip": "Learn it. Cantrips are free to cast.",
 				"addAsScroll": "Add as a spell scroll",
 				"addAsScrollHint": "Single use. No mana, no magical ability needed.",
-				"notACasterWarning": "{name} cannot cast spells of this tier, so it would sit on the list uncastable.",
+				"tierLockedWarning": "{name} cannot cast spells of this tier. It would be added to their spell list, but they would not be able to use it.",
 				"labelTier": "Tier",
 				"labelValue": "Value",
 				"labelCasting": "Casting",
--- a/src/documents/actor/character.ts
+++ b/src/documents/actor/character.ts
@@ -271,9 +271,15 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
 		actorData.resources.mana.value = actorData.resources.mana.current;
 		actorData.resources.mana.max = this._prepareMaxMana(actorData);
 
-		// Prepare highest unlocked spell tier (only if not manually set)
-		actorData.resources.highestUnlockedSpellTier ??=
-			this._prepareHighestUnlockedSpellTier(actorData);
+		// Prepare highest unlocked spell tier. A stored number is a manual
+		// override; a stored null means derive. Reading the override from the
+		// source keeps the derivation live — assigning onto the prepared value
+		// and re-checking it would freeze the first computed number in place.
+		const spellTierOverride = (
+			this.system._source as { resources?: { highestUnlockedSpellTier?: number | null } }
+		).resources?.highestUnlockedSpellTier;
+		actorData.resources.highestUnlockedSpellTier =
+			spellTierOverride ?? this._prepareHighestUnlockedSpellTier(actorData);
 
 		// Prepare Inventory Slots
 		const baseInventorySlots = 10 + actorData.abilities.strength.mod;
@@ -525,14 +531,10 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
 		return maxMana;
 	}
 
-	_prepareHighestUnlockedSpellTier(_: NimbleCharacterData): number | null {
-		const classes = Object.values(this.classes ?? {});
-		if (classes.length === 0) return 0;
-		// Check if there are any spellcasting classes - if not, return null
-		const isSpellCaster = this.system.resources.mana.max > 0;
-
-		if (!isSpellCaster) return null;
-		// Only update if value is null e.g not migrated
+	// Derived from the spell grants on the character's class, subclass, and
+	// feature items. A character with no eligible tiered grant — whatever
+	// resource they hold — has no unlocked tier and is not a caster.
+	_prepareHighestUnlockedSpellTier(_: NimbleCharacterData): number {
 		return getHighestSpellTier(this);
 	}
 
--- a/src/documents/actor/character.ts
+++ b/src/documents/actor/character.ts
@@ -273,7 +273,7 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
 
 		// Prepare highest unlocked spell tier. A stored number is a manual
 		// override; a stored null means derive. Reading the override from the
-		// source keeps the derivation live — assigning onto the prepared value
+		// source keeps the derivation live. Assigning onto the prepared value
 		// and re-checking it would freeze the first computed number in place.
 		const spellTierOverride = (
 			this.system._source as { resources?: { highestUnlockedSpellTier?: number | null } }
@@ -532,8 +532,8 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
 	}
 
 	// Derived from the spell grants on the character's class, subclass, and
-	// feature items. A character with no eligible tiered grant — whatever
-	// resource they hold — has no unlocked tier and is not a caster.
+	// feature items. A character with no eligible tiered grant, whatever
+	// resource they hold, has no unlocked tier and is not a caster.
 	_prepareHighestUnlockedSpellTier(_: NimbleCharacterData): number {
 		return getHighestSpellTier(this);
 	}
--- a/src/models/rules/grantSpells.ts
+++ b/src/models/rules/grantSpells.ts
@@ -18,7 +18,7 @@ function schema() {
 				label: 'NIMBLE.rules.grantSpells.schools.label',
 			},
 		),
-		tiers: new fields.ArrayField(new fields.NumberField({ integer: true, min: 0 }), {
+		tiers: new fields.ArrayField(new fields.NumberField({ integer: true, min: 0, max: 9 }), {
 			required: false,
 			nullable: false,
 			initial: [0],
--- a/src/utils/spell/getHighestSpellTier.ts
+++ b/src/utils/spell/getHighestSpellTier.ts
@@ -1,20 +1,76 @@
-/** Minimal interface for actors that have character levels */
-interface ActorWithLevels {
-	levels: { character: number };
+/**
+ * Item types whose spell grants can unlock cast tiers. Spells granted by
+ * other item types (wands, scrolls, backgrounds) do not make a character a
+ * caster and do not raise their unlocked tier.
+ */
+const TIER_GRANTING_ITEM_TYPES = new Set(['class', 'subclass', 'feature']);
+
+interface GrantRuleLike {
+	type?: string;
+	disabled?: boolean;
+	tiers?: unknown;
+	predicate?: unknown;
+}
+
+interface RuleBackedItemLike {
+	type?: string;
+	rules?: Map<string, GrantRuleLike>;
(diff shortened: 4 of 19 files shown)


public/lang/en.json
```diff
@@ -3379,7 +3379,7 @@
 				"addToSpellListHintCantrip": "Learn it. Cantrips are free to cast.",
 				"addAsScroll": "Add as a spell scroll",
 				"addAsScrollHint": "Single use. No mana, no magical ability needed.",
-				"noManaWarning": "{name} has no mana, so this spell would sit on the list uncastable.",
+				"notACasterWarning": "{name} cannot cast spells of this tier, so it would sit on the list uncastable.",
```
trevlar: Update the text to 

`{name} cannot cast spells of this tier. It would be added to their spell list, but they wouldn’t be able to use it.`

src/models/rules/grantSpells.ts
```diff
@@ -18,7 +18,7 @@ function schema() {
 				label: 'NIMBLE.rules.grantSpells.schools.label',
 			},
 		),
-		tiers: new fields.ArrayField(new fields.NumberField({ integer: true, min: 0 }), {
+		tiers: new fields.ArrayField(new fields.NumberField({ integer: true, min: 0, max: 9 }), {
```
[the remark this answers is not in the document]
Fronix: This is fine, in the future we might have no max since it's just the theoretical max of the rules.

src/documents/actor/character.ts
```diff
@@ -265,9 +265,15 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
...
 		actorData.resources.mana.max = this._prepareMaxMana(actorData);
 
-		// Prepare highest unlocked spell tier (only if not manually set)
-		actorData.resources.highestUnlockedSpellTier ??=
-			this._prepareHighestUnlockedSpellTier(actorData);
+		// Prepare highest unlocked spell tier. A stored number is a manual
+		// override; a stored null means derive. Reading the override from the
+		// source keeps the derivation live — assigning onto the prepared value
+		// and re-checking it would freeze the first computed number in place.
+		const spellTierOverride = (
```
trevlar, line 272: Reading the override from `_source` instead of `??=` in prepared data is the fix for the tier freezing at the first computed value, and nothing pins it there. Add an actor-prepare test covering the three states: a stored number wins over derivation, a stored `null` derives, and a second prepare cycle re-derives rather than sticking to the first number. `src/documents/actor/prepareDataAccumulatorReset.test.ts` already stubs a prepare cycle if you want a pattern.

src/utils/spell/getHighestSpellTier.ts
```diff
@@ -1,20 +1,87 @@
...
+function hasLevelThreshold(rule: GrantRuleLike): boolean {
+	const predicate = rule.predicate;
+	if (!predicate || typeof predicate !== 'object') return false;
+
+	const source =
+		'_source' in predicate && predicate._source && typeof predicate._source === 'object'
+			? predicate._source
+			: predicate;
+
+	const level = (source as { level?: { min?: unknown } }).level;
```
trevlar: `hasLevelThreshold` only finds a *top-level* `level.min`. A grant whose level is within `$and`/`$or` is dropped even though `appliesTo()` would pass, contradicting the JSDoc line "a grant gated on more than a level is honored in full." Nothing in `packs/` hits this today, and the Rules Builder can't emit logical keys, so it's a homebrew-only hole, but either walk the logical nodes or narrow the comment to say "a level alongside sibling conditions."

src/view/sheets/pages/PlayerCharacterSettingsTab.svelte
```diff
@@ -229,7 +232,7 @@
 				type="button"
 				disabled={!state.editingEnabled}
 				aria-label="Reset Highest Unlocked Spell Tier"
-				onclick={() => updateHighestUnlockedSpellTier(getHighestSpellTier(actor))}
+				onclick={() => resetHighestUnlockedSpellTier()}
```
trevlar, line 235: `resetHighestUnlockedSpellTier` is new and untested. One assertion that it writes `null` (not the current computed number) is enough.

public/lang/en.json
trevlar: The key is `notACasterWarning` but `showTierWarning` fires whenever the scroll's tier exceeds the unlocked tier, so a level-4 Mage sees it on a tier-3 scroll. The string text is already tier-shaped; rename the key to match, e.g. `tierLockedWarning`.

src/documents/actor/character.ts
```diff
@@ -519,14 +525,10 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
...
-		const classes = Object.values(this.classes ?? {});
-		if (classes.length === 0) return 0;
-		// Check if there are any spellcasting classes - if not, return null
-		const isSpellCaster = this.system.resources.mana.max > 0;
-
-		if (!isSpellCaster) return null;
-		// Only update if value is null e.g not migrated
+	// Derived from the spell grants on the character's class, subclass, and
+	// feature items. A character with no eligible tiered grant — whatever
+	// resource they hold — has no unlocked tier and is not a caster.
```
trevlar: Three em dashes in the new comments here and on lines 527- 528. Commas or periods.

trevlar (dismissed):
**🔧 Push back — the derivation is solid and well covered, but the override/derive precedence that this PR exists to fix has no test**

**What's good**

- `getHighestSpellTier.packs.test.ts` asserting against the real authored pack ladders is the right way to catch this class of bug, and the Berserker case pins the non-caster path.
- Delegating the threshold check to `appliesTo()` rather than reading `level.min` yourself, with the `{min:2,max:4}` test proving it, is the part that would have been easiest to get wrong.
- I checked all six mana classes (Mage, Oathsworn, Shadowmancer, Shepherd, Songweaver, Stormshifter) — each has a level-predicated `grantSpells` ladder, so none loses its tier to this change. The two unpredicated grants you skip (At Any Cost, Tempest Mage) are "learn a tier 1-3 spell" picks at level 7, where the class ladder already gives tier 3, so dropping them is correct.

---

1. Test the override/derive precedence in `prepareDerivedData`.
2. Handle or document `$and`/`$or` in `hasLevelThreshold`.
3. Test that reset writes `null`.
4. Rename `notACasterWarning` to match the tier condition.
5. Drop the em dashes from the new comments.

Fronix:
All five points addressed.

- Actor prepare test added (`prepareHighestUnlockedSpellTier.test.ts`): a stored number wins, a stored null derives, and a second prepare cycle re-derives. It fails against the `??=` variant, so it pins the fix.
- `hasLevelThreshold` now recurses through `$and` and `$or`. The predicate schema has no `$not`, so nothing is silently skipped, and the JSDoc says what the code does. Three nested predicate tests cover it.
- `resetHighestUnlockedSpellTier` has a test asserting the null write.
- The key is renamed to `tierLockedWarning` and carries your wording.
- Em dashes removed from the three comments.

I also added the user-facing docs this change needed: the spell grants section in `character-options.md`, the override and reset plus the scroll warning in `character-sheet.md`, and the corrected caster sentence in `creation.md`.

trevlar (approved):
**✅ Approved — the derivation matches the printed book for all six caster classes**

I pulled the ladders out of every `grantSpells` rule in `packs/` and checked them against the first-printing Heroes book. Mage, Shepherd, Songweaver, Stormshifter, Oathsworn and Shadowmancer all match rung for rung. The table you deleted matched none of them. Reviewed in the context of the stack, 920 → **921** → 922 → 925 → 926; nothing below blocks it.

- **src/documents/actor/character.ts:281** — The derivation runs here, but ability modifier tags (`intelligence:3` and the rest) don't reach the domain until line 296. A grant predicated on an ability modifier would fail its predicate and get dropped. Nothing in `packs/` gates a spell grant that way and nothing up the stack does either, so this breaks nothing today. What's wrong now is the docstring on `getHighestSpellTier`, which promises "a grant gated on more than a level is honoured in full rather than in part". Ability mods are already final here, since `_prepareEarlyDerivedData` recomputes them, so moving this block below the ability tag loop at 296-298 makes the promise true. Otherwise the docstring should say level and class tags only.

- **src/utils/spell/getHighestSpellTier.ts:93** — `rule.appliesTo && !rule.appliesTo()` reads a missing `appliesTo` as "applies". Every rule coming out of `RulesManager` is a real `NimbleBaseRule`, so production is fine, but raw pack JSON is exactly the shape that carries a predicate and no `appliesTo`. A validation script or importer passing it here would unlock every tier at level 1. Flip the default: make `appliesTo` required on `GrantRuleLike` and skip the rule when it isn't a function. Both test files already bind a real one, so nothing in the suite changes.

- **docs/documentation/homebrew/character-options.md:84** — On a `selectSchool` or `selectSpell` grant, `tiers` is the pool the player picks from, not a ladder. Someone writing "pick 1 spell from tiers 1 to 3 at level 5" would unlock tier 3 at level 5. Nothing in `packs/` does this today; every selection grant there lists a single tier. The **Spellcasting** bullet added in 926 covers cost pools but not this, so it's still worth a sentence.

- **src/documents/actor/character.ts:274** — These four lines explain why the override is read off `_source`, which the test "re-derives on a later prepare cycle when the grants have changed" already pins. The two paragraphs above `hasLevelThreshold` in `getHighestSpellTier.ts:29` do the same job. STYLE_GUIDE.md:29 asks for code that explains itself, and a test holds behaviour down better than a comment does.

**What's good**

- Your open Shadowmancer question is settled. The book gives "Tier 3 Spells" at LEVEL 7, printed alongside the subclass feature. The pack is right and #640's table is wrong, so you can close that note out.
- Cutting caster status loose from mana is the right call, and it's what lets Pilfered Power work in 925. `getHighestSpellTier.packs.test.ts` reading the real compendium JSON through the real `Predicate` is why I trust the derivation that 926 then turns into a hard cast gate.

---

1. Move tier derivation below the ability tag loop, or correct the docstring
2. Make `appliesTo` required, skip when absent
3. Document `tiers` on selection-mode grants
4. Trim comments the tests already pin

Issue #232, opened 2026-01-13
Class resource tracking


trevlar:
Class resources are also something we should have editable on the character sheet.

DoubleKing-Prime:
Oops....accidentally clicked close, so I reopened it. I'll move it back to Backlog from Done

PR #922, opened 2026-08-30
feat(spells): let a class declare what its tiered spells cost

Fronix:
## Summary

Spell cost is hardcoded to mana at the spell's tier, and mana is hardcoded to a
field on the class. A class that pays with something else has to be squeezed into
the mana field and mislabelled. This adds a field on the class item that names a
charge pool and an amount, and a resolver that reads it.

Gameplay does not change for any existing caster. Nothing in the packs sets the new
field yet, so every caster still resolves to tier in mana. What is visible is the class
sheet, which gains a Spellcasting section in its config tab for setting the field. The
Shadowmancer content that uses it is the next PR.

image

## Changes

- New `spellcasting` field on the class item. It names a pool, an amount, an optional
  overdraft consequence, and the last level at which that consequence applies.
- The class sheet edits the new field. The cost controls appear once a pool is named,
  and the level bound once a consequence is chosen, the same way mana recovery hides
  behind a mana formula. The pool is chosen with the rules builder's charge pool
  picker, which lists the actor's pools on an owned class and falls back to free text
  in a compendium.
- New cost resolver. Given an actor and a spell it returns nothing for a cantrip, the
  cast tier in mana by default, or the class's declared pool cost.
- `paySpellCost` does validate, confirm and spend as one operation. Splitting them let
  an awaited confirmation drain the pool between the check and the spend, which left
  the cast free and still posted the card.
- A class may declare that its spells always cast at the highest unlocked tier. The
  cast dialog then fixes the tier control instead of offering it. The tier is resolved
  on the activation path, so macros and the skip-dialog route get it too.
- Overdraft. A declared cost may permit exceeding the pool and say what that costs,
  with a confirmation first. Above the declared level the overdraw is still offered
  but nothing is applied, and the confirmation says the cost is settled at the table.
- The spell list and the level-up card now show the resolved cost, so a pool-paying
  class shows its pool rather than mana it does not hold.

## Test Plan

Existing casters must behave exactly as before, since nothing declares the new field:

1. Cast a tiered spell as a Mage. It should cost its tier in mana.
2. Upcast one. It should cost the chosen tier in mana.
3. Cast a cantrip. It should cost nothing.
4. Check a Commander's Combat Dice and a Berserker's Fury Dice still work.

Then check the new editor:

5. Open a class item and go to Config. It should show a Spellcasting section with a
   Cast at Highest Tier checkbox and a Spell Cost Pool control, and nothing else.
6. Name a pool. Cost Per Cast and Casting Beyond the Pool should appear.
7. Choose Half Max HP Damage. Consequence Stops After Level should appear. Clear the
   box and it should store no bound rather than a level.

The new paths are covered by unit tests over the resolver, which build an actor and
an item and assert on outcomes and failure codes.

## Known limit

Most authored spells carry no class restriction. A character with more than one class
casting a spell that names none falls back to mana rather than guessing which class to
charge. Attributing a spell by the school that granted it would fix this and is not
built.

## Checklist

## Related Issues

References #640. Does not close it, since the Shadowmancer content lands in the next PR.

References #232. This is the spending half of class resource tracking. A class says
which pool its spells cost, and casting draws from that pool. Showing and editing the
pool on the character sheet comes later in the stack.

## Summary by CodeRabbit

- **New Features**
  - Added configurable class spellcasting, including highest-tier casting, custom charge-pool costs, deterministic costs, and overdraft consequences.
  - Spell casting now displays resolved mana or pool costs and supports overdraft confirmation with consequence previews.
  - Upcast dialogs and level-up spell cards reflect character-specific costs and pinned cast tiers.
  - Added class sheet controls for spellcasting and overdraft behavior.

- **Documentation**
  - Documented spellcasting configuration, charge-pool costs, casting windows, and overdraft behavior.

- **Bug Fixes**
  - Improved handling of missing, insufficient, or changing spell resources during activation.


src/view/sheets/components/CastSpellActionPanel.svelte.ts
```diff
@@ -91,9 +96,18 @@ export function createSpellPanelState(
...
-		return getSystemData(spell).tier ?? 0;
+	/**
+	 * The cost of casting the spell as the actor would actually pay it: the
+	 * spell's tier in mana by default, or the flat pool cost the actor's class
+	 * declares. Returns null when the cast is free, so the indicator is omitted.
+	 */
+	function getSpellCostLabel(spell: Item): string | null {
+		const actor = getActor() as unknown as SpellCostActorLike;
+		const spellLike = spell as unknown as SpellLike;
+		const cost = resolveSpellCost(actor, spellLike, {
```
trevlar: `getSpellCostLabel` (`CastSpellActionPanel.svelte.ts:107`) rebuilds the charge-pool map and re-evaluates the amount formula per spell per render, which only costs anything once a pool-paying class exists

src/utils/spell/spellCost.ts
```diff
@@ -0,0 +1,203 @@
...
+	};
+}
+
+/**
+ * Renders a resolved cost as the short label the sheet and the cast dialog
+ * both show. Returns null for a free cast so callers can omit the indicator.
+ */
+export function formatSpellCostLabel(cost: ResolvedSpellCost): string | null {
+	if (cost.type === 'none') return null;
+	if (cost.type === 'pool') return `${cost.amount} ${cost.poolLabel}`;
```
trevlar: the pool label at `spellCost.ts:178` is composed in JS while the mana branch goes through `localize()`.

trevlar (approved):
**✅ Approved**

Clean seam for a genuinely fiddly change: cost resolution, validation, and spending are separable but `paySpellCost` keeps them atomic where it matters, and the whole thing ships inert since nothing in the packs declares the field yet.

- `spendSpellCost` re-validating rather than trusting the caller — with a test that refills the pool during the awaited confirmation — is exactly the right shape for that race.
- The multiclass attribution rules in `getClassSpellcasting` are the subtle part and they're the best-tested part, particularly the "holds no mana, so the single declared cost must be what pays" case.

src/managers/ItemActivationManager.ts
```diff
@@ -136,6 +166,16 @@ class ItemActivationManager {
...
 		if (!dialogData) return { activation: null, rolls: null };
 
+		// A pinned cast tier applies on every activation path, so the macro and
+		// skip-dialog routes synthesize the upcast the dialog would have chosen.
+		if (this.#item.type === 'spell' && !dialogData.upcast) {
+			const synthesized = synthesizePinnedUpcast(
+				this.#item as unknown as SpellLike,
+				this.pinnedCastTier,
+			);
+			if (synthesized) dialogData.upcast = synthesized;
```
trevlar: When `synthesizePinnedUpcast` returns null, `this.spellCost` is still the value resolved at line 153 against `pinnedCastTier`, so the caster pays the pinned tier for a cast that resolves at the spell's base tier. Tick **Cast at Highest Tier** on a class with no pool named, give the character tier 5 unlocked, and a tier‑1 spell with `scaling.mode: 'none'` costs 5 mana for a tier‑1 effect. I confirmed it: `resolvePinnedCastTier` gives 5, `synthesizePinnedUpcast` gives null, and `resolveSpellCost(..., { castTier: 5 })` gives `{ type: 'mana', amount: 5 }`, which is what `spendSpellCost` deducts. Re-resolve the cost with no `castTier` when nothing was synthesized and `dialogData.upcast` is still undefined, so the charge matches the tier the cast lands at. Nothing in the suite pins the current number, so a test on that pair would be worth having.

src/view/dialogs/SpellUpcastDialog.svelte
```diff
@@ -38,30 +39,50 @@
...
+
 	const baseMana = $derived(spell.tier);
 	const resources = $derived(actor?.system?.resources);
 	const currentMana = $derived(resources?.mana?.current ?? 0);
 	// A character always carries a derived unlocked tier. An actor with no
 	// resources node (an NPC or monster) has no ladder to climb, so it casts at
 	// the spell's own tier.
 	const maxTier = $derived(resources ? (resources.highestUnlockedSpellTier ?? 0) : spell.tier);
-	const maxMana = $derived(enforceManaCost ? Math.min(currentMana, maxTier) : maxTier);
+	const maxMana = $derived(
```
trevlar: With a pool cost, the slider's ceiling drops to `maxTier` and the label stays flat, so a pool-paying caster with no pinned tier can drag to their highest unlocked tier and pay the same one charge as a base cast. Shadowmancer won't hit this because it pins the tier too, but any class that names a pool and leaves the checkbox off gets free upcasting. Either scale the pool amount with the upcast steps, or keep the slider at the base tier for the pool cost.

src/utils/spell/spendSpellCost.ts
```diff
@@ -0,0 +1,66 @@
...
+
+	// Re-checked here rather than trusted from the caller: an overdraft
+	// confirmation is awaited between the first check and this spend, and
+	// anything that resolves in that window can drain the pool.
+	const validation = validateSpellCost(actor, cost);
+	if (!validation.ok) return validation;
+
+	const pools = buildEffectiveChargePoolMap(asChargePoolActor(actor));
+	const poolEntry = findChargePoolByIdentifier(pools, cost.poolIdentifier);
+	if (!poolEntry) return validation;
```
trevlar: `validateSpellCost` two lines up already failed on a missing pool, so this branch can't fire, and if it ever did, it would report `ok: true` and hand out a free cast. Return a `poolMissing` failure instead of the validation that you know succeeded.

src/view/sheets/components/CastSpellActionPanel.svelte.ts
```diff
@@ -91,9 +97,36 @@ export function createSpellPanelState(
...
+		);
+	});
+
+	/**
+	 * The cost of casting the spell as the actor would actually pay it: the
+	 * spell's tier in mana by default, or the flat pool cost the actor's class
+	 * declares. Returns null when the cast is free, so the indicator is omitted.
+	 */
+	function getSpellCostLabel(spell: Item): string | null {
+		if (spellCostLabels.has(spell)) return spellCostLabels.get(spell) ?? null;
```
trevlar: Every spell the template asks about comes from the same `spells` array the map was built from, so the fallback below never runs, and it quietly duplicates the resolution logic with a non-reactive `getActor()`. AGENTS.md asks for a throw over a no-op fallback here.

trevlar (changes requested):
**🔧 Push back — a pinned cast tier charges mana for an upcast that never happens**

**What's good**
- Folding validate, confirm and spend into `paySpellCost`, then re-checking inside `spendSpellCost`, closes the drain-during-confirmation window properly, and the refill test at paySpellCost.test.ts:120 is exactly the case that matters.
- Deferring the dialog's charge spends alongside the pool nodes is the right call, and `#consumeChargePools` reading `currentValue` fresh at commit time means the spell cost and the dialog spend stack instead of clobbering each other.

---

1. Re-resolve mana cost when no upcast is synthesized
2. Add a test for pinned tier plus non-scaling spell
3. Stop free upcasting for a pool cost with no pin
4. Return a failure on the missing-pool branch
5. Drop the unreachable spell cost label fallback

Fronix:
All five points addressed, plus the two nits.

- A pinned tier on a spell that does not scale no longer charges the pinned tier. `resolveEffectiveCastTier` returns the pin only when `synthesizePinnedUpcast` would apply, and the manager, the cast window, and both spell cards all read it, so what is shown and what is charged cannot drift apart. Test: pinned tier 5, `scaling.mode: 'none'`, tier 1 spell, charged 1.
- A pool cost with no pin offers no upcast. The rule lives in `computeUpcastBounds`, so the manager also refuses an upcast submitted around the dialog. Cost per cast is flat by design, per the class config wording, and the docs say so now.
- `spendSpellCost` returns `{ ok: false, code: 'poolMissing' }` on the missing pool branch instead of handing out a free cast.
- `CastSpellActionPanel` throws for a spell it does not list instead of the unreachable fallback.
- CodeRabbit nits: `Math.max` on mana max and baseMax, and `getRollData` declared on `SpellCostActorLike`.

One extra: a pinned tier on an Upcast Choice spell now opens the dialog rather than silently taking choice 0, the same way a variable charge spend already forces it.

The CONFLICTING flag was stale, the base was already an ancestor. The branch is rebased on the updated stack.

Fronix:
All three inline findings are fixed, and one of the outside-diff ones is not this PR's.

- **Upcast Choice picked silently (`spellCost.ts:191`).** The earlier fix covered the Alt and skip-dialog path through `mustPrompt`, but `#resolveDialogData` returns on `fastForward` before that check, so a macro still took `choiceIndex: 0`. A fast-forwarded activation of such a spell is now refused with a message naming the spell, matching how an unofferable variable spend is refused before the dialog opens. A caller that knows the answer can pass `upcastChoiceIndex` in the activation options. Two tests, both verified to fail with the guard removed.
- **Multiple matching class declarations (`spellCost.ts:81`).** Fixed here now. It was fixed one PR up the stack, which is why this branch still looked wrong; the guard is `declared.length > 1` before the restricted and unrestricted split, so it covers both cases rather than only the restricted one.
- **Docs (`spells.md:92`).** Corrected. The page claimed Alt always casts at base tier; under a class that pins the tier it casts at the pinned tier. It now says that, and notes that a macro skipping the window cannot cast a pinned Upcast Choice spell without naming the enhancement.

On `ItemActivationManager.ts:794`, the clamp after `@spent`: that line comes from #933 and is already on dev (`034bab69`), outside this PR's diff. #946 was the follow-up that made the spend and `@spent` one number. Worth its own issue if the residual race still stands, but changing it here would be a behaviour change this PR does not claim.

Separately, an audit of the surrounding invariants turned up a pre-existing gap that neither this PR nor the stack introduces: mana affordability is only ever checked in the cast window, so the skip-dialog, Alt, and fast-forward paths deduct with `Math.max(0, current - cost)` and let an unaffordable cast through. dev does exactly the same in `spell.ts`, so this stack preserves the behaviour rather than regressing it. I would rather file that than widen these PRs again.

Fronix:
The docs thread was right to stay open: I had fixed two of its three parts and missed the third.

- The closing line said the total mana is deducted. For a class that pays from a charge pool it is the pool, so it now names both.
- The charge pool bullet read "no slider either, unless it also casts at its highest tier", which implies a pinned pool class gets one. It does not: the slider sits behind `{#if pinnedCastTier === null}` in `SpellUpcastDialog.svelte`, so a class that casts at its highest tier has no slider whether it pays with mana or a pool. The bullet says that now, and says such a class spends charges rather than mana.

The forced-tier half of the finding was already corrected in the previous push.

trevlar (approved):
**✅ Approved**

Clean seam, and it reads right as the middle of the stack (#920 → #921 → this → #925 → #926) rather than as a standalone change. Spell cost moves from a hardcoded tier-in-mana deduction to a resolver a class can redirect, and nothing that casts today changes: no pack sets `spellcasting` yet, so every existing caster still resolves to tier in mana. The mana-less caster and its coverage arrive with the Shadowmancer content in #925, which is the right place for them.

- `paySpellCost` as one operation is the right call, and `spendSpellCost` re-validating instead of trusting the caller is pinned by a test that actually drains the pool during the confirmation.
- Multiclass attribution is the part that could quietly charge the wrong pool, and every branch has a test naming the case it covers, including the mana-less multiclass fallback.

PR #925, opened 2026-08-30
feat(shadowmancer): Pilfered Power instead of mana

### LINKED ISSUE #640: [Automation] Shadowmancer - Pilfered Power: limited-use spellcasting resource

Fronix:
## Summary

A Shadowmancer in a live game cannot cast tiered spells. Their sheet reads
"0/1 mana" for a class the rulebook gives no mana, and the cast dialog refuses
the spell because that number is zero. This gives them Pilfered Power instead:
a pool of DEX uses that refreshes on a Safe Rest, one use per cast at any tier.

This is the PR that makes the previous three visible.

## Changes

- Pilfered Power is a charge pool on the level 2 feature. Its maximum is the
  Dexterity modifier floored at zero, and it refreshes on a Safe Rest, not a
  Field Rest.
- The class mana formula is cleared and replaced with a declared pool spell cost.
- Spells always cast at the highest unlocked tier, so the cast dialog fixes the
  tier instead of offering it.
- Casting with an empty pool is allowed. It asks first, then deals half the
  character's maximum hit points in damage.
- From level 12 the damage stops. Greedy Pact replaces it with a saving throw the
  system cannot express, so the overdraw is still offered and the confirmation
  says to resolve it at the table.
- Heart of Burning Fire returns a use on each initiative roll.
- Migration 048 does the same to a character's embedded copy of the class and
  features, and clears the mana the class never had.

## Test Plan

1. Make a Shadowmancer and take them to level 2. Pilfered Power should read DEX
   out of DEX, and there should be no mana bar.
2. Take them to level 10 and cast Vampiric Greed. The dialog should say
   "Cost: 1 Pilfered Power" and "This spell is cast at tier 4", with no tier slider.
3. Cast it. The pool should drop by one, and the card should show the spell resolved
   at tier 4.
4. Empty the pool and cast again. Confirm the overdraw. You should take half your
   maximum hit points in damage and the pool should stay at zero.
5. Cancel that same prompt instead. Nothing should happen and no card should post.
6. Take a Field Rest. The pool should not refill. Take a Safe Rest. It should.
7. At level 11 with Pact of the Red Dragon, roll initiative. The pool should gain a use.
8. At level 12, overdraw again. No damage, and the prompt should mention the table.
9. On a character who already has an embedded Shadowmancer class, load the world and
   check the migration ran: no mana bar, Pilfered Power present, mana current at zero.

## Note on the migration

There is no test for it. A migration runs once per world, so a test asserting it still
produces what it produced carries no signal after that. Step 9 above is the check.

## Checklist

## Related Issues

Closes #640.

References #232 without closing it. Pilfered Power is the first class resource that is
tracked and spent end to end. Other classes still have no resource of their own, and
spending one outside a spell cast is not possible yet.

Shadowmancer no longer uses the inactive mana pool. The level 2 Pilfered Power feature now gives a DEX-based charge pool.

- The pool refreshes during a Safe Rest.
- Each Shadowmancer spell cast uses one charge.
- Spells cast at the highest unlocked tier.
- Empty-pool casting remains possible after confirmation.
- Before level 12, an overdraw deals half maximum HP damage.
- From level 12, the overdraw uses the Greedy Pact saving throw.
- Heart of Burning Fire restores one use at combat start.
- The system does not automate Hollow One restrictions or specified free casts.
- Homebrew authors can use the `pilfered-power` cost pool and existing pool rules.

## Engine changes

- Cleared the Shadowmancer mana formula.
- Added highest-tier spellcasting with the `pilfered-power` cost pool.
- Updated spell-cost and overdraw handling.
- Updated resource-bound calculations for the class that declares the cost.
- Added support for homebrew classes with blank identifiers.
- Migrated embedded class and feature copies.
- Cleared stale positive Shadowmancer mana values.
- Added migration **052**, and set `LATEST_SCHEMA_VERSION` to **52**.
- Added migration tests and migrated-item source comparisons.
- Added and re-exported `Migration052ShadowmancerPilferedPower`.
- Reused the shared migration `toSnapshotId` helper.
- Used the existing `resourceCounter` and `modifyPool` rule types.
- No new rule types were added.
- No settings were added or changed.
- No `en.json` keys were added or removed.
- Updated character-sheet documentation for pool costs, overdraw, manual exceptions, and Heart of Burning Fire.


src/migration/migrations/Migration048ShadowmancerPilferedPower.ts
```diff
@@ -0,0 +1,211 @@
+import { MigrationBase } from '../MigrationBase.js';
+
+/** Namespace the source ids below are written under. */
+const SNAPSHOT_PREFIX = 'Compendium.nimble.';
+
+/** Every namespace a stored id could carry: the stable id, or the dev rebrand. */
+const STORED_PREFIXES = [SNAPSHOT_PREFIX, 'Compendium.nimble-dev.'];
```
trevlar: `SNAPSHOT_PREFIX`, `STORED_PREFIXES`, and `toSnapshotId` duplicate what already exists in `src/migration/compendiumSourceId.ts`. Migrations 035, 041, 045, and 047 already use the shared helper. Remove lines 3–19 and import it instead:

  `import { toSnapshotId } from '../compendiumSourceId.js';`

  `AGENTS.md` calls out DRY as a required constraint, and the shared helper already has the documentation this version repeats.

docs/documentation/characters/character-sheet.md
```diff
@@ -62,8 +62,22 @@ Your class features, subclass features, ancestry, background, and boons, grouped
...
 
+#### Casting without enough left to pay
+
+A class that pays from a pool may also declare what happens when you cast anyway with the pool empty. The Shadowmancer is the example that ships with the system: Pilfered Power is a pool of uses equal to your Dexterity, spells always cast at your highest unlocked tier for one use, and casting on an empty pool asks you to confirm and then costs you half your maximum hit points.
+
+::: warning Greedy Pact is not automated
+From level 12 the Shadowmancer's Greedy Pact replaces that fixed penalty with a saving throw that has three different outcomes, one of which changes the tier the spell is cast at. The system cannot express that yet, so from level 12 it stops applying the penalty instead of applying the wrong one. The confirmation still offers the overdraw and says the cost is settled at the table. Roll Greedy Pact by hand and apply the result yourself.
+:::
+
+::: warning Heart of Burning Fire does not expire
```
trevlar: the section listing what isn't automated only calls out two cases. There are a few more worth mentioning:

  * Reaver prevents tiered spells from being cast with Pilfered Power, but the app still provides the pool and allows those casts to spend from it.
  * We'll ALL Burn!, Master of Nightfrost, and the crit invocation can all cast without Pilfering Power, but those exceptions aren't automated either.

  Adding one more note here would make the limitations clear.

trevlar (changes requested):
**🔧 Push back: one duplicated helper to clean up, then this looks good to go**

**What's good**

* The numbers all line up with the printed rules: DEX uses, the Safe Rest reset, half max HP, casting at the highest unlocked tier, and Greedy Pact arriving at level 12 with `overdraftMaxLevel: 11`.
* The migration matches by compendium source ID with a class-and-name fallback, only appends rules, and uses the rule ID for idempotence, so it's safe to run more than once.

---

1. Import `toSnapshotId` instead of duplicating it.
2. Document the remaining Pilfered Power cases that aren't automated.

Fronix:
Both points addressed.

- The local `toSnapshotId` is gone; the migration imports the shared helper from `compendiumSourceId.ts` the way 047 does.
- The docs now list the exceptions that are not automated: Hollow One cutting a Reaver off from Pilfered Power, and the free casts from We'll ALL Burn!, Shadowfrost, and Hungering Shadows. One correction to your comment: the Pact of the Abyssal Depths free cast is Shadowfrost at level 7 (Heroes 2.0.2 line 1327). Master of Nightfrost only grants temp HP on a shadow crit, so it is not one of the exceptions.

Also in this push, since it blocks the merge to dev: the migration is renumbered to 052, because dev now ships `Migration048LayOnHandsPool` and #944 holds 049 to 051. `LATEST_SCHEMA_VERSION` is 52. The runner tolerates the gap; `MigrationList` selects by max and by filter and never indexes by version. The mana clear now only touches a positive current, so an undefined value is no longer written as 0 and logged as cleared, and the migration has its first unit test.

trevlar (dismissed):
**🔧 Push back — Heart of Burning Fire uses the initiative trigger this repo already rejected for the same clause**

**Must fix before merge**

- **packs/classFeatures/core/shadowmancer/shadowmancer-subclasses/pact-of-the-red-dragon/heart-of-burning-fire.json:24** — `onInitiativeRolled` fires once per initiative roll and nothing dedupes it per combat. Paired with `"mode": "add"`, a second roll for the same combatant hands back a second use, clamped only at the pool max. Re-rolling is reachable: `combat.rollInitiative` has no already-rolled guard, and the `ActionTracker` roll button doesn't carry the `combatant.initiative !== null` check that `PlayerCharacterCoreTab` does. Migration041 hit this exact clause on Master Commander ("When you roll Initiative, regain 1 spent use of Coordinated Strike") and chose `encounterStart` instead, with the reasoning written out at `Migration041CommanderInitiativeRegain.ts:20`. Both commander packs still ship `encounterStart`. Use it here too. Migration020 gets away with `onInitiativeRolled` because it's `"mode": "refresh"`, which is idempotent.

- **src/migration/migrations/Migration049ShadowmancerPilferedPower.ts:70** — This trigger has to change with the pack file. Leave it as `onInitiativeRolled` and every existing Heart of Burning Fire gets a rule the pack no longer ships, so migrated characters and freshly dragged ones behave differently. Idempotency won't catch it later either: the id `heart-of-burning-fire-initiative-regain` is already present, so a follow-up migration would skip the item.

- **packs/classes/core/shadowmancer.json:125** — `overdraftMaxLevel` is compared against `actor.levels.character` in `spellCost.ts:124`, but Greedy Pact is a Shadowmancer level 12 feature. A Shadowmancer 5 / Berserker 7 is character level 12 with no Greedy Pact, and stops taking the half-max-HP damage they should still take. `levels.classes.shadowmancer` is right there next to `levels.character`, so gating on class level is a small change to the base branch.

**What's good**

- Every "not automated" warning in the docs page checks out against the rulebook: DEX uses, half max HP per overdraw, Greedy Pact at 12, the Heart of Burning Fire expiry, Hollow One, and all three free-cast features.
- The migration matches on source id with a class+name fallback and appends by rule id, same shape as 037 through 041.

---

1. Switch Heart of Burning Fire to `encounterStart`
2. Match that trigger in Migration049
3. Gate the overdraft bound on Shadowmancer level

Fronix:
Thanks, all three are addressed.
1. Heart of Burning Fire now uses `encounterStart`, the same as Master Commander (40105f19).
2. Migration049 uses the same trigger. A new test runs the migration on each Shadowmancer pack item with its rules removed and compares the result with the pack, so the two cannot drift.
3. Fixed in #922 (b8cf0147, 1c7ddd7e). The bound now counts levels in the class that declares the cost, and the lookup uses the item's identifier getter, so a homebrew class with a blank identifier works too. The docs say levels in other classes do not count.

Fronix:
Both points are handled.

1. Version 52: already done in a8e810fc. The review ran on c81afb8a, the head before this branch was rebased onto the current dev. dev merged #951, which took 049 through 051, so the Shadowmancer migration is now `Migration052ShadowmancerPilferedPower` with `LATEST_SCHEMA_VERSION = 52`.
2. Class matching: fixed in 10ffa0b4. `updateActor` and `#updateClass` now share `#isShadowmancerClass`, which accepts the compendium source id, the identifier, or the name when the identifier is blank. The name is read only in that blank case, the same fallback the item's `identifier` getter makes, so a homebrew class named Shadowmancer with its own identifier is still skipped. Three tests cover it: mana cleared on a copied class, the spell cost declaration written on a copied class, and a class that only shares the name left alone.

Issue #916, opened 2026-08-20
A way to add custom actor stats or attributes.


Asherlock:
### Problem or Motivation

It would enable the creation of new classes or subclasses. I'm currently working on both and being able to establish new core mechanics by introducing custom resources would be amazing. 

### Proposed Solution

A supported way to add custom actor resources (value/max pairs) without touching core schema — ideally something akin dnd5e's spare resource slots.

### Alternatives Considered

I've tried module combinations such as bar brawl and custom fields but when I reference calculations like #system.attributes.hp.max# x2, the value of a value/max field becomes uneditable if accessible at all. 

### Additional Context

Please and thank you!

Fronix:
This seems like a very niche feature, what do you mean by "establish new core mechanics"? Are we talking stats, skills, custom interactions?
I'm afraid of something like this getting out of hand, we are homebrew friendly but not "frankenstein your own nimble"-friendly. We already support custom classes and subclasses, the rule system has support to have resources of different kinds (which could be expanded to a degree).

Could you explain a bit more what this would look like?

Asherlock:
Sure, say for instance I want to introduce the Artificer class to the vtt. I'd like to be able to implement it's mana dice pool. 
Or do one better and introduce a brand new custom class like a Alchemist and want to use a resource other than mana, or even simply rename the resource bar "mana" to something more appropriate for the new class. 

I don't understand how you can support custom classes and subclasses without the ability to create new class resources that would enable to creation of those new classes or subclasses. 

Another example might be gunslinger who uses Grit and maybe has a dice pool other than fury dice, maybe they're d8s, or perhaps they have a certain number of special rounds they can fire represented on the character sheet or token hud. 
I don't understand how you can support custom classes and subclasses without being able to create or edit existing resources.

Fronix:
Thanks, that helps a lot. What you're describing is pools and consumers rather than custom schema fields, and most of it already exists. You can build new class resources, it just lives in the rules system rather than the class schema.

One note on your first example: Artificer is official Nimble content rather than homebrew, so that one is a content question on our side and not something you should have to build yourself. We are just waiting on the go-ahead to add it since it's currently Patreon only content.

[...]

PR #926, opened 2026-08-30
feat(charges): let a charge pool show in the sheet header

Fronix:
## Summary

Clearing the Shadowmancer's mana formula empties the slot in the sheet header where
mana used to be, and puts nothing back. This adds a toggle on the charge pool rule
that puts a pool there instead.

Review of the stack then found several defects that are fixed here rather than left
for later, and the cast path gained live tests that drive the real cast window.

## Changes

**A pool may be shown where mana is shown**

- New `showAsResource` toggle on the charge pool rule. Off by default.
- The sheet header renders mana and every pool that sets the toggle in one column.
- `hidden` wins, so the two toggles cannot contradict each other on screen.
- Promotion is additive: a promoted pool keeps its badge on the item that grants it.
- The maximum is read-only in the header; the current value is editable and writes
  through the same path the charges dialog uses.
- `ManaBar` becomes `ResourceBar`, themed so it reads in light and dark mode.
- Pilfered Power sets the toggle, so a Shadowmancer sees it in the header.

**Defects found in review**

- A pinned cast is held to the caster's unlocked tier. A class that casts at its
  highest tier never sees the tier control, so nothing compared the spell against the
  caster's ladder: a Shadowmancer could cast a spell above their tier where a mana
  caster was refused. Enforced on every activation path, so a macro cannot bypass it.
- A charge pool created while its maximum still read zero stayed at zero when the
  maximum later rose. Taking Shadowmancer to level 2 before assigning Dexterity left
  Pilfered Power stuck at 0. It now takes its initial value once the maximum appears.
- The chat card reports the tier a pinned cast resolved at. A pinned class reaches a
  higher tier without an upcast when the spell does not scale, so the card had nothing
  to read it from and disagreed with the cast window.
- The cast window's warnings are localized. An upcast refusal now travels as a code
  and is mapped to its message by hand, which surfaced that one refusal had no message
  at all and would have shown the player a raw translation key.
- `Math.clamp` arguments are passed in the documented order.

**Documentation**

- The Greedy Pact note said a Shadowmancer of level 12 or higher still takes the level
  2 penalty automatically. The opposite is true: the penalty stops and the table
  settles it. Corrected.
- Heart of Burning Fire's regained use should expire at the end of combat and does not.
  Recorded for players, and tracked as #932.

## Test Plan

1. Open a Shadowmancer. The header shows Pilfered Power where mana would be, with the
   current value editable and the maximum greyed out.
2. Type a new number into it. The pool changes, and so does the badge on the feature.
3. Check the Features tab. The badge is still there.
4. Switch to dark mode. The bar still reads correctly.
5. Open a Mage. Their mana bar looks and behaves exactly as before.
6. In the rules builder, add a charge pool to a feature and turn "Show as a resource"
   on. It appears in the header. Turn "Hide from sheet" on as well. It disappears.
7. Take a fresh Shadowmancer to level 2 before assigning Dexterity. Pilfered Power
   fills once Dexterity is set rather than staying at zero.

Automated coverage runs against a live Foundry V14 world (`pnpm test:integration`,
local only): a Shadowmancer and a Mage cast through the real cast window, a Mage
upcasts by moving the window's own tier slider, tier derivation is checked against
each class's authored grants, and every class in the packs is built and asked whether
the system still considers it incomplete.

## Checklist

## Related Issues

References #916 without closing it. This makes existing pool values readable and
editable in one more place, which is part of what that issue is blocked on, but it
does not add the actor resource record that issue needs.

References #232 without closing it. Class resources now sit where mana sits, with
editable current and max values, which is what the issue body asks for. A comment on
that issue also asks for a way to point any ability's cost at a feature's charges. That
part is still open.

References #932, filed from this work: the regained use that should expire at the end
of combat needs a temporary resource concept the system does not have yet.

## Summary by CodeRabbit

- **New Features**
  - Character sheets now display selected charge pools as editable resource bars with current and maximum values.
  - Added support for variable resource costs, overdraft confirmations, and related spell-cost notifications.
  - Spell casting now prevents spells above the character’s highest unlocked tier, including pinned casts.
  - Shadowmancer’s Pilfered Power pool appears in the sheet resources area.

- **Bug Fixes**
  - Corrected hit point, hit dice, and resource percentage calculations.
  - Improved preservation and synchronization of charge-pool values.

- **Documentation**
  - Expanded guidance for resource bars, spell costs, upcasting, and feature limitations.


trevlar:
@coderabbit

Try another review please. I adjusted the settings and your previous attempt was blocked for the branch.

src/managers/ItemActivationManager.ts
```diff
@@ -150,6 +156,19 @@ class ItemActivationManager {
 
 		if (this.#item.type === 'spell' && this.actor) {
 			this.pinnedCastTier = resolvePinnedCastTier(this.actor, this.#item);
+
+			// A pinned class never sees the tier control, so the bound is enforced
+			// here instead. Checked on every activation path, not just the dialog,
+			// so a macro cannot cast above the caster's ladder either.
+			if (this.pinnedCastTier !== null && exceedsUnlockedSpellTier(this.actor, this.#item)) {
```
trevlar: The guard reads `this.pinnedCastTier !== null && exceedsUnlockedSpellTier(...)`, but the comment above it says the bound is "checked on every activation path, not just the dialog, so a macro cannot bypass it." That is only true for pinned casters. A mana caster with a spell whose `scaling.mode` is `none` never reaches `applyUpcastDeltas` — `canUpcast` is false, `synthesizePinnedUpcast` returns null for a null pinned tier — so nothing compares that spell's tier against the ladder on any path. Drop the `pinnedCastTier !== null` clause: `exceedsUnlockedSpellTier` already returns `false` for cantrips and for any spell at or below the unlocked tier, so a legal cast is unaffected and the guard becomes what the comment says it is. If you'd rather keep the narrower scope, correct the comment instead.

src/managers/ItemActivationManager.ts
```diff
@@ -217,8 +236,27 @@ class ItemActivationManager {
...
 			} catch (error) {
+				// Mapped by hand so a code without a message is a type error.
+				if (error instanceof UpcastError) {
+					const { warnings } = CONFIG.NIMBLE.spellUpcastDialog;
+					const messageKeys: Record<UpcastErrorCode, string> = {
+						cantripCannotUpcast: warnings.cantripCannotUpcast,
+						spellCannotUpcast: warnings.spellCannotUpcast,
+						insufficientMana: warnings.insufficientMana,
+						belowBaseTier: warnings.minMana,
+						aboveUnlockedTier: warnings.aboveUnlockedTier,
```
trevlar: `aboveUnlockedTier` is raised by `validateAndComputeUpcast` when the *mana spent* exceeds the ladder, but you map it to `warnings.aboveUnlockedTier` ("This spell is above your highest unlocked spell tier"). The sentence that fits that refusal is `warnings.aboveMaxTier` ("Cannot spend more mana than your highest unlocked spell tier"). It looks unreachable today, but the hand-mapping exists precisely so that code lands on the right sentence.

types/components/ResourceBar.d.ts
```diff
@@ -0,0 +1,8 @@
+export interface ResourceBarProps {
+	current: number;
+	max: number;
+	disableControls?: boolean;
```
trevlar: `disableControls` has no caller anywhere. AGENTS.md's YAGNI rule is explicit that every abstraction needs a current caller and that unsupported paths shouldn't get a no-op fallback. So either use it or lose it.


Related: `ManaBar` used a discriminated union, so enabling controls forced the handlers to exist; `ResourceBarProps` makes `updateCurrent` plainly optional, and the input silently no-ops when it's omitted. Worth keeping the union shape now that two callers pass different subsets.

src/managers/ItemActivationManager.ts
```diff
@@ -150,6 +156,19 @@ class ItemActivationManager {
 
 		if (this.#item.type === 'spell' && this.actor) {
 			this.pinnedCastTier = resolvePinnedCastTier(this.actor, this.#item);
+
+			// A pinned class never sees the tier control, so the bound is enforced
+			// here instead. Checked on every activation path, not just the dialog,
+			// so a macro cannot cast above the caster's ladder either.
+			if (this.pinnedCastTier !== null && exceedsUnlockedSpellTier(this.actor, this.#item)) {
+				ui.notifications?.warn(
+					localize('NIMBLE.spells.spellUpcastDialog.warnings.aboveUnlockedTier', {
```
trevlar: Raw key string here, while the catch block twelve lines down reads the same family through `CONFIG.NIMBLE.spellUpcastDialog.warnings`. You already added the constant; use it.

src/utils/chargePool/helpers.ts
```diff
@@ -666,6 +690,7 @@ function areChargePoolStatesEqual(left: ChargePoolState, right: ChargePoolState)
 		left.dieSize === right.dieSize &&
 		left.icon === right.icon &&
 		left.hidden === right.hidden &&
+		left.showAsResource === right.showAsResource &&
```
trevlar, line 800: `areChargePoolStatesEqual` gained `showAsResource` but not `seeded`, so a pool whose only change is the marker is treated as unchanged and never persisted. The practical effect today is that pools stored before the marker existed never get it backfilled; `wasSeeded`'s `?? max > 0` fallback keeps behavior right, so nothing breaks, but the field is now part of `ChargePoolState` and the next code path that reads it will be reading a value that was never written.

trevlar (changes requested):
**🔧 Push back — the new tier guard only covers half the casters its own comment claims**

**What's good**

- The seeding fix is stored rather than derived, and the tests actually walk the interesting sequence — seeded, spent to zero, maximum lapses, maximum returns — instead of just asserting the happy path.
- `hidden` winning over `showAsResource` is enforced in one place (`getResourcePools`) and pinned by a test, so the two toggles can't drift apart.

---

1. Guard the tier bound for non-pinned casters too, or fix the comment.
2. Map `aboveUnlockedTier` to `warnings.aboveMaxTier`.
3. Drop `disableControls`; restore the handler-requiring union.
4. Use the CONFIG constant, not a raw key string.
5. Compare `seeded` in `areChargePoolStatesEqual`.

Fronix:
All five points addressed, plus the CodeRabbit finding.

- The ladder check runs on every cast, pinned or not. An actor with no ladder at all, so every NPC, stays unbounded, which matches what `computeUpcastBounds` already does. Two tests: a mana caster with a non-scaling spell above the ladder is refused, an NPC is not.
- `aboveUnlockedTier` maps to `warnings.aboveMaxTier`, and the raw key literals are gone in favour of the CONFIG paths, for both that code and `upcastFailed`.
- `ResourceBar`: `disableControls` removed, and the props are a union again, so an editable bar cannot render without its update handler.
- `areChargePoolStatesEqual` compares `seeded`, with a test that a seeded-only change is persisted.
- CodeRabbit: a spell open to two classes that both declare a cost now falls back to mana instead of picking `declared[0]` by item order.
- The level-up conformance test closes its sheet in `afterAll` inside try/finally.

Two more from a self-review before pushing: the chat card no longer prints a pinned tier the cast window deliberately does not show for a non-scaling spell, and the live test in `spellTierAndCost.test.ts` now pins the tier on a spell that actually scales (Vampiric Greed), since Shadow Trap has no scaling block and the assertion could never have passed on a clean run.

types/components/ResourceBar.d.ts
```diff
@@ -0,0 +1,17 @@
...
+	max: number;
+	updateCurrent: (value: number) => void;
+}
+
+interface ResourceBarWithFixedMax extends ResourceBarBaseProps {
+	disableMaxEdit: true;
+	updateMax?: (value: number) => void;
+}
+
+interface ResourceBarWithEditableMax extends ResourceBarBaseProps {
```
trevlar: Nothing constructs this arm of the union. Both `ResourceBar` uses in `PlayerCharacterSheet.svelte` pass `disableMaxEdit={true}`, so `ResourceBarWithEditableMax`, the `disableMaxEdit = false` default in the component, and the `updateMax` prop are all unreachable. The mana call site wires `updateMax={updateMaxMana}` into an input that is permanently `disabled`, which leaves `updateMaxMana` in `PlayerCharacterSheet.state.svelte.ts:348` dead as well. AGENTS.md is explicit that every abstraction needs a current caller, and that an unsupported path throws rather than sitting as a no-op. Collapse this to the one shape that has callers: drop `updateMax` and `disableMaxEdit`, render the maximum read-only, and delete `updateMaxMana` and the prop that feeds it.

src/view/sheets/PlayerCharacterSheet.svelte
```diff
@@ -279,31 +281,49 @@
...
+					updateCurrent={updateCurrentMana}
+					updateMax={updateMaxMana}
+					disableMaxEdit={true}
+				/>
+			{/if}
+
+			{#each resourcePools as pool (pool.id)}
+				<h3 class="nimble-heading nimble-heading--resource">
+					{pool.label}
+					<i class={pool.icon || 'fa-solid fa-bolt'}></i>
```
trevlar: `ChargeUiConfig.defaultPoolIcon` already holds `'fa-solid fa-bolt'`, and the three other charge-pool readouts read it from there: `ChargeIndicator.svelte:138`, `ConfigureChargesDialog.svelte:146`, `ChargeAdjustmentCard.svelte:36`. A fourth copy of the literal means the next change to the default pool icon misses the sheet header. Import `ChargeUiConfig` and write `pool.icon || ChargeUiConfig.defaultPoolIcon`.

src/view/sheets/components/ResourceBar.svelte
```diff
@@ -0,0 +1,126 @@
...
+
+<div
+	class="nimble-resource-bar"
+	style="--nimble-resource-bar-percentage: {max > 0
+		? Math.clamp(Math.round((current / max) * 100), 0, 100)
+		: 0}%"
+>
+	<div class="nimble-resource-bar__bar">
+		<div class="nimble-resource-bar__values">
+			<input
```
trevlar, line 15: Neither input has an accessible name. One unnamed mana bar was tolerable; now mana plus every promoted pool stack is in the same column, so a screen reader reads a run of anonymous spin buttons with no way to tell Pilfered Power from mana. The `<h3>` that names each bar lives outside the component, so nothing associates it. Take a `label` prop and set `aria-label` on both inputs from it.

trevlar (dismissed):
**🔧 Push back — the new resource bar ships a shape nothing calls, and re-types a constant that already exists**

The feature itself is right, and the unit coverage behind the seeding fix and the upcast refusal codes is genuinely good. Two documented constraints slipped through in the sheet-header code.

**Must fix before merge**

- **types/components/ResourceBar.d.ts:12** — Nothing constructs this arm of the union. Both `ResourceBar` uses in `PlayerCharacterSheet.svelte` pass `disableMaxEdit={true}`, so `ResourceBarWithEditableMax`, the `disableMaxEdit = false` default in the component, and the `updateMax` prop are all unreachable. The mana call site wires `updateMax={updateMaxMana}` into an input that is permanently `disabled`, which leaves `updateMaxMana` in `PlayerCharacterSheet.state.svelte.ts:348` dead as well. AGENTS.md is explicit that every abstraction needs a current caller, and that an unsupported path throws rather than sitting as a no-op. Collapse this to the one shape that has callers: drop `updateMax` and `disableMaxEdit`, render the maximum read-only, and delete `updateMaxMana` and the prop that feeds it.

- **src/view/sheets/PlayerCharacterSheet.svelte:316** — `ChargeUiConfig.defaultPoolIcon` already holds `'fa-solid fa-bolt'`, and the three other charge-pool readouts read it from there: `ChargeIndicator.svelte:138`, `ConfigureChargesDialog.svelte:146`, `ChargeAdjustmentCard.svelte:36`. A fourth copy of the literal means the next change to the default pool icon misses the sheet header. Import `ChargeUiConfig` and write `pool.icon || ChargeUiConfig.defaultPoolIcon`.

- **src/view/sheets/components/ResourceBar.svelte:21** — Neither input has an accessible name. One unnamed mana bar was tolerable; now mana plus every promoted pool stack in the same column, so a screen reader reads a run of anonymous spin buttons with no way to tell Pilfered Power from mana. The `<h3>` that names each bar lives outside the component, so nothing associates it. Take a `label` prop and set `aria-label` on both inputs from it.

- **src/view/sheets/components/HitPointBar.svelte:46** — The PR body lists the `Math.clamp` argument order as fixed, but only the copy that moved into `ResourceBar` got corrected. This line plus `NPCSheet.svelte:21`, `HitDiceBar.svelte:73` and `PlayerCharacterSheet.state.svelte.ts:28` still pass `(0, value, 100)`. They compute the right number by accident, which is why nothing ever caught them.

**What's good**

- Storing the `seeded` marker instead of inferring it from `max > 0` is the right call, and it's what keeps a pool the player already spent from refilling when its maximum lapses to zero and returns. `helpers.test.ts` pins all four cases, the legacy read included.
- Turning upcast refusals into codes and mapping them by hand in `ItemActivationManager` is worth the churn. `Record<UpcastErrorCode, string>` is what surfaced the refusal with no message, and it keeps it surfaced.

---

1. Drop the unused editable-max path from ResourceBar
2. Use `ChargeUiConfig.defaultPoolIcon` instead of the literal
3. Give the resource bar inputs accessible names
4. Fix the four remaining `Math.clamp` argument orders

trevlar (approved):
**✅ Approved**

The `showAsResource` toggle lands cleanly: promotion is additive, `hidden` still wins, and the header writes through `adjustPool`, the same path the charges dialog already uses.

- The `seeded` marker is the right shape for the zero-maximum bug. Storing it instead of inferring it from the max means a pool that lapses to zero and comes back doesn't quietly refill, and `helpers.test.ts` pins all three cases.
- Turning the upcast refusal into a code is what surfaced the message that was missing. `Record<UpcastErrorCode, string>` means the next code can't ship without one.

PR #927, opened 2026-08-30
fix(rules): recalculate maxHpBonus on level-up instead of banking it

### LINKED ISSUE #499: maxHPBonus Rule is not recalculated on Level-Up

Fronix:
## Summary

`maxHpBonus` wrote its contribution into the stored `system.attributes.hp.bonus` when an item was added and unwound it on delete. A `perLevel` rule on anything other than a class item never saw the level change, so the banked total froze at whatever it was worth when the item was acquired. The contribution is derived on every prepare cycle now, so it tracks the actor's level instead of a stale copy.

## Changes

- `MaxHpBonusRule` drops `preCreate` / `preUpdate` / `afterDelete` and exposes `resolvedBonus()`, resolved on demand.
- `character.ts#_prepareHitPoints` sums those into `hp.max`. It has to happen there rather than in a `prePrepareData` hook, because `hp.max` feeds `_populateDerivedTags()`, which runs ahead of the rule sweep. That makes the rule evaluate its predicate earlier than any early-phase rule, so it declares `appliesInPrePrepareData` explicitly; the base class infers that from a `prePrepareData` method it does not have.
- `hp.bonus` keeps its old meaning of the player's manually entered bonus. The Edit Hit Points dialog adds the rule total on top of it and gets its per-item values from `resolvedBonus()` so the breakdown and `hp.max` cannot drift apart.
- `Migration044MaxHpBonusDerived` (schema 43 to 44) removes the previously banked amount, otherwise existing characters double count.
- The import preview counted only the flat `hp.bonus`, which used to carry the banked total, so imported characters with Veteran or Experienced previewed short.
- `docs/system/rules.md` gains a section on rules read outside the hook sweep.

## Test Plan

Unit tests cover the arithmetic; the live test guards that nothing writes to stored data. To check by hand:

1. Put a `{ "type": "maxHpBonus", "value": 2, "perLevel": true }` rule on a **feature** item (not a class item) and add it to a level 1 character. Max HP goes up by 2.
2. Level the character to 2. Max HP now reflects 4, and it keeps tracking on further level ups. On `dev` this stays at 2.
3. Open Edit Hit Points. The "From Rules" total and the max HP on the sheet agree.
4. Type a number into the manual bonus field and save. It is stored on its own and the rule bonus is not folded into it.
5. On a world that predates this branch, load once and confirm the migration runs: max HP should be unchanged from before the upgrade for flat rules.

## Checklist

## Related Issues

Closes #499

## Notes for the reviewer

Three things I did not fix, flagged rather than buried:

1. **The migration cannot be exact for a drifted `perLevel` rule.** Nothing recorded what each rule had banked, so it reconstructs the amount as what the rule resolves to today. That is exact for a flat rule, and every `maxHpBonus` in the shipped compendiums is flat. Where the reconstruction overshoots what is stored, it clamps at zero, which is right unless the character also set a manual bonus.
2. **A predicated `maxHpBonus` sees only base tags.** At `_prepareHitPoints` time only `size:*` and `disposition:*` exist, so a predicate on `class:*` or `level:*` resolves to 0. This is inherent: `hp.max` feeds the derived tags, so it cannot depend on them. The old code sidestepped it by ignoring predicates entirely. No shipped rule has a predicate, but it is worth a decision.
3. **An export taken before this migration double counts on import.** Its `hp.bonus` still holds the banked amount and world migration does not re-run for it.

`pnpm test:integration` needs a live Foundry instance, which I could not run; `tests/integration/itemGrants.test.ts` is reasoned about, not executed.


src/documents/actor/character.ts
```diff
@@ -462,8 +463,19 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
...
 		if (classes.length === 0) return;
 
+		// Summed here rather than through a `prePrepareData` hook because `hp.max`
+		// has to be final before `_populateDerivedTags()` runs, which is ahead of
+		// the rule sweep.
+		const rules = this.rules as unknown as MaxHpBonusRule[];
+		const bonusFromRules = rules.reduce(
+			(acc, rule) => acc + (rule.type === 'maxHpBonus' ? rule.resolvedBonus() : 0),
+			0,
+		);
```
trevlar, line 473: Adding the bonuses to `hp.max` is what closes #499, but that behavior isn’t currently tested. `maxHpBonus.test.ts` only tests `resolvedBonus()` in isolation, while the live test only confirms that the stored data remains unchanged. That leaves the reduction logic untested, including filtering for `rule.type === 'maxHpBonus'`, adding the bonuses to `hp.max`, and returning early for classless actors.

`src/documents/actor/earlyModifierPass.test.ts` already has a character stub for calling `_prepareEarlyDerivedData`, so adding a test there should be straightforward. Please cover that mixed rule types only add the `maxHpBonus` rules and that increasing the stub’s level increases `hp.max`.

src/import/playerCharacter/buildImportPreview.ts
```diff
@@ -85,22 +85,57 @@ function itemLevel(item: ParsedActorItem | undefined): number | null {
...
+	let total = 0;
+	for (const item of items) {
+		const rules = (item?.system?.rules ?? []) as ParsedMaxHpBonusRule[];
+		if (!Array.isArray(rules)) continue;
+
+		for (const rule of rules) {
+			if (rule?.type !== 'maxHpBonus' || rule.disabled) continue;
+
+			const value = Number(rule.value) || 0;
+			total += rule.perLevel ? value * level : value;
```
trevlar, line 109: The preview counts predicated rules at full value, but `resolvedBonus()` always resolves them to 0 for the reason above. As a result, the preview promises max HP that the imported character won’t actually have.

Skip any rule with a non-empty predicate in the preview so the two values stay consistent.

trevlar (changes requested):
**🔧 Push back — the fix is right, but the line that actually fixes #499 has no test, and two edge cases in the new code are worth closing**

The core call is correct: deriving the contribution instead of banking it is the only way `perLevel` can track a level-up, and you traced the prepare-order constraint properly (`_prepareEarlyDerivedData` → `_populateDerivedTags` → sweep). The `appliesInPrePrepareData` override, the `hp.bonus` re-definition as "manual only", and the docs section are all well-argued.

**What's good**

- Replacing three write-back hooks (`preCreate` / `preUpdate` / `afterDelete`) with a single derived read is the right shape. It removes the whole class of reconcile bugs that #499 was one instance of, rather than patching the one that got reported.
- Rewriting the live test to assert on `actor._source` rather than the prepared value is the correct guard for "nothing writes it back", and the docs section on rules read outside the hook sweep will save the next person the same archaeology.

---

1. Test the `hp.max` summation in `_prepareHitPoints`.
2. Warn on any derived-tag predicate for `maxHpBonus`.
3. Skip predicated rules in the import preview so it matches runtime.

Fronix:
Pushed 0c3aa237 for points 1 and 2. On point 3 I want to push back, because I think skipping predicated rules trades one wrong number for a different wrong number.

**1. Test the `hp.max` summation.** You were right, and the miss is worse than it looks: `earlyModifierPass.test.ts` already stubs the exact pass, and stubs this method out with `vi.fn()` on line 53. Six tests now run it for real, covering the `rule.type` filter, the sum into `hp.max`, a `perLevel` rule rescaling when the level rises, and the classless early return.

**2. Warn on a derived-tag predicate.** Done, but not as a list of tag prefixes. That list cannot work: `extractReferencedKeys` returns the bare prefix, so an actor's `class:fighter` tag and the identical one a feature item carries through `system.class` both come back as `class`. Warning on the prefix would fire on a rule that works fine.

Instead `maxHpBonus` gets an `afterPrepareData` hook that re-tests its own predicate once the domain is complete and warns when the answer flipped. A predicate that fails during the HP pass and passes afterwards is, by definition, one that contributed nothing. No prefix list, no false positives.

**3. Skipping predicated rules in the import preview.** Here is my hesitation.

"`resolvedBonus()` always resolves them to 0" is not quite true. `test()` unions the actor's domain with `item.getDomain()` (`item/base.svelte.ts:149`), and the item's own tags are already populated at that point. So `identifier:`, `type:`, `class:` on a feature, `school:`, `tier:`, `objectType:` and the actor's `size:` and `disposition:` all resolve normally. Only predicates on tags the *actor* derives later resolve to 0.

The preview reads raw JSON with no actor and no item documents, so it cannot evaluate any of that. It can count every predicated rule, which over-reports for the late-tag case you found, or skip every predicated rule, which under-reports for a `size:small` or `type:boon` predicate that genuinely applies. Both are wrong, just in opposite directions.

I left it counting, because over-reporting is the failure the user notices and corrects, while under-reporting silently hides HP the character will actually have. If you would rather it skip, say so and I will change it. What I would prefer is a third option: mark the number as an estimate in the preview when any predicated `maxHpBonus` is present, so it stops claiming a precision it does not have. Happy to do that here or as a follow-up.

Two things still outstanding from the PR description, unchanged: the migration cannot exactly reconstruct a drifted `perLevel` rule, and an export taken before this migration double counts on import.

trevlar (changes requested):
T**🔧 Push back — the seed strands mana on level down, which is now visible where it wasn't before**

The core fix is right and the investigation behind it is better than most. One follow-on state needs closing before merge.

**Must fix before merge**

- **src/documents/actor/character.ts:1735** — (`src/view/ui/ctTopTracker/resources.utils.ts:320`). Add the same clamp `configureMana` already does at `character.ts:772` — push `'system.resources.mana.current'` into `actorUpdates` when current exceeds the post-level-down max. Worth a test pinning `2/0` can't happen. Note this doesn't change the level-down/level-up refill you documented and accepted; it just stops the stranded display.

- **src/hooks/manaSeeding.ts:28** — you key the stash by `typedActor.id` specifically to disambiguate actors inside one batched update, but unlinked token actors share the base actor's `id`. `typedActor.uuid` is unique per synthetic actor and is a drop-in swap that makes the key actually do what the comment above it promises.
- **src/hooks/manaSeeding.test.ts:225** — the `typeof previousMax !== 'number'` guard at `manaSeeding.ts:71` is the one branch in the new logic with no test. A post-hook firing without its pre-hook (another module short-circuiting, a hook registered mid-operation) should be a silent no-op — cheap to pin.

**What's good**

- Keying the stash by actor id, then proving it with the batched-update test at `manaSeeding.test.ts:194`, is the kind of thing that normally ships broken. Note `base.svelte.ts:1137` has the same single-slot shape for `previousHp` and presumably the same latent bug — separate PR, but you found it.
- The `game.user?.id !== userId` gate is the right call over a GM-only or owner check: no client attempts a write the server would reject, and exactly one client writes.

---

1. Clamp current mana on level down; add a test
2. Key the stash by `uuid`, not `id`
3. Test the missing-stash no-op branch

Fronix:
@trevlar fixed

trevlar (approved):
**✅ Approved**

Deriving the contribution instead of banking it is the right shape for this, and you traced the ordering constraint correctly — `_prepareEarlyDerivedData` runs after `prepareBaseData` sets `levels` and after `prepareEmbeddedDocuments` populates `this.rules`, so `@level` and the rule list are both live by the time `_prepareHitPoints` reads them.

- The `appliesInPrePrepareData` override and the `afterPrepareData` phase-mismatch warning are a genuinely good pair: the base introspection would have silently dropped the guardrails, and `LATE_PREDICATE_KEYS` doesn't cover `class:`/`level:`. Flipping the integration test to assert on `_source` is exactly the right thing to pin.
- `docs/system/rules.md` earns its place — "rules read outside the hook sweep" is the kind of thing the next person would otherwise rediscover the hard way.

**Non-blocking, two things worth a follow-up:**

- **src/view/dialogs/EditHitPointsDialog.svelte:44** — the dialog calls `resolvedBonus()` after prepare, so the predicate resolves against the *full* domain, while `_prepareHitPoints` resolved it against base tags only. For a rule predicated on `class:*`/`level:*` the "From Rules" total and `totalMaxHp` will exceed the sheet's `hp.max` — the drift the PR body says this closes. Caching the early-pass result on the rule and reading that from the dialog would make the invariant actually hold, and would also stop the dialog's call from clobbering `#rejectedInEarlyPass` (line 77) and suppressing the warning you just added.

Your three reviewer notes all check out. On note 1 specifically: the reconstruction is tighter than you claim — the old `preUpdate` added a flat `value` per level increment on top of `value * levelAtAdd`, so `value * level` is exact for `perLevel` rules on class items too, and every mainstream item-add path goes through `Item.createDocuments` (which does fire `preCreate`), so the banked amount really is there to remove. Note 3 is the one with no clean answer — the export carries no schema stamp, so worth calling out in the release notes rather than trying to detect it.

PR #929, opened 2026-08-31
fix(shepherd): grant 2 Sacred Graces at level 5

trevlar:
## Summary

The Shepherd was only ever offered **one** Sacred Grace per qualifying level, including level 5 where the rulebook says "Choose 2 Sacred Graces". This corrects the compendium data so the Shepherd gets 2 graces at level 5 and 1 each at levels 9 and 13 — 4 total, matching the rulebook.

## Changes

- `packs/classFeatures/core/shepherd/shepherd-progression/sacred-graces.json`: split the single `levelUpOptions` entry (`applyAtLevels: [5, 9, 13]`, `selectionCount: 1`) into two entries with disjoint levels:

  | id | levels | count | label |
  |---|---|---|---|
  | `sacred-grace-initial` | 5 | 2 | "Choose 2 Sacred Graces" |
  | `sacred-grace` | 9, 13 | 1 | "Choose a Sacred Grace" |

- Added `describe('Shepherd — sacred-grace selection counts')` to `src/utils/classProgression/shepherd.test.ts`, asserting the exact per-level pick count and the 4-grace total.

This is a **data-only fix** — no schema or resolver change was needed. `levelUpOptions[].selectionCount` already supports per-level counts; the Shepherd data just never expressed it. The new shape copies the pattern already used by [`mage-progression/spell-shaper.json`](packs/classFeatures/core/mage/mage-progression/spell-shaper.json), which has the identical progression (2 at the first level, then 1 at 9 and 13), and by `hunter-progression/thrill-of-the-hunt.json`.

## Test Plan

Automated (this is what proves the fix):

```bash
pnpm vitest run src/utils/classProgression/shepherd.test.ts
```

The new tests drive the **real** feature resolver against the on-disk compendium via the shared `classProgression` fixture. Before the data change they failed for exactly the reported bug:

```
expected 1 to be 2   ← level 5 selection count
expected 3 to be 4   ← total graces across the progression
```

Levels 9 and 13 were already correct and passed throughout.

Manual verification in Foundry (**not done by me — see checklist**): create or level a Shepherd to 5 and confirm the level-up dialog asks for 2 Sacred Graces, then confirm levels 9 and 13 each ask for 1.

## Checklist

- [x] `pnpm check` passes (format, lint, circular deps, type-check, tests) — 3788 tests / 245 files green
- [x] No hardcoded user-facing strings (used `localize()`) — n/a, no code changed; the option labels are compendium data, matching how `spell-shaper.json` stores them
- [ ] Tested in Foundry with no console errors — **not done.** Setup prompted to repoint the local `systems/nimble` symlink away from another active worktree, so I left it alone rather than disturb it. Needs a reviewer to click through the level-up dialog.

## Related Issues

Reported in conversation, no tracking issue.

### DIFF
--- a/packs/classFeatures/core/shepherd/shepherd-progression/sacred-graces.json
+++ b/packs/classFeatures/core/shepherd/shepherd-progression/sacred-graces.json
@@ -46,10 +46,21 @@
 		"selectionCountByLevel": {},
 		"levelUpOptions": [
 			{
+				"id": "sacred-grace-initial",
+				"label": "Choose 2 Sacred Graces",
+				"applyAtLevels": [
+					5
+				],
+				"rules": [],
+				"selectionGroups": [
+					"sacred-grace"
+				],
+				"selectionCount": 2
+			},
+			{
 				"id": "sacred-grace",
 				"label": "Choose a Sacred Grace",
 				"applyAtLevels": [
-					5,
 					9,
 					13
 				],
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -605,6 +605,14 @@
 			"confirmLevelDown": "Confirm Level Down",
 			"levelUpInProgress": "Cannot revert level while a level up is in progress."
 		},
+		"levelCorrectionDialog": {
+			"title": "Level Correction",
+			"hint": "This character is missing choices from an earlier level. Make them here to bring the sheet up to date.",
+			"gapHeading": "Level {level}",
+			"nothingMissing": "This character is not missing any level choices.",
+			"sheetWarning": "Missing level choices",
+			"sheetWarningTooltip": "Missing choices from level {level}. Click to correct."
+		},
 		"skillPointAssignment": {
 			"decrementSkillPoints": "Decrement Skill Points",
 			"incrementSkillPoints": "Increment Skill Points",
--- a/src/documents/actor/applyLevelCorrection.test.ts
+++ b/src/documents/actor/applyLevelCorrection.test.ts
@@ -1,0 +1,113 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+import { NimbleCharacter } from './character.js';
+
+interface HistoryEntry {
+	level: number;
+	grantedFeatureIds: string[];
+}
+
+function makeCharacter(levelUpHistory: HistoryEntry[]) {
+	return {
+		system: { levelUpHistory },
+		createEmbeddedDocuments: vi.fn(async (_type: string, sources: Array<{ name: string }>) =>
+			sources.map((source, position) => ({ id: `granted-${position}`, name: source.name })),
+		),
+		update: vi.fn(async () => undefined),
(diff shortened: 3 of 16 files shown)


Fronix (dismissed):
The data matches the rulebook, and the shape matches `spell-shaper.json` and `thrill-of-the-hunt.json`. `selectionCountByLevel` was the wrong field, because `getClassFeatures.ts:431` removes each group that an applicable `levelUpOption` claims. You chose the correct field.

Correct these two problems before you merge.

**The player never sees the new label.**

`LevelUpFeatureOptionPicker.svelte:48` draws the option list only when more than one option applies. Only one option applies at level 5, so the dialog never shows "Choose 2 Sacred Graces". The player sees only the count from `selectionCount`. Show the label of the option when only one option applies.

**A Shepherd above level 4 keeps one grace.**

This branch corrects the data, but it does not correct an existing character. The dialog holds the selected option id in local state and never stores it, so nothing marks the character as short one grace. A migration can count the Sacred Grace items on each Shepherd and find the difference. But a migration cannot choose which grace to add, because the player must choose it.

Tell me which correction you want. You can show a prompt at the next level up, or you can put an instruction in the release notes.

trevlar:
## Fixed
- **Player never saw the new label** — the picker now draws the sole applicable option's label as its hint, so leveling to 5 reads "Choose 2 Sacred Graces"; the "(Choose one)" hint is kept only when a level actually offers alternatives (`LevelUpFeatureOptionPicker.svelte:47`).
- **A Shepherd above level 4 keeps one grace** — a new audit recomputes, from the class data, which feature pools a character is still owed picks from, and the sheet shows a warning banner that opens a level correction dialog to make those picks (`findMissingLevelSelections.ts:140`, `PlayerCharacterSheet.svelte:342`, `CharacterLevelCorrectionDialog.svelte`).
- **Corrections outliving a level down** — each granted pick is appended to the `levelUpHistory` entry for the level that owed it, so `revertLastLevelUp` still removes it (`character.ts:1289`).

## Not fixed
- Nothing.

## Approach
Not a migration. A migration can count the graces but cannot make the choice, and there is nothing to store the outstanding pick in — so the shortfall is recomputed from the class data on every sheet open instead, keyed on class/level/owned-feature-sources so the feature index is not rebuilt on unrelated updates.

The audit is general, not Shepherd-specific: it walks levels 1..current, sums each pool's required picks, and compares that against the pool members the character owns. A level whose level-up option offers more than one alternative is skipped — that choice is never stored, so a missing pick there cannot be told apart from an alternative that needed none (this is what keeps the Commander's "Fit for Any Battlefield" silent). A shortfall spanning several levels is reported once against the earliest short level, which also stops the dialog offering the same candidate twice.

False positives were the main risk, so `findMissingLevelSelections.test.ts` builds a correctly-levelled character of every class to 20 with the real resolver against the real compendium and asserts a clean audit, then removes graces to pin the Shepherd cases.

Fronix (approved):
LGTM

PR #933, opened 2026-09-05
feat(oathsworn): automate Lay on Hands as a tracked healing pool

### LINKED ISSUE #632: [Automation] Oathsworn - Lay on Hands: healing power pool
## Context
Oathsworn class feature **Lay on Hands** (Level 1). A flexible HP-restoration pool that scales with level and recharges on a Safe Rest. Torment (Oathbreaker L7) also consumes from this pool to boost damage.

Pack file: `packs/classFeatures/core/oathsworn/oathsworn-progression/lay-on-hands.json`.

## Rules from Rulebook

> "Lay on Hands. Gain a magical pool of healing power. This pool's maximum is always equal to 5xLVL and recharges on a Safe Rest. Action: Touch a target and spend any amount of remaining healing power to restore that many HP."

## Implementation
- Resource counter with max = 5 × LVL, recharges on Safe Rest
- Action: prompt for amount to spend, heal target for that amount
- Exposed as a sheet resource/action button
- Pool must be spendable by other features (e.g., Torment) for non-healing effects

## Rule Types
- `chargePool` (scaling max, Safe Rest recharge)
- `chargeConsumer` in `variable` cost mode: prompts for the amount to spend and exposes it to the item's effect formulas as `@spent`

## Blocked By
Nothing. #576 and #582 are both closed.

Part of #365

Fronix:
## Summary

Automates the Oathsworn's Lay on Hands as a tracked healing pool (#632): a charge pool of `5 * @level` that refills on a Safe Rest, and a new `variable` cost mode on `chargeConsumer` that asks the player how much to spend and exposes it to the item's own effect formulas as `@spent`.

## Changes

- **New `variable` cost mode on `chargeConsumer`.** `cost` becomes the smallest legal spend, `maxCost` the largest (blank meaning whatever the pool has left). The activation dialog asks for the amount, the item's damage and healing formulas read it as `@spent`, and the chat card reports the spend. Fixed consumers are untouched.
- **Lay on Hands ships the pool.** Item-scoped under the `lay-on-hands` identifier, which is what lets a later feature such as Torment spend from it for something other than healing. Migration048 backfills copies already on characters.
- **An item with a variable consumer always opens its roll window**, since the amount is player input with no default.
- **Pool spends now settle only after the `preUseItem` gate.** Charges and dice were both deducted during `getData()`, which runs before the gate that validates them, so a refused use could keep the resources it had already eaten. Spending the pool to empty hit this on Lay on Hands itself.
- **Spends respect `automation.resourceSpending`.** The prompt still runs, because the amount drives the healing formula; only the deduction is skipped.
- **An item with two variable consumers on one pool is refused**, naming the pool, rather than silently honouring one of them.
- Drops the placeholder `0` damage preview from the roll dialog for activations that deal no damage.

Note for the reviewer: `497f3cba chore(lang): babele update` is regenerated es/fr translation data, roughly 1150 lines across 12 files. It is unrelated to this issue and does not require reviewing.

## Test Plan

Automated: `pnpm check` passes (4088 unit tests). `pnpm test:integration` passes (108 tests), including `tests/integration/chargeSpend.test.ts`, which drives the whole flow through real DOM clicks and is verified under shuffled test order.

By hand, on a level 1 Oathsworn:

1. Confirm the sheet shows a Lay on Hands charge pill at 5/5.
2. Target a wounded token and click the feature card's activate button. The roll window opens even though the feature does not roll an attack, and the spend prompt is bounded 1 to 5.
3. Set the amount to 3 and roll. The card reports 3 healing and the pool drops to 2/5.
4. Click Apply Healing. The target gains exactly 3 HP.
5. Safe Rest. The pool returns to 5/5.
6. **The boundary worth checking:** spend all 5. Before this branch the charges vanished and no card was posted. It should heal 5 and leave the pool at 0.
7. Turn off Automation > Resource Spending and repeat step 3. The prompt still appears and the heal still lands, but the pool does not move.
8. Load a world stamped at schema version 47 with an existing Lay on Hands on a character and confirm Migration048 runs and backfills the pool.

## Checklist

## Related Issues

Closes #632
Part of #365

## Summary by CodeRabbit

- **New Features**
  - Added fixed- and variable-cost charge spending with bounded player-selected amounts.
  - Added pool-specific prompts, cost summaries, validation, conflict detection, and chat-card reporting.
  - Updated Lay on Hands with a level-scaled healing pool that refills after Safe Rest.
  - Added migration support for existing Lay on Hands features.

- **Localization**
  - Expanded English, French, and Spanish Nimble translations and rule labels.

- **Documentation**
  - Documented charge pools, variable spending, roll windows, healing formulas, and chat-card behavior.


trevlar (approved):
**✅ Approved**

Solid work. The variable cost mode is cleanly separated from the fixed path (`includeVariable` opt-in rather than a new consumer type), and the two bug fixes you folded in — deferring both charge and dice spends past the `preUseItem` gate, and reserving same-pool fixed costs before sizing the variable spend — are the right calls and are each pinned by a test.

- `tests/integration/chargeSpend.test.ts` driving the whole flow through real DOM clicks, including the spend-to-empty boundary, is exactly the coverage this needed.
- `Migration048` being conservative about a GM's own rules and formula edits is the right default, and `MigrationList.test.ts` closes the "added a migration without bumping the constant" gap for everyone after you.

Non-blocking notes, none of which need to hold the merge:

- **src/documents/item/spell.ts:95** — the spell card's flags don't carry `chargeConsumption` the way `base.svelte.ts:295` now does, so a spell with a variable consumer would spend the pool without the card reporting it. No shipped spell does this yet.
- **src/view/dialogs/itemActivationConfigDialogHelpers.ts:94** — a hidden pool (and a `maxCost` resolving below `cost` at :99) drops the prompt while `#hasVariableChargeSpend()` still forces the dialog open, so the item posts a card that heals 0 and spends nothing. The empty-pool case is caught by the gate; these two aren't. A refusal naming the pool, like the conflict path, would match AGENTS' "throw, don't no-op".
- **src/managers/ItemActivationManager.ts:687** — `#resolveSpentCharges` clamps `@spent` against live pool state but the deduction here uses the dialog's raw count, so a pool that moved under an open dialog heals for less than it spends. Deducting the same clamped number would keep the two in step.
- **tests/integration/liveHelpers.ts:218** — the old JSDoc about rule automation and health-state sync now sits above `getAutomationToggle`, and `getRuleAutomationEnabled` is left undocumented. Looks like the new block landed under the old one rather than replacing it.
- **docs/documentation/reference/_partials/chargeConsumer.md:16** — no trailing newline; every other partial in that directory has one.

PR #944, opened 2026-09-06
feat(rest): swap class options and move skill points on a rest

### LINKED ISSUE #938: [Foundation] Class option respec: re-pick a choice and ungrant what it created
> *This was generated by AI during triage.*

## Purpose

Foundational support for **re-picking a choice already made**: swapping one selected class option for another, and removing the item the first choice granted.

This is not a class feature. It is **one mechanic printed twelve times**, once per class, plus a general rule in the Core Rules that all twelve are instances of. Every class checklist carries it as an unticked box, and every one of those boxes is the same work.

The Core Rules state the general form as a Downtime Activity (CoreRules 2.0.2, "Downtime"):

> "**Retrain.** Spend time doing activities to retrain any of your chosen abilities, features, or if it makes sense in the story, possibly even your subclass."

Each class then prints a flavoured version, gated on a Safe Rest instead of Downtime. The wording is templated: "Whenever you *\<class flavoured activity\>* during a Safe Rest, you may choose different *\<Class\>* options available to you."

## Rules that need it

| Class | Feature | Checklist | Primarily re-picks |
|---|---|---|---|
| Hunter | Remember the Wild | #590 | Thrill of the Hunt abilities |
| The Cheat | Trade Secrets | #608 | Underhanded Abilities |
| Commander | Rigorous Training | #615 | Commander's Orders, Combat Tactics, Weapon Masteries |
| Mage | Study! | #627 | Spellshaper abilities, Elemental Mastery schools |
| Oathsworn | Serve Selflessly | #634 | Sacred Decrees, Master of Radiance spells |
| Shadowmancer | Supplicate | #643 | Lesser and Greater Invocations, Shadowmastery |
| Shepherd | Serve | #652 | Sacred Graces, Master of Twilight spells |
| Songweaver | Perform! | #660 | Lyrical Weaponry, A "People" Person, Windbag |
| Stormshifter | Be Wild | #667 | Chimeric Boons, Stormcaller spells |
| Zephyr | Focus | #676 | Martial Arts abilities |
| Berserker | Wrath & Ruin | #599 | Savage Arsenal picks |
| Artificer | Reconfigure! | not filed | Eureka! picks, Constant Tinkering upgrades |
| Hexbinder | Bubble, Bubble! | not filed | Afflictions, Mystic Marks, Familiar Boons |

Two related features share the machinery:

- **Jack of All Trades** (Songweaver L2): "When you Safe Rest, you may move a skill point as if you just leveled up." The only class feature that respecs a **core** character resource rather than class options.
- **More Skilled** (CoreRules, every level 1 to 20): "you may move 1 point from one skill to another (as long as the skill doesn't become negative)". The same skill point move, on level up rather than rest. This is the only level up re-pick in the entire corpus.

**Story-based subclasses** (Heroes 2.0.2) are the third case: "At your GM's discretion, these can be chosen at any point in a campaign when a story demands it, **replacing your existing subclass**." That applies to Oathbreaker, Spellblade, Reaver and Beastmaster.

None of the twelve respec features exists in `packs/` today. Only Jack of All Trades is shipped.

## Half of this already exists

**Level down already does the ungrant.** `levelUpHistory` entries record `grantedFeatureIds`, `grantedSpellIds` and `poolMaxBonuses`; `CharacterLevelDownDialogState.svelte.ts` previews exactly what will be removed, including pool max bonuses that are not distinct items; and `character.ts:1860` deletes the granted items via `deleteEmbeddedDocuments`.

A respec is close to: find the history entry that granted the option, delete that item, grant the replacement, and update the entry. The hard part, knowing what a past choice created so it can be undone, is already solved and already has a preview UI to model the new one on.

## What can be re-picked: the real design decision

"Different *\<Class\>* options" is not precise, and the Core Rules version is broader still ("any of your chosen abilities, features, or ... possibly even your subclass"). This scope question should be answered **once, here**, rather than eleven times across the class checklists.

The selections a respec could plausibly target:

**Per class, the option lists** (picks / pool size):

| Class | Selection | Picks | Pool |
|---|---|---|---|
| Berserker | Savage Arsenal | 7 | 12 |
| The Cheat | Underhanded Abilities | 8 | 10 |
| Commander | Commander's Orders / Combat Tactics / Weapon Mastery | 2 / up to 6 / 2 | 5 / 5 / 3 |
| Hunter | Thrill of the Hunt | 7 | 14 |
| Mage | Spellshaper / Elemental Mastery | 4 / 3 | 8 / schools |
| Oathsworn | Sacred Decrees | 6 | 10 |
| Shadowmancer | Lesser / Greater Invocations | 3 / 5 | 10 / 11 |
| Shepherd | Sacred Graces | 4 | 8 |
| Songweaver | Lyrical Weaponry / A "People" Person | 4 / 2 | 5 / 4 |
| Stormshifter | Chimeric Boons | 5 (7 with Fang & Claw L15) | 9 |
| Zephyr | Martial Arts | 8 | 11 |

**Cross class:** subclass (1, at L3), Key Stat Increases (4), Secondary Stat Increases (4), the L20 "+1 to any 2 stats", Epic Boon (1 of 12, at L19), skill points, languages, ancestry and background.

Note that the Cheat picks **8 of 10** and the Songweaver picks **4 of 5**. At those ratios a respec is nearly a no-op, which is worth knowing before building elaborate selection UI.

## Automate bookkeeping, not the decision

Per CLAUDE.md guideline 7, this feature divides unusually cleanly.

**The trigger can never fire automatically.** "A notable act of destruction", "a notable selfless act", "a day with wild animals": every one is a GM judgement about the fiction. There is no signal to detect and the system must not try to infer one. This is a player or GM initiated control that happens to be gated to Safe Rests.

**Shadowmancer's Supplicate is explicitly refusable**: "you may **beg** them to allow you to choose different Shadowmancer options (**they may ask for something in return**)". The Patron's answer is the GM's, and the price is the GM's. The system offers the swap; it does not grant it.

**What the system should do** is the bookkeeping nobody wants to do by hand: show which options are currently taken and which are available, enforce the pool and the pick count, delete the old item with its rules cleanly, grant the new one, and keep `levelUpHistory` truthful so level down still works afterwards.

## What a solution must answer

- **Scope.** Class option lists only, or also subclass, stat increases, Epic Boon and skill points? The Core Rules Retrain entry is the broadest reading; the class features say "options". Answer once.
- **Prerequisites between options.** The Cheat's Sunder Armor (Heavy) requires Sunder Armor (Medium). Swapping away the prerequisite must be blocked or must cascade.
- **Picks that changed a stat.** Shadowmancer's Fiendish Boon raises DEX or INT by 1 permanently and costs a Hit Die. Unwinding it is not just an item delete.
- **Picks that were taken instead of something else.** Commander's Combat Tactic levels offer "choose another Combat Ability **or** gain +1 max Combat Dice", and the bonus is recorded in `poolMaxBonuses` rather than as an item. The level down preview already handles this case and is the model to follow.
- **Picks that consumed other picks.** Hunter's Beastmaster consumes 2 of the 7 Thrill of the Hunt slots.
- **Where the control lives.** The Safe Rest card (`src/view/chat/SafeRestCard.svelte`) already exists, as does the level up dialog. A respec is closer to the level up dialog's selection UI than to anything on the rest card, so the likely shape is a rest card control that opens a scoped version of the existing dialog.
- **Does it need the class feature at all?** The Core Rules Retrain downtime activity applies to every character regardless of class. The twelve class features arguably only move it from Downtime to a Safe Rest. If the control ships unconditionally, the class features become `note` items and this closes twelve checklist boxes rather than one.

## Blocks

#590, #599, #608, #615, #627, #634, #643, #652, #660, #667, #676

Artificer and Hexbinder are not shipped as classes yet, so their two variants have no checklist.

## Current behaviour

No respec exists. None of the twelve features is in `packs/`. The option pools themselves are all shipped and their sizes match the books.

### Unrelated bug found while scoping this

`packs/classFeatures/core/songweaver/songweaver-progression/jack-of-all-trades.json` has the wrong activation:

```json
"cost": { "details": "An ally makes a roll", "quantity": 1, "type": "none", "isReaction": true }
```

The feature is "When you Safe Rest, you may move a skill point as if you just leveled up." There is no reaction, no trigger and no ally. The cost block looks copied from Songweaver's Inspiration. The description is correct; only the activation is wrong.

Fronix:
## Summary

Several classes let a character re-pick the class options they chose at earlier levels, mostly on a Safe Rest. This adds that surface: the rest dialogs offer the pools the character has picks in, apply the swap when the rest is confirmed, and report what changed on the chat card. Closes #938.

## Screenshots
image

### Choosing a homebrewed order for Commander
image

### Homebrewed order offering to move a skill point
image

### Chat note after safe rest
image

## Changes

- **Three declarative rule types.** `optionSwap` (which pools, on which rest), `skillPointMove` (how many skill points may move), and `poolMaxBonus` (raise a charge pool's maximum). They are read on demand rather than through a lifecycle hook.
- **Pool bonus picks are ordinary items.** The Commander's "+1 Max Combat Die" was a number on the level-up history plus one renamed cosmetic item. It is now one granted item per pick carrying a `poolMaxBonus` rule, so a pick can be traded like any other. Migration 048 converts existing characters.
- **Alternatives fold into one pool.** A level offering a choice between options contributes every alternative, including items an option grants outright, and pools that share a group merge. That is what makes the die tradeable for a Combat Ability and back.
- **The rest dialogs.** A "Change my options" row on Safe Rest and Field Rest, quoting the feature that offers the swap. Each pool shows the held picks with the rest of the pool folded behind a "Show more options" row. A single-pick pool swaps on one click. Skill points move net zero, floored so a skill's bonus never goes below +0 and capped at +12.
- **Applying and reporting.** A replacement inherits the level-up history entry of the pick it replaces, so level down needs no knowledge that a swap happened. Nothing is written until the rest is confirmed. The chat card lists what changed. A grant that cannot be resolved aborts the whole swap rather than deleting a pick with nothing to put back.
- **Content.** Eleven class features gain an `optionSwap` rule (Berserker, Commander, Hunter, Mage, Oathsworn, Shadowmancer, Shepherd, Songweaver, Stormshifter, The Cheat, Zephyr). Jack of All Trades gains its `skillPointMove` rule and its activation is corrected.
- **Migrations 048, 049, 050.** Convert pool bonuses to items, grant the swap features to existing characters past the printed level, and bring world and embedded copies of Jack of All Trades in line.
- **Docs.** `playing/rest-and-recovery.md` and rule reference partials for the three new rule types.

## Test Plan

Automated:

- `pnpm check` passes (format, lint, circular deps, type-check, 4160 unit tests).
- `pnpm test:integration optionSwap` passes 12 of 12 against a live v14 world. It drives the real dialogs on a Commander, a Berserker, and a Songweaver. Note that Foundry serves the worktree's `dist`, which `pnpm dev` does not rebuild, so run `pnpm build` first or point at the Vite dev server with `FOUNDRY_URL=http://localhost:30001`.

By hand, on a Commander at level 10 or higher who holds Rigorous Training:

1. Open the character sheet and click Safe Rest. The dialog is wider than usual and shows a "Change my options" row.
2. Expand it. Rigorous Training's own text is quoted, and one pool named "Fit for Any Battlefield" holds the Orders, the Tactics, and any "+1 Max Combat Die" picks. Weapon Mastery is its own pool.
3. Click "Show N more options" on a pool. In a single-pick pool, clicking another option swaps it in. In a multi-pick pool, the other cards stay greyed until you deselect one of yours.
4. Trade a die for a Combat Ability and confirm the rest. One die item is removed, the ability is granted, the Combat Dice maximum drops by one, and the chat card lists the change.
5. Close the dialog without resting after making a change: nothing on the sheet moves.
6. Level down past the level that granted the replacement: the replacement is removed, not the option that used to sit there.
7. On a Songweaver with Jack of All Trades, take a point from one skill and give it to another. The give button stays disabled until a point has been taken, a skill holding no points cannot give one, and the card lists both halves.

Migration check: open a world created before this branch with a Commander who took the max Combat Die option. The number becomes one item per point and the maximum is unchanged.

## Checklist

## Related Issues

Closes #938

## Notes for the reviewer

- Migration number 048 is also claimed on `feat/lay-on-hands-pool` and `feat/shadowmancer-pilfered-power`. Whichever merges second needs renumbering.
- Migration tests are gitignored by repo policy, so 048, 049, and 050 ship without committed tests. They were written and run locally.
- Merging pools means a Commander can trade a level 2 Order for a Tactic. The rulebook says "different Commander options available to you" without naming a pool, so the system offers it and the table decides.
- A second copy of "+1 Max Combat Die" cannot be added through the swap: picks are identified by source, so the held copy stands for the die. An integration test pins this.

## Summary by CodeRabbit

* **New Features**
  * Added Safe Rest and Field Rest controls for swapping eligible class features and reallocating skill points.
  * Added class-specific option-swap features across multiple core classes, triggered by relevant rest activities.
  * Added “Jack of All Trades” skill-point transfers during Safe Rest.
  * Added support for increasing maximum combat dice through granted features and repeatable picks.
  * Rest chat cards now show added and removed options.
* **Documentation**
  * Expanded guidance for option swaps, skill-point moves, repeatable choices, and combat-die bonuses.
* **Bug Fixes**
  * Improved validation, rollback, migration handling, and charge-spending behavior.


docs/documentation/reference/_partials/poolMaxBonus.md
```diff
@@ -0,0 +1,12 @@
...
+- **Pool identifier** → `combat-dice`
+- **Amount** → `1`
+
+The bonus lives on the item, so a Commander who takes the option three times holds three items and has three more maximum Combat Dice. Removing one item removes exactly its share.
+
+This is why the bonus is an item rather than a number on the character: it makes the pick behave like every other granted item. It is recorded against the level that granted it, so levelling down removes it cleanly, and the level-up dialog grants it through the same path as any other option.
+
+Note that the bonus item is not itself a member of a swappable pool, so **Option Swap** does not offer to trade it for a Combat Ability. It sits in the class progression group rather than in the ability pool.
+
+The pool identifier must match the `identifier` of a **Charge Pool** rule somewhere on the character. A bonus naming a pool that does not exist does nothing. To change a pool's die size or its current value rather than its maximum, use **Modify Pool**.
```
trevlar, line 12: This says the bonus item is not part of a swappable pool, so **Option Swap** cannot trade it for a Combat Ability. That’s the opposite of both the implementation and the rulebook.

  The Commander’s level 6, 8, 10, 12, and 16 entries each say, “Choose another Combat Ability or gain +1 max Combat Dice,” so these are intentionally alternatives at the same level. `fit-for-any-battlefield.json` represents that correctly: `max-combat-die` and `combat-ability` share the same `applyAtLevels`. Because `collectPoolRequirements` uses `alternatives: 'union'`, it combines their groups and includes the die’s `grantItem` UUID in `extraCandidateUuids`. That puts the die in the same option pool as Orders and Tactics.

  The integration test at `tests/integration/optionSwap.test.ts:313` swaps it, and `rest-and-recovery.md` says it can be swapped as well. Since this partial is included in the generated player reference, it currently gives the wrong answer to the exact question players are likely to ask.

  Rewrite the paragraph to say the die is part of the option pool. The actual limitation is that a held copy represents the die, so swapping cannot add a second copy.

src/view/dialogs/components/characterCreator/FeatureGroupSelection.svelte.ts
```diff
@@ -3,20 +3,12 @@ import type {
...
- * Converts kebab-case to Title Case
- * e.g., "thrill-of-the-hunt" -> "Thrill Of The Hunt"
- */
-export function formatGroupName(name: string): string {
-	return name
-		.split('-')
-		.map((word) => word.charAt(0).toUpperCase() + word.slice(1))
-		.join(' ');
-}
+export { formatGroupName };
```
trevlar: `export { formatGroupName };` only re-exports the promoted utility, and nothing imports it from this file. Delete the export and keep the local import from `#utils/formatGroupName.js`.

src/view/sheets/pages/ClassProgressionTabUtils.ts
```diff
@@ -7,9 +7,7 @@ import { ABILITY_SCORE_LEVELS, SUBCLASS_LEVELS } from './ClassProgressionTabCons
 
 type AbilityScoreRecord = { [K in (typeof ABILITY_SCORE_LEVELS)[number]]: AbilityScoreDataEntry };
 
-export function formatGroupName(groupName: string): string {
-	return groupName.replace(/-/g, ' ').replace(/\b\w/g, (char) => char.toUpperCase());
-}
+export { default as formatGroupName } from '#utils/formatGroupName.js';
```
trevlar: This is the same kind of shim, but it still has three consumers:

  * `ClassProgressionLevelRow.state.svelte.ts:3`
  * `ClassProgressionLevelRowUtils.ts:6`
  * `ClassProgressionTab.state.svelte.ts:17`

  `AGENTS.md` lists “No barrel exports” as a critical rule, and the style guide says a promotion should update all existing call sites. Point those imports directly to `#utils/formatGroupName.js`, then remove the re-export.

src/utils/collectSwappableOptions.ts
```diff
@@ -0,0 +1,141 @@
+import collectPoolCandidates from '#utils/collectPoolCandidates.ts';
+import collectPoolRequirements, {
+	buildPoolKey,
+	type PoolRequirement,
+} from '#utils/collectPoolRequirements.ts';
+import type { ClassFeatureIndex } from '#utils/getClassFeatures.ts';
+
+/** One pool of class options a character may re-pick from. */
+export interface SwappableOptionPool {
```
trevlar, line 11: “Pool” now refers to two different concepts: charge or dice pools, represented by names such as `chargePool`, `modifyPool`, `poolIdentifier`, and `poolMaxBonus`; and the new option pools represented by `SwappableOptionPool`, `poolKey`, `poolGroups`, and `collectPoolCandidates`.

  `poolMaxBonus` belongs to both concepts: it modifies a charge pool while also participating in an option pool. That overlap may be what led to the incorrect documentation above. The option-pool terminology already existed privately in `findMissingLevelSelections`, but this PR promotes it to `src/utils/` and the Shared Code Inventory, where it now sits beside `chargePool` without anything distinguishing the two.

  Names such as `OptionPool` and `collectOptionPoolCandidates` would make the distinction clear. This is inexpensive to change now and will become harder once more callers depend on the current names.

src/utils/summarizeOptionSwap.ts
```diff
@@ -0,0 +1,57 @@
...
+		);
+		const nameOf = (uuid: string) => nameByUuid.get(uuid) || uuid;
+
+		const removed = pool.ownedUuids.filter((uuid) => !wanted.has(uuid)).map(nameOf);
+		const added = selected.filter((uuid) => !owned.has(uuid)).map(nameOf);
+		if (removed.length === 0 && added.length === 0) continue;
+
+		changes.push({
+			// The same heading the rest dialog showed the pool under.
+			label: pool.displayName || formatGroupName(pool.poolKey),
```
trevlar: `formatGroupName(pool.poolKey)` receives a `+`-joined key rather than a single group name. A merged pool without a `displayName` renders as `Commander Orders+commander Tactics`.

  Format each entry in `pool.poolGroups` and join the results, or fall back to the first group. The same issue appears in `src/view/dialogs/components/optionSwap/OptionSwapSection.state.svelte.ts:91`.

src/documents/actor/character.ts
```diff
@@ -1659,105 +1837,32 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
...
-						continue;
-
-					const grantUuid = rule.grantItemUuid;
-					// Match the full compendium UUID exactly — a bare id substring can false-match.
-					const alreadyOwned = this.items.some(
-						(owned) =>
-							((owned as unknown as { _stats?: { compendiumSource?: string } })._stats
-								?.compendiumSource ?? '') === grantUuid,
-					);
+	async #applyRestOptionSwap(restData: RestManager.Data): Promise<OptionChange[]> {
```
trevlar, line 1842: `(restData as { optionSwap?: OptionSwapSubmitData }).optionSwap` casts around a field the dialogs genuinely submit. `RestManager.Data` already declares `optionChanges`, so add `optionSwap?: OptionSwapSubmitData` beside it and keep the dialog-to-manager contract typed. `types/optionSwap.d.ts` is type-only, so this should not reintroduce the cycle fixed in `e335e848`.

src/migration/migrations/Migration049GrantOptionSwapFeatures.ts
```diff
@@ -0,0 +1,231 @@
+import { toInstalledId, toSnapshotId } from '../compendiumSourceId.js';
+import { MigrationBase } from '../MigrationBase.js';
+
+interface SwapFeature {
+	uuid: string;
+	level: number;
+	name: string;
+	img: string;
+	folder: string;
```
trevlar: `SwapFeature` declares `folder`, and all 11 class entries provide it, but line 220 creates the item with `folder: null`. Either use the supplied value or remove the unused field.

trevlar (dismissed):
**🔧 Push back: one shipped doc partial contradicts the implementation, and two re-export shims violate the style guide**

The mechanic itself looks solid. I traced the swap through the full flow, from the offer and pool collection through planning, application, history updates, and the character card. I also checked all 11 new class features against the rulebook and didn’t find any structural issues.

**What looks good**

* The grant-before-remove order, the rollback in `#undoOptionSwap`, and the short-create guard all protect against the failure that could actually cost a player an option. That path is handled well.

* All 11 features match the rulebook’s levels and wording. Having `planOptionSwap` inherit the vacated `levelUpHistory` entry also means level-down behavior works without additional changes.

**Requested changes**

1. Correct `poolMaxBonus.md` to say the die is swappable.
2. Remove the unused `formatGroupName` re-export.
3. Update the three `ClassProgressionTabUtils` consumers and remove that re-export.
4. Distinguish charge pools from option pools in the naming.
5. Format merged pools from their individual group names.
6. Add `optionSwap` to `RestManager.Data`.
7. Either use or remove `SwapFeature.folder`.

docs/documentation/reference/_partials/poolMaxBonus.md
Fronix: Fixed in 0a6813d1. The paragraph now says the die shares the option pool with the Combat Abilities and can be swapped either way, and that a held copy counts as one pick so a swap cannot add a second copy.

src/view/dialogs/components/characterCreator/FeatureGroupSelection.svelte.ts
Fronix: Fixed in 63690033. The re-export is gone; the file keeps its local import.

src/view/sheets/pages/ClassProgressionTabUtils.ts
Fronix: Fixed in 00a80194. The three consumers import from `#utils/formatGroupName.js` directly and the re-export is removed.

src/utils/collectSwappableOptions.ts
Fronix: Done in 8ac90628. The promoted utilities now carry the name: `collectOptionPoolRequirements`, `collectOptionPoolCandidates`, `OptionPoolRequirement`, `buildOptionPoolKey`, with the inventory rows updated. I left the `poolKey` and `poolGroups` fields as they are, since they predate this PR in `findMissingLevelSelections` and the level correction dialog.

src/utils/summarizeOptionSwap.ts
Fronix: Fixed in 83a76bf9 and e0c80c51. The collector now always sets `displayName`, falling back to the formatted group names joined with a slash, so the summarizer and the dialog state both read `pool.displayName` and cannot drift apart. Covered by a collector test that blanks the Commander feature name.

src/documents/actor/character.ts
Fronix: Fixed in 6450794b. `RestManager.Data` declares `optionSwap` and the cast is gone. The constructor now spreads the defaults instead of using `mergeObject`, since its typing rejected the Map inside the payload; the defaults are three scalars so the behaviour is the same.

src/migration/migrations/Migration049GrantOptionSwapFeatures.ts
Fronix: Fixed in 1895b15a. The field is removed from `SwapFeature` and all eleven entries. The item is embedded on the actor, so `folder: null` is correct there.

src/documents/actor/character.ts
```diff
@@ -1835,15 +1936,34 @@ export class NimbleCharacter extends NimbleBaseActor<'character'> {
...
 		});
 
+		if (unreverted.length > 0) {
+			ui.notifications?.warn(
+				localize('NIMBLE.levelDownDialog.skillPointsMovedAway', {
+					skills: unreverted.join(', '),
+				}),
+			);
+		}
+
```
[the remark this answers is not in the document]
Fronix: Not sure we need to deal with this, leveling down is only really something you do when you want to repick something because of a mistake. If it shows up as an issue later we'll deal with it.

trevlar (changes requested):
**🔧 Changes needed — pool bonus picks should update one item, not stack up copies**

**Must fix before merge**

- **src/documents/actor/character.ts:1890** — Change this so a pool bonus lives on exactly one item: create the item if the character doesn't have it, add to its `amount` if they do, delete it when the amount hits zero. Right now this loop creates a new item for every pick, so a Commander who takes `+1 Max Combat Die` at levels 4, 6 and 8 ends up with three identical rows on their sheet. Two things follow from the change. Level-down needs to know how many points a given level added to a shared item, so record that on the history entry (an item id on its own can't say). And the swap has to decrement the amount instead of deleting the item, or trading one die away would wipe all three points at once.

- **src/utils/summarizeOptionSwap.ts:51** — The skill name is printed twice on the chat card. `label` is the skill name, and the string you push into `added`/`removed` starts with the same name, so `OptionChangesList` renders "Stealth" in the label column and "Stealth 3 to 2" right next to it. Use `` `${from} to ${to}` `` for the value; the label column already says which skill it is. `summarizeOptionSwap.test.ts:80` locks in the current wording, so update that expectation too.

- **src/documents/actor/character.ts:1418** — The delete and the actor update below are wrapped in a try/catch that rolls back and notifies. This create isn't. If `createEmbeddedDocuments` rejects, the error escapes `applyOptionSwap`, then `#applyRestOptionSwap`, then `triggerRest`, and nothing catches it: `PlayerCharacterSheet.svelte:409` and `:420` call it bare. The player clicks Safe Rest and nothing happens at all. No swap, no rest, no chat card, no error. Wrap it like the writes below it.

- **src/documents/actor/character.ts:1452** — This fires on every rest where the dialog was shown, even when the player changed nothing. The plan comes back empty, `levelUpHistory` maps to an identical array, and you still call `this.update()` and force `sheet.render(true)`. Return early when the plan is empty and `skillPoints` is empty.

**What's good**

- The failure handling in `applyOptionSwap` is careful: grant before delete, a short create result treated as failure, a survivor check after the delete, and a rollback that keeps going past its own first error.
- Test coverage matches the size of the change. Twelve integration tests drive the real dialogs, and the unit tests cover the pool merge, the pick count, the history inheritance and both skill floors.

---

1. One pool bonus item: create, increment, delete at zero
2. Record per-level bonus points so level-down can decrement
3. Make the swap decrement the amount, not delete the item
4. Stop printing the skill name twice on the chat card
5. Update the summarizeOptionSwap expectation to match
6. Catch a failed grant like the later writes are caught
7. Skip the actor update when nothing changed

Fronix:
> Change this so a pool bonus lives on exactly one item: create the item if the character doesn't have it, add to its `amount` if they do, delete it when the amount hits zero. Right now this loop creates a new item for every pick, so a Commander who takes `+1 Max Combat Die` at levels 4, 6 and 8 ends up with three identical rows on their sheet. Two things follow from the change. Level-down needs to know how many points a given level added to a shared item, so record that on the history entry (an item id on its own can't say). And the swap has to decrement the amount instead of deleting the item, or trading one die away would wipe all three points at once.

Hmm, yeah I did not expect this to have that type of effect on just normal leveling up. I'll think a bit and see if there isn't a more clean way.

We can't keep a history of everything, I'm pretty sure that could have some real bad effects later down the line when we have even more items and more complex classes. I need to find a way to do this without I think, mostly because each item should be it's own history, so if you level down, that item just gets removed and there is nothing that needs to be recalculated since it contains everything it gives.

trevlar (commented):
**🔧 Push back — two problems in the swap surface, both in the display layer**

The mechanic underneath is solid. The pool merging, the pick accounting and the plan/apply split all hold up, and the migration ids line up with the pack JSON. Two things in front of the player need a fix first.

**Must fix before merge**

- **src/view/dialogs/components/optionSwap/OptionSwapSection.svelte:176** — Unfolding a pool's alternatives doesn't call `onToggle`, so `fitWindow` never runs and `dialog.setPosition()` is never called. That's the growth that actually matters: the section toggle adds a couple of rows, but "Show 9 more options" adds nine cards. On a Commander at level 10 the window runs past the bottom of the screen and takes the footer with it, so the rest can't be confirmed without moving the window by hand. `state.releaseLastPick` hits the same path through `unfoldAvailable`. Call `onToggle` from both.
- **src/utils/summarizeOptionSwap.ts:70** — The skill name goes into the row label and again into the value text, so the chat card renders "Stealth" next to "Stealth 3 to 2". The connector is also a bare English literal, which the localize rule on the checklist rules out. Use the internationalized version for compendium options and use the name as a fallback value. Put the range behind a key, something like `NIMBLE.optionSwap.skillPointRange` with `{from}` and `{to}`. `summarizeOptionSwap.test.ts:123` pins the current strings, so it moves with the fix.

- **src/documents/actor/character.ts:1453** — `applyOptionSwap` runs on every rest where the section was shown, even when the player changed nothing, and it always writes the whole `system.levelUpHistory` array back plus a `sheet.render(true)`. An ordinary safe rest for anyone holding one of the eleven new features now costs an extra actor update. Return early when the plan has no grants, no deletions and no skill points.

**What's good**

- Giving a replacement the `levelUpHistory` entry of the pick it replaces is the right call. Level down stays ignorant that a swap happened, which is why `revertLastLevelUp` needed nothing but the skill clamp.
- Refusing a pool whose released pick has left the history, rather than granting a replacement no level records, is the failure mode I'd have gone looking for. Nice that `planOptionSwap.test.ts:204` pins it.

---

1. Refit the window when a pool unfolds
2. Stop printing the skill name twice, localize the range
3. Skip the write when nothing changed

Fronix:
@trevlar closing this PR in favour of a new one. The PR has become convoluted and drifted

PR #951, opened 2026-09-10
feat(rest): swap class options and move skill points on a rest

### LINKED ISSUE #938: [Foundation] Class option respec: re-pick a choice and ungrant what it created

Fronix:
## Summary

Several classes let a character re-pick the class options they chose at earlier levels, mostly on a Safe Rest. This adds that surface: the rest dialogs offer the pools the character holds picks in, apply the swap when the rest is confirmed, and report what changed on the chat card. It replaces #944, which was closed after the layout and the holdings model were reworked. Closes #938.

## Sceenshots & Video
image
image

https://github.com/user-attachments/assets/35cac5d8-aad0-4b83-a854-fd3b88655c01

## Changes

- **Three declarative rule types.** `optionSwap` (which pools, on which rest), `skillPointMove` (how many skill points may move), and `poolMaxBonus` (raise a charge pool's maximum). They are read on demand rather than through a lifecycle hook.
- **A pick is an item on the sheet.** The window reads what the sheet holds, whose compendium source is a pool member, however it got there: a level up, the level correction dialog, a drag from the compendium, or a GM's hand. The class level says how many picks each pool grants. The level-up history is used only to order copies, never to decide what is held.
- **Pool bonus picks are ordinary items.** The Commander's "+1 Max Combat Die" is one granted item per pick carrying a `poolMaxBonus` rule, so a pick can be traded like any other and can be held more than once when its grant allows duplicates.
- **Alternatives fold into one pool.** A level offering a choice between options contributes every alternative, including items an option grants outright, and pools that share a group merge so no pick counts twice. The pool's groups keep the order the levels first name them.
- **The rest dialogs.** A "Change my options" row on Safe Rest and Field Rest. One card per feature that makes the offer, with its icon, name and what it offers. A pool is a fixed set of places, one chip per pick; giving a pick up leaves an empty place, and a choice from the fold below fills the first empty one. A pool whose members come from more than one option list shows one small heading per list, with the count kept for the whole pool. Resting the pointer on a chip or a card icon opens a hover card with the description. A skill point moves through two selects, net zero. The confirm button names the swap while one is ready.
- **Applying and reporting.** A swap is a one-for-one trade, or a fill when the sheet holds fewer picks than the levels grant. Holdings are re-read at apply time. A replacement inherits the level-up history entry of the pick it replaces where one exists. Nothing is written until the rest is confirmed, and a grant that cannot be resolved aborts the swap. The chat card lists what changed, naming a repeated option once with its count.
- **Content.** Eleven class features gain an `optionSwap` rule (Berserker, Commander, Hunter, Mage, Oathsworn, Shadowmancer, Shepherd, Songweaver, Stormshifter, The Cheat, Zephyr). Jack of All Trades gains its `skillPointMove` rule and its activation is corrected.
- **Migrations 049, 050, 051.** Convert pool bonuses to items, grant the swap features to existing characters past the printed level, and bring world and embedded copies of Jack of All Trades in line.
- **Docs.** `playing/rest-and-recovery.md`, the reference partials for the three rule types, and the style guide inventory row for the offer resolver.

## Test Plan

Automated:

- `pnpm check` passes (format, lint, circular deps, type-check, 4283 unit tests).
- `pnpm test:integration optionSwap` passes 27 of 27 against a live v14 world. It drives the real dialogs on a Commander, a Berserker and a Songweaver, including the hover card, the given-up card, a hand-built sheet, and a hand-dragged duplicate. Foundry serves the worktree's `dist`, which `pnpm dev` does not rebuild, so run `pnpm build` first or point at the Vite dev server with `FOUNDRY_URL=http://localhost:30001`.

By hand, on a Commander at level 8 or higher who holds Rigorous Training:

1. Open the character sheet and click Safe Rest. The window shows a "Change my options" row.
2. Expand it. One card, Rigorous Training, holds the pool "Fit for Any Battlefield" with the held picks as chips under the headings Commanders Orders, Combat Tactics and +1 Max Combat Die, and the count chosen / granted. Weapon Mastery is its own pool.
3. Rest the pointer on a chip. A card with the option's icon, name and description opens beside it and closes when the pointer leaves.
4. Click the check on a die chip. The chip becomes "Empty. Choose below." and the other options unfold, split by the same headings. Click an ability. It takes the empty place, the status line reads "You give up +1 Max Combat Die and take ...", and the footer reads "Safe Rest and swap options".
5. Confirm the rest. One die item is removed, the ability is granted, the Combat Dice maximum drops by one, and the chat card lists the change.
6. Close the window without resting after making a change: nothing on the sheet moves.
7. Drag a second "+1 Max Combat Die" onto the sheet from the compendium and open the window again. It shows as one more place, and the count reads one over the grant with a sentence saying so. Nothing blocks the rest.
8. On a Songweaver with Jack of All Trades, choose a skill to give the point and a skill to take it. The second select stays disabled until the first is chosen, a skill holding no points cannot give one, no skill goes past +12, and the card lists both halves.

Migration check: open a world created before this branch with a Commander who took the max Combat Die option. The number becomes one item per point and the maximum is unchanged.

## Checklist

## Related Issues

Closes #938
Replaces #944

## Notes for the reviewer

- Migration numbers 049 to 051 follow Lay on Hands at 048 on `dev`. Any other open branch that claims 049 needs renumbering when it merges second.
- Migration tests are gitignored by repo policy, so 049 to 051 ship without committed tests. They were written and run locally.
- The rulebook says "different <Class> options available to you" and never says a new pick must be the same kind as the one given up. The window therefore counts a pool as one total, shows the lists as headings, and does not warn when a Commander trades an Order for a Tactic.
- The list headings come from the group key, so "Commander's Orders" reads "Commanders Orders". The feature data model has no field for the label the packs carry; adding one is a small follow-up.
- Every `NIMBLE.optionSwap` string in `public/lang/en.json` is new or reworded in this PR and would benefit from a read.


src/view/dialogs/components/optionSwap/OptionSwapSection.state.svelte.ts
```diff
@@ -0,0 +1,573 @@
...
+ */
+export function createOptionSwapSectionState(getProps: () => OptionSwapSectionStateProps) {
+	let isExpanded = $state(false);
+	let unfoldedPools = $state<Set<string>>(new Set());
+	let placesByPool = $state<Map<string, OptionSwapPlaceEntry[]>>(new Map());
+	let skillMove = $state<{ from: string; to: string }>({ from: '', to: '' });
+
+	const pools = $derived(getProps().offer?.pools ?? []);
+	const sources = $derived(getProps().offer?.sources ?? []);
+	const skillBudget = $derived(getProps().offer?.skillPoints ?? 0);
```
trevlar, line 87: `skillBudget` is read here and then only ever tested against zero. The window holds a single `{from, to}` pair, so it moves exactly one point no matter what the rules add up to. Two features granting a point each, or one rule authored with `points: 2`, still gets the player one move. The reference partial you shipped says the opposite: "Several rules on one character add up, so a character with two such features may move two points." Either honor the budget with one move per point, or drop `points` from the schema and that sentence from the docs. AGENTS.md is explicit that an unsupported path should throw rather than quietly no-op.

src/view/dialogs/components/optionSwap/OptionSwapSection.svelte
```diff
@@ -0,0 +1,647 @@
...
+
+{#snippet placeChips(view: OptionSwapPoolView, places: OptionSwapPlace[])}
+	<ul class="nimble-option-swap__chips">
+		{#each places as place (place.index)}
+			<!-- svelte-ignore a11y_no_noninteractive_element_interactions -->
+			<li
+				class="nimble-option-swap__chip"
+				class:nimble-option-swap__chip--held={place.feature}
+				class:nimble-option-swap__chip--empty={!place.feature}
+				onmouseenter={(event) =>
```
trevlar, line 104: The hover card only opens on `onmouseenter`. A player tabbing through the chips gets the option's name from the `aria-label` but never its description, and there's no touch path at all. `onfocusin` / `onfocusout` on the same element, alongside the mouse handlers, covers both.

trevlar (approved):
`skillBudget` is read here and then only ever tested against zero. The window holds a single `{from, to}` pair, so it moves exactly one point no matter what the rules add up to. Two features granting a point each, or one rule authored with `points: 2`, still gets the player one move. The reference partial you shipped says the opposite: "Several rules on one character add up, so a character with two such features may move two points." Either honor the budget with one move per point, or drop `points` from the schema and that sentence from the docs. AGENTS.md is explicit that an unsupported path should throw rather than quietly no-op.

Issue #613, opened 2026-04-08
[Automation] Commander - Coordinated Strike!: shared free-attack Order with regen


Fronix:
## Context
Commander's signature Level 1 Order and the mechanical spine of the class: a shared free-action weapon attack for the Commander and an ally, gated by an INT/Safe Rest counter, with regeneration and rider effects added at higher levels. Multi-actor free action; needs its own thread because several later features modify it.

Pack file: `packs/classFeatures/core/commander/commanders-order/coordinated-strike.json`

## Rules from Rulebook

> "Coordinated Strike! (1/round) Free action: you and an ally within 6 spaces both immediately make a weapon attack or cast a centrip for free. You can do this INT times/Safe Rest."

> "LEVEL 5 Master Commander. When you roll Initiative, regain 1 spent use of Coordinated Strike (it is lost if not spent during that encounter). Attacks made from your Coordinated Strikes also now ignore disadvantage."

> "LEVEL 9/13/17 Master Commander(2-4). +1 use of Coordinated Strike/Safe Rest."

> "LEVEL 18 Unparalleled Tactics. The first time each encounter you use Coordinated Strike, an ally who can hear you also gains 1 action to use on their next turn."

> "LEVEL 20 Captain of Legions. The first time each encounter you use Coordinated Strike, EVERY ally within 12 spaces gains +1 action (replaces Unparalleled Tactics)."

## Implementation
- `resourceCounter` (per Safe Rest) seeded to INT mod, scaling +1 at L9/L13/L17
- (1/round) lockout on Coordinated Strike use
- Targeting: pick ally within 6 spaces; prompt both to resolve attack/centrip
- L5 Master Commander hook: on-initiative regain 1 charge (ephemeral, lost if unused)
- L5 ignore-disadvantage flag on the resulting attacks
- L18 Unparalleled Tactics: first-per-encounter grant action to one ally (grantAction)
- L20 Captain of Legions: first-per-encounter grant action to ALL allies within 12 (multi-target grantAction); supersedes L18
- Vanguard subclass L7 Experienced Commander adds +1 use and +1 ally target
- Vanguard subclass L11 Survey the Battlefield adds additional regain-on-initiative
- Bulwark subclass L7 Juggernaut adds damage = armor + 1 to primary die
- Spellblade subclass Withering Strike overlays necrotic + undead-tag rider

## Rule Types
- `resourceCounter` (#576)
- `grantAction` (#582)
[...]

PR #957, opened 2026-09-13
feat(commander): Combat Dice expiry, Hold the Line! healing, two pack fixes, Migration055

Fronix:
## Summary

Commander class branch under the class checklist triage rules: the Combat Dice expiry and Hold the Line! healing that shipped rules can do, two pack bug fixes, pack assertions, and Migration055 so existing characters get the same changes.

## Changes

- Fit for Any Battlefield: the `combat-dice` pool gains an `encounterEnd` recovery that sets it to 0. Combat Dice are lost when combat ends.
- Hold the Line!: the warning-only card gets a `3 * @level` healing node. On a creature at 0 HP that gives the same result as "set their HP to 3x your LVL", and a reminder note on the card names the real effect. The "(M)" marker leaves the description.
- I Can Do This ALL DAY!: a Reaction costs 1 action per the Core Rules; the cost was `none`. A trailing empty paragraph is removed.
- Commanding Presence: the card carried a psychic damage node that only existed to show the DC, so it rolled damage the feature never deals. The node is removed; the top-level Combat Tactics note the other four tactics carry stays.
- Migration055CommanderChecklist (LATEST_SCHEMA_VERSION 55): matches embedded and world copies by compendium source id, then class plus name; every write happens only while the stored value still equals the old pack value; GM-added nodes and rules survive, and a GM-added heal is not doubled; idempotent. 5 pack assertions in `commander.test.ts` cover the pack changes.

Known engine behaviour, not changed here: with no token targeted, a card with a healing node offers to heal the user (`src/documents/item/base.svelte.ts` healing fallback). Breath of Life, Lay on Hands, Heal and Healing Salve already behave this way.

## Test Plan

1. `pnpm check` passes.
2. In a v14 world at schema 54 with a Commander who owns the four items, reload as GM: the migration logs one line per item and the items show the new state.
3. Level 5 Commander, ally at 0 HP targeted: activate Hold the Line!, click Apply Healing on the card. Ally HP is 15, the use pool shows 0/1.
4. STR 3 Commander in combat: roll initiative, the Combat Dice pool shows 3/3. End the combat: 0/3.
5. With Survey the Battlefield: spend a Coordinated Strike use (2/2 to 1/2), start a new combat: back to 2/2 and max Combat Dice STR+1.
6. At level 18: the first Coordinated Strike of an encounter grants the targeted ally an action next turn; the second does not.

Run live on v14.367 through the real UI, 7 of 7 checks, on characters built from the old pack state. Only known core console noise.

## Checklist

## Related Issues

Builds Hold the Line! (#617) and Combat Dice expiry (#612). The live run also confirmed Survey the Battlefield (#619), Coordinated Strike! base, Unparalleled Tactics and Captain of Legions (#613) as shipped. Ticks and the #612 close follow the merge. Part of #685.

- Combat Dice reset to 0 when an encounter ends.
- Hold the Line! heals a targeted ally for `3 * `@level``.
- I Can Do This ALL DAY! costs 1 action.
- Commanding Presence no longer deals psychic damage. Its Combat Tactics note remains.
- Migration 055 updates matching existing character and world item copies.
- GM changes remain preserved.

## Engine changes

- Added `Migration055CommanderChecklist`.
- Set `MigrationRunnerBase.LATEST_SCHEMA_VERSION` to 55.
- Exported migrations 053, 054, and 055.
- Updated Commander pack content and added Commander pack assertions.
- No new settings, rule types, or `en.json` keys were added. No `en.json` keys were removed.


trevlar (changes requested):
**🔧 Push back — Commanding Presence loses the only thing on it that said "Combat Die"**

**Must fix before merge**

- **packs/classFeatures/core/commander/combat-tactics/commanding-presence.json:22** — Dropping the psychic damage node is right; the rulebook gives this tactic no damage. But the `IsAy46e12s3J0Ao1` flavor note went out with it, and that note was never part of the damage node. It sits at the top level with `parentNode: null` and carries the Combat Tactics header text, "1/attack, you can expend a Combat Die to add one of the following effects to your attack." Heavy Strike, Inerrant Strike, Lunging Strike and Sweeping Strike all still ship it. Commanding Presence is now the only one of the five without it, and its description never mentions a Combat Die either, so nothing left on the item tells a player the tactic costs one. Migration037 lifted that note out of its placeholder damage node precisely to keep it. Keep the note and remove only the damage node. If you do mean to drop it, drop it from all five and say why in rules terms.

- **src/migration/migrations/Migration055CommanderChecklist.ts:196** — This gate keeps the flavor note only when a damage node survives, but the note has `parentNode: null` and never hung off the damage node. It's the shared Combat Tactics header text the other four tactics still carry, so whether a GM kept a damage node has no bearing on it. As written, this deletes the note from every Commander in an existing world, and undoing that takes another migration. Keep `#dropPresenceDamage` to the damage node and drop the note branch.

**What's good**

- The `encounterEnd` / `set 0` recovery on the `combat-dice` pool matches the rule, and the new pack assertion pins it against regressions.
- Every write is guarded on the stored value still equaling the old pack value, so the migration really is idempotent and GM edits survive it.

---

1. Keep the Combat Tactics note on Commanding Presence
2. Stop the migration deleting that note

Fronix:
Both requested changes are in (fa87c13d, 1e8ec061). Commanding Presence keeps the Combat Tactics note and drops only the psychic damage node; a pack assertion checks the note matches the one Heavy Strike carries. The migration now removes only the damage node and never touches the note. The PR body is updated to match.

trevlar (approved):
**✅ Approved**

Four Commander pack fixes, plus a migration that brings them to characters who already own the items. Good to merge.

- Every migration write checks the stored value against the old pack data first, so a GM edit stays put and a second run changes nothing. I compared the guards against the dev pack and they match exactly.
- The reaction cost matches the rulebook. Heroic Reactions cost 1 action, and Face Me! and Reposition! already ship that way.

PR #967, opened 2026-09-15
feat(hunter): Forager, Skilled Tracker and Keen Sight roll modes with Migration059

Fronix:
## Summary

The Hunter class branch under the class checklist triage rules: two roll-mode features and one subclass feature that shipped rules can do, one description fix, pack assertions, and Migration059 so existing characters get the same changes. Second branch of the stack, on top of #965 (The Cheat).

## Changes

- Forager (level 1): a player-ticked advantage option on every skill check, labelled "Finding food and water in the wild". Whether the check is about finding food and water is a Table's Call at roll time, the same reading as Gambler's Edge.
- Skilled Tracker (Keeper of the Shadowpath): a player-ticked advantage option on every skill check, labelled "Tracking a creature". The description also gains its missing final period.
- Keen Sight (level 13): a standing skill roll mode that gives every Perception check advantage. No predicate, since the book text has no condition.
- Tests: pack assertions in `hunter.test.ts` pin every rule field the migration copies, the three labels and the fixed description.
- Migration059HunterChecklist (LATEST_SCHEMA_VERSION 59) extends the ChecklistMigrationBase from #965: adds each rule only while no rule with the same id or covering the same clause exists (a skill roll mode also has to match on mode); the description write happens only while the stored text still equals the old pack text; idempotent. Migration tests stay local per repo convention.

Known engine behaviour, not changed here: situational roll options appear only on skill checks that open the roll dialog, so not on the Assess action.

## Test Plan

1. `pnpm check` passes.
2. In a v14 world at schema 58 with a Hunter who owns Forager, Skilled Tracker and Keen Sight in the old state, reload as GM: the migration logs one line per item and the items show the new state; a second reload logs nothing.
3. Roll any skill check on a Hunter: the dialog lists "Finding food and water in the wild" and, with Skilled Tracker, "Tracking a creature". Ticking one rolls with advantage.
4. Roll a Perception check on a Hunter with Keen Sight: the roll is 2d20 keep high with no tick needed. Other skills stay at a flat d20.

Run live on v14.367 through the real UI on characters built from the old pack state. No console errors.

## Checklist

- [x] Added or updated unit tests
- [ ] PR targets the `dev` branch (targets `feat/cheat-checklist`, stacked on #965; retarget to `dev` after that merges)
- [x] `pnpm check` passes (format, lint, circular deps, type-check, tests)
- [x] No hardcoded user-facing strings (used `localize()`)
- [x] Tested in Foundry with no console errors
[...]

Issue #692, opened 2026-04-08
[EPIC] Zephyr automation


Fronix:
All Zephyr class automation — Iron Defense, Burst of Speed maneuvers, Swift Fists, Martial Arts Abilities, and subclass features (Way of Pain, Way of Flame).

Parent: #365

Fronix:
> *This was generated by AI during triage.*

## Stale blocker sweep

4 checklists under this epic cite a closed foundation as a blocker. **A closed foundation does not mean the feature is unblocked.** #943 is the reference for re-reading those lines.

| Checklist | Cites (all closed) | Needs care |
|---|---|---|
| #676 | #576 | #576 |
| #677 | #574, #576, #579, #581, #582, #583, #584 | #576, #581, #582, #584 |
| #678 | #576, #579 | #576 |
| #679 | #574, #575, #576 | #576 |

"Needs care" flags the citations that shipped under a different name or a narrower scope than the checklist assumes:

- **#576 `resourceCounter`** never existed under that name. It is `chargePool` + `chargeConsumer`. Usually genuinely unblocked.
- **#581 `grantMovement`** grants a persistent movement **mode**, not a one-off "move N spaces for free". Free-move features that cite it are still blocked, now by #935.
- **#582** was rescoped before closing. Raw action grants and cost overrides are served; self-directed activation grants and free Move are not.
- **#584 `triggerOnEvent`** was never built. Each rule type has its **own** trigger vocabulary and they differ. #943 carries the table; this citation cannot be resolved generically.

### Class specifics

- **Zephyr is the most wound-dependent class in the game.** Way of Flame is built almost entirely on wound conversion (Exploding Soul, Chain Reaction), and **Unyielding Resolve** is the single most load-bearing quote for **#934**: "Ignore the first Wound you would suffer each encounter (when Wounded abilities, such as Kinetic Momentum, still trigger)." A negated wound must still fire the observation trigger.
- **Mighty Soul** needs both **#934** and **#936**: "gain a Wound to add your STR to the result you rolled. You may repeat this any number of times." It is the only stacking retroactive modifier in the books.
- **Windstep**, **Blur** and **Windstrider** are **#935**.
- **Focus** is **#938**.

Berserker's three checklists (#600, #601, #602) have already been corrected in place and can be used as the worked example.

Issue #677, opened 2026-04-08
[Automation] Zephyr - Martial Arts Abilities checklist (11 picks)


Fronix:
## Context
Checklist for all 11 Martial Arts Abilities selectable via Martial Master at L4/6/8/10/12/14/16/18. Each is an independent feature grant with its own mechanical profile.

Pack directory: `packs/classFeatures/core/zephyr/martial-arts-ability/`

## Abilities

- [ ] **Airshift** - "You cannot be Grappled while conscious. While moving, you may travel across all terrain as normal ground, ignoring all ill effects (e.g., walls/ceilings, water, treetops, lava, spikes, clouds)." `conditionImmunity` (Grappled) + `grantMovement` (all-terrain). Blocked by #581, #585.
- [ ] **Mighty Soul** - "You cannot be moved against your will. Whenever you would fail a saving throw, you may gain a Wound in order to add your STR to the result you rolled. You may repeat this any number of times." Forced-movement immunity + post-roll save modifier with Wound cost. `triggerOnEvent` (onSaveFail). Blocked by #584.
- [ ] **Blur** - "(1/encounter) When you Defend, you may first move up to half your speed away, taking no damage if you are now out of range or have Full Cover." `resourceCounter` (1/enc) + `triggerOnEvent` (onDefend) + `grantAction` (pre-Defend move) + `modifyIncomingAttack` (cancel). Blocked by #576, #582, #583, #584.
- [ ] **Bodily Discipline** - "You may spend 1 action to end any non-Wound condition on yourself." `grantAction` + condition-removal. Blocked by #582.
- [ ] **Quickstrike** - "When you Interpose, you may first make an unarmed strike against the enemy for free." `triggerOnEvent` (onInterpose) + `grantAction` (free unarmed strike). Blocked by #582, #584.
- [ ] **Enduring Soul** - "Each time you roll Initiative, gain Hit Dice equal to the actions you get on your first turn. These Hit Dice expire at the end of combat if unused." `triggerOnEvent` (onInitiative) + temporary HD grant. Blocked by #584.
- [ ] **Use Momentum** - "Whenever you avoid all of the damage of a melee attack (whether it misses or you Defend), you may swap places with the attacker and then choose another target that is now within the attack's reach, and they are hit instead." `triggerOnEvent` (onAttackAvoided) + `modifyIncomingAttack` (redirect). Blocked by #583, #584.
[...]

Issue #676, opened 2026-04-08
[Automation] Zephyr - Progression passives checklist


Fronix:
## Context
Checklist grouping the remaining Zephyr progression features (levels 1-20) that don't warrant dedicated tracking issues. Each entry includes the rulebook quote and the rule type(s) needed.

**Already filed or wired, NOT in this checklist:**
- Burst of Speed (L2) - this batch
- Swift Fists (L1) - this batch
- Unyielding Resolve (L4/10/17) - this batch
- Infuse Strength (L6) - this batch
- #374 Zephyr's Iron Defense Missing Rules - existing tracking issue for the L1 Iron Defense novel AC formula (and its L13 doubling). Do NOT re-file. Warning: coordinate with #374 for the L13 scaling step.
- **Swift Feet** (L2) - speedBonus + initiativeBonus wired. Done. **#398 is now obsolete and can be closed.**
- **Reverberating Strikes** (L5) - meleeDamageBonus wired. Done.

## Features

- [x] **Iron Defense** (L1) - Warning: see #374 (existing). Novel AC formula DEX+STR while unarmored.
- [ ] **Iron Defense (2)** (L13) - "Your armor is doubled while unarmored." Conditional AC doubling while unarmored. Coordinate with #374.
- [x] **Swift Feet** (L2) - "While unarmored, gain +2 speed and +LVL Initiative." `speedBonus` + `initiativeBonus`. Wired. **#398 obsolete, can be closed.**
- [x] **Swift Feet (2)** (L9) - "Gain an additional +2 speed as long as you are unarmored." Additional `speedBonus` step, conditional on unarmored.
- [ ] **Kinetic Momentum** (L3) - "Whenever you gain a Wound, gain a Burst of Speed." `triggerOnEvent` (onWound) + `resourceCounter` (Burst of Speed pool). Blocked by #576, #584.
- [ ] **Ethereal Projection** (L3) - "(1/day) By meditating for at least 10 minutes, you can project an ethereal version of yourself up to 30 ft. away..." Narrative ability with a 1/day counter. `resourceCounter` (1/day) + `note`. Blocked by #576.
- [ ] **Martial Master** (L4/6/8/10/12/14/16/18) - Choose a Martial Arts Ability at each tier. Sheet-level choice mechanic; each pick grants a feature from the Martial Arts Abilities list.
[...]

PR #969, opened 2026-09-15
feat(zephyr): Swift Fists damage, Quickstrike, Ethereal Projection and Martial Master text with Migration060

Fronix:
## Summary

The Zephyr class branch under the class checklist triage rules: two features and one Martial Arts Ability that shipped rules can do, two description fixes, pack assertions, and Migration060 so existing characters get the same changes. Third and last branch of the stack, on top of #967 (Hunter).

## Changes

- Swift Fists (level 1): ships an `unarmedDamage` rule of 1d4 plus STR, and the description gains the damage clause the book has. PR #447 removed this rule when unarmed proficiency moved to the class and left the engine default at 1d4 plus STR for everyone. The core rule for unarmed strikes is 1 plus STR, so the default is wrong for every other class; the rule on Swift Fists carries the Zephyr case so that default can be fixed later without touching Zephyr. The Rushed Attacks exemption stays text (no shipped rule models it) and its checklist line stays open.
- Quickstrike (Martial Arts Ability): a player-clicked action grant of 1 now, gated by a hidden pool that refreshes when the Zephyr's turn ends, the same lockout the book gives a reaction, with a reminder on the card that the action buys the free unarmed strike against that enemy. This is a Stand-in: the system cannot tie the grant to the Interpose click or restrict what the action is spent on.
- Ethereal Projection (level 3): a one-use pool that refreshes on a Safe Rest, spent by the activation. Per Day is once per Safe Rest in this project.
- Martial Arts Ability: the empty description gets the book text for the picks at levels 4 through 18. Existing characters past those levels add their picks by hand.
- Tests: pack assertions in `zephyr.test.ts` pin every rule field the migration copies, the labels, the note and both descriptions.
- Migration060ZephyrChecklist (LATEST_SCHEMA_VERSION 60) extends the ChecklistMigrationBase from #965. The base now treats any existing `unarmedDamage` rule as covering the one Swift Fists ships, whatever formula it holds, because the rule is a single setter and the last one to run would win; so a copy that still carries the Migration010 rule, or a GM's own formula, gets no second one, and it folds the pre-rename `class-features` pack name onto the current one, so copies made before that rename match too. Every write happens only while the stored value still equals the old pack value; GM edits and added rules survive; idempotent. Migration tests stay local per repo convention.

Known engine behaviour, not changed here: activating Quickstrike outside a started combat spends the use but grants nothing, the same as Cheat!. A local draft records the unarmed damage default bug for a separate issue.

## Test Plan

1. `pnpm check` passes.
2. In a v14 world at schema 59 with a Zephyr who owns Swift Fists, Quickstrike, Ethereal Projection and Martial Arts Ability in the old state, reload as GM: the migration logs one line per item and the items show the new state; a GM rule on Ethereal Projection is kept in front; a second reload logs nothing.
3. In combat on the Zephyr's turn with 3 actions, activate Quickstrike: actions 4, pool 0/1, card shows the reminder. A second activation is refused. The pool is 1/1 again once the Zephyr's turn ends, and stays 1/1 through the next turn. End combat: 1/1.
4. Activate Ethereal Projection: pool 0/1, second activation refused. A Field Rest leaves it at 0/1. A Safe Rest brings it to 1/1.
5. The Actions tab lists Unarmed Strike with the 1d4 plus STR damage.

Run live on v14.367 through the real UI on characters built from the old pack state. Only the intended no-charges notices in the console.

## Checklist

- [ ] PR targets the `dev` branch (targets `feat/hunter-checklist`, stacked on #967; retarget to `dev` after that merges)

## Related Issues

Builds Quickstrike (#677), Ethereal Projection and the Martial Master text (#676) and the Swift Fists damage part of #673. Ticks follow the merge. Part of #692.

Players gain corrected Zephyr rules:

- Swift Fists adds `1d4 + STR` unarmed damage.
- Quickstrike grants one action once per round.
- Ethereal Projection gains one use per Safe Rest.
- Martial Arts Ability lists its level-based choices.

Engine changes:

- Added and registered `Migration060ZephyrChecklist`.
- Increased the latest schema version from 59 to 60.
- Updated Zephyr pack content with `unarmedDamage`, `chargePool`, `chargeConsumer`, and `actionDelta` rules.
- When existing characters lack these rules, Migration 060 adds them without duplicates.
- The migration preserves GM edits and is idempotent.
- Added Zephyr pack assertions and migration coverage.
- No settings changed.
- No new rule types were introduced.
- No `en.json` keys were added or removed.

Reported validation passes with `pnpm check` and Foundry v14.367 testing.


src/migration/migrations/Migration060ZephyrChecklist.ts
```diff
@@ -0,0 +1,169 @@
...
+		},
+		{
+			sourceId: `${FEATURE_PREFIX}hzPmX5UlMRWJukc3`,
+			name: 'martial arts ability',
+			description: { from: '', to: MARTIAL_ARTS_ABILITY_DESCRIPTION },
+		},
+	];
+}
+
+export { Migration060ZephyrChecklist };
```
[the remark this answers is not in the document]
Fronix: We don't check in migration tests

src/migration/ChecklistMigrationBase.ts
```diff
@@ -58,6 +59,8 @@ function coversSameClause(existing: RuleSource, wanted: RuleSource): boolean {
 			return existing.poolIdentifier === wanted.poolIdentifier;
 		case 'actionDelta':
 			return existing.target === wanted.target;
```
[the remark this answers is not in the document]
Fronix: Kept by design, the same answer as on #965. The match runs inside one feature item, so a self action delta on the Quickstrike copy is that feature's own automation whatever its values, and a same-id rule a GM has edited is theirs to keep. The migration only fills a copy that lacks the rule; it never rewrites one that is there.

src/migration/ChecklistMigrationBase.ts
```diff
@@ -58,6 +59,8 @@ function coversSameClause(existing: RuleSource, wanted: RuleSource): boolean {
...
+		case 'unarmedDamage':
+			return existing.value === wanted.value;
```
trevlar: Keying `unarmedDamage` on the formula means only an identical formula counts as covering the clause. A GM who set Swift Fists to `1d6 + @abilities.strength.mod` gets a second `unarmedDamage` rule appended, and since the rule's `afterPrepareData` does a plain `setProperty(actor.system, 'unarmedDamage', ...)`, whichever rule runs last wins. That's array order, not the GM's intent, and it's the opposite of what the class doc promises a few lines up.

Unlike a pool or a consumer, `unarmedDamage` is a single global setter: any rule of that type on the item already covers it, whatever formula it holds. `return true;` is the right answer here (the type match already happened at the top of the function), and it still dedupes the Migration010 leftover you're targeting.

src/utils/classProgression/zephyr.test.ts
```diff
@@ -206,3 +207,125 @@ describe('Zephyr selection pools per level', () => {
...
+			const [pool] = pools;
+			expect(pool.disabled).toBe(false);
+			expect(pool.priority).toBe(1);
+			expect(pool.identifier).toBe('quickstrike-round');
+			expect(pool.label).toBe('Quickstrike (once per round)');
+			expect(pool.scope).toBe('item');
+			expect(pool.max).toBe('1');
+			expect(pool.initial).toBe('max');
+			expect(pool.dieSize).toBeNull();
+			expect(pool.hidden).toBe(true);
```
trevlar, line 259: The migration copies `showAsResource` onto existing Quickstrike items, but this block never asserts it, so the pack and `QUICKSTRIKE_RULES` could drift apart on that field with the suite still green. Ethereal Projection's pool test pins it; add the same line here.

trevlar (changes requested):
**🔧 Push back — one line in the shared base can clobber a GM's own unarmed damage rule**

**Must fix before merge**
- **[src/migration/ChecklistMigrationBase.ts:63]** — Keying `unarmedDamage` on the formula means only an identical formula counts as covering the clause. A GM who set Swift Fists to `1d6 + @abilities.strength.mod` gets a second `unarmedDamage` rule appended, and since the rule's `afterPrepareData` does a plain `setProperty(actor.system, 'unarmedDamage', …)`, whichever rule runs last wins. That's array order, not the GM's intent, and it's the opposite of what the class doc promises a few lines up. Unlike a pool or a consumer, `unarmedDamage` is a single global setter: any rule of that type on the item already covers it, whatever formula it holds. `return true;` is the right answer here (the type match already happened at the top of the function), and it still dedupes the Migration010 leftover you're targeting.

- **[src/utils/classProgression/zephyr.test.ts:259]** — The migration copies `showAsResource` onto existing Quickstrike items, but this block never asserts it, so the pack and `QUICKSTRIKE_RULES` could drift apart on that field with the suite still green. Ethereal Projection's pool test pins it; add the same line here.

**What's good**
- Quickstrike reuses the Cheat! shape verbatim: hidden gate pool, fixed consumer, `actionDelta` of 1, reminder note. The one stand-in in this PR is a stand-in the codebase already committed to.
- Swift Fists at 1d4+STR and the Martial Arts Ability level list match the book line for line, levels 6 through 18 included.

---

1. Any `unarmedDamage` rule should count as covering
2. Pin `showAsResource` on Quickstrike's pool test

src/migration/ChecklistMigrationBase.ts
Fronix: Fixed, thanks. The clause now returns true for any existing unarmedDamage rule on the item, whatever formula it holds. A GM's own formula stays alone, and the Migration010 leftover still dedupes.

src/utils/classProgression/zephyr.test.ts
Fronix: Added. The Quickstrike pool test now pins showAsResource the same way the Ethereal Projection test does.

Issue #23, opened 2025-06-28
Concentration not applying on characters when you use spells that have the concentration property


Guipmaru:
I don't know if it's something that could be easily implemented, that everytime a character uses a spell that has the concentration property, it auto applies it to them:

![Image](https://github.com/user-attachments/assets/95784c46-7dba-4385-9864-59d0306b0bfa)

trevlar:
> *This was generated by AI during triage.*

## Triage Notes

**What we've established so far:**

- The request is confirmed. Every piece exists except the wiring: `concentration` is a fully registered condition (id, label, description, icon), it is a selectable spell property on `SpellDataModel`, `SpellCardDataModel` carries a `concentration` boolean, and the spells tab, cast panel, character creator, level-up helper and PDF export all display it. Nothing applies the condition when a spell is cast.
- A GM cannot do this manually today either. `applyCondition`'s triggers are `onHit`/`onCrit`/`onMiss` plus `onTurnStart`/`onTurnEnd`/`onKill`/`onWound`/`onSaveFail`/`onRest`/`onInitiative` — there is no activation trigger. The rule implements `onItemUsed`, which is damage-gated, so a utility spell such as Fly would never fire it.
- The reliable event already exists and is unexposed. `onItemActivated` fires once when an item is used regardless of damage, and `markTarget`, `toggleEffect` and `diceConsumer` all consume it. `applyCondition` simply does not offer it as a trigger.
- **Spell data audit: property flags are correct, four durations are not.** 86 spell items ship across `packs/spells` (76) and `packs/secretSpells` (10). Exactly 12 carry the `concentration` property, and that matches the rulebook's 12 with no spell missing the flag and none carrying it spuriously. But four have the wrong duration — **Fly, Greater Windform, Lesser Windform and Radiant Bond are all "up to 10 minutes" in the rulebook and all store 1 minute.** The other eight are correct, Memory Veil included (`special`, "As long as the caster remains conscious"). This is a data defect independent of the automation, and it matters here because a concentration condition carrying a duration would inherit the wrong value.
- The request covers roughly a quarter of the rules mechanic. Per the Nimble rules: a character can concentrate on only **one** activity at a time; being **crit** while concentrating forces a DC 10 STR save, and failing breaks it; and it breaks automatically at 0 HP or when incapacitated. Only "apply it on cast" is asked for here.
[...]

Issue #950, opened 2026-09-07
Concentration is never broken: crit save, 0 HP, and incapacitated


trevlar:
> *This was generated by AI during triage.*

## Problem

Concentration can be applied but never broken. Split out of #23, which covers applying the condition to the caster, showing it on the cast card, holding only one at a time, and scroll support. The breaking half is deferred here because it needs a primitive the system does not have.

Per the Nimble rules (Concentration):

> A character can only concentrate on one activity at a time. Whenever a character is **crit** while concentrating, they must make a **DC 10 STR save**. Failing this means Concentration is broken and the activity fails. Concentration is automatically broken whenever a character drops to 0 HP or is incapacitated.

The one-at-a-time clause is handled in #23, since it falls out of applying the condition. The other two do not:

1. **Crit while concentrating → DC 10 STR save, failure breaks it.** The event channel exists: `onAttackReceived` carries `isCritical` and does fire on crits. What is missing is any way for a rule to demand a save in response — tracked as #949. Rules can already react to a save's outcome (`onSaveResolved`, and `applyCondition`'s `onSaveFail` trigger), so only the initiating half is blocked.
2. **Automatically broken at 0 HP or when incapacitated.** No save, no roll — an unconditional removal when the caster's state changes. `getActorHealthState` and the dying handling already exist, and `incapacitated` is a registered condition that aliases `paralyzed`, `stunned` and `unconscious`, so "or is incapacitated" has to account for those aliases rather than matching one id.

## Desired outcome

A concentrating character who is crit is asked for a DC 10 STR save, and losing it removes their `concentration` condition. A concentrating character who drops to 0 HP or becomes incapacitated loses it outright, with no save.

## Acceptance cases

- [ ] Being crit while concentrating prompts a DC 10 STR save to the affected player, not only the GM
- [ ] Failing that save removes the `concentration` condition
- [ ] Passing it leaves the condition in place
- [ ] Being crit while **not** concentrating prompts nothing
- [ ] A non-critical hit while concentrating prompts nothing
- [ ] Dropping to 0 HP removes `concentration` with no save offered
- [ ] Becoming incapacitated removes `concentration` with no save offered, including via the aliased conditions (`paralyzed`, `stunned`, `unconscious`)
- [ ] Losing concentration is announced, so the table can see the activity ended
[...]

Issue #949, opened 2026-09-07
[Foundation] Rule-initiated saving throws: save in response to an event


trevlar:
> *This was generated by AI during triage.*

## Purpose

Foundational support for a **rule demanding a saving throw in response to an event**. Today a save can be asked for in exactly two ways, and neither of them is reactive.

**What already exists and is not what this asks for:**

- **Saves can be authored onto an activation.** `savingThrow` is an effect-tree node type, so "DC 18 DEX save, half on save" works when it is part of the thing being activated. The Great Worm's Crush is served fine this way.
- **Saves can be rolled from text.** `savingThrow` is a registered enricher, so a description can offer a clickable save.
- **Rules can react to a save's outcome.** A `saveResolved` hook dispatches `onSaveResolved` to the actor's rules, and `applyCondition` exposes an `onSaveFail` trigger.
- **Rules can modify a save.** `savingThrowBonus` and `savingThrowRollMode` both ship.

**What is missing** is the step that starts one: no rule can say "when X happens to you, make a DC N save, and here is what changes on a failure." Every save in the system today originates from a human activating an item or clicking an enricher. A save that must fire because a creature was crit, took damage, or entered a state cannot be expressed at all.

## Why it matters

The immediate consumer is concentration breaking (tracked separately, and split out of #23):

> Whenever a character is **crit** while concentrating, they must make a **DC 10 STR save**. Failing this means Concentration is broken and the activity fails.

Nothing about that originates with an activation. The trigger is "you were crit while holding a condition", which is the defender's side of someone else's attack. The event channel for it already exists — `onAttackReceived` carries `isCritical` and does fire on crits — but there is no way to turn that into a save.

This is the reactive-save shape generally, so it is worth building once rather than per feature. Anything worded "when this happens to you, save or lose it" needs it.

## Desired outcome

A rule can initiate a saving throw in response to a dispatched event: naming the save type and DC, targeting the actor the event concerns, surfacing it so the owning player rolls it, and letting the outcome drive a consequence — including removing a condition.

## Acceptance cases

- [ ] A rule can declare a save to be prompted on a supported event, with a save type and a DC
- [ ] The prompted save reaches the affected actor's owner, not only the GM
- [ ] The save's result dispatches through the existing `saveResolved` path, so `onSaveFail` and `onSaveResolved` consumers work unchanged
- [ ] A failed prompted save can remove a condition from the actor
[...]

PR #975, opened 2026-09-17
fix(spells): apply concentration to the caster on cast (#23)

trevlar:
Closes #23

Casting a spell whose **Concentration** property is ticked now applies the `concentration` condition to the caster, automatically, with no GM click.

## What changed

**Application on cast.** New `src/utils/applyCasterConcentration.ts` reads the `concentration` property off the activated item's tags, ends the caster's existing concentration, then applies a fresh one through `applyConditionToActor` crediting the new item. It is called from `NimbleBaseItem._createActivationCard`, the shared tail of every `activate()`, so it covers a spell, a scroll (which reaches it through the base path), a roll-less utility spell such as Fly, and a spell that buffs somebody else. The condition always lands on `item.actor`, never on the card's targets.

Derived from the **property**, not from per-spell data, so a homebrew spell or scroll that ticks the box works with no authoring.

**One at a time.** `applyConditionToActor` refuses a condition the actor already carries, so the prior instance is removed before the new one is applied. A caster who casts a second concentration spell ends with exactly one, credited to the new spell.

**Card indicator.** New `ConcentrationNode.svelte`, rendered by both `SpellCard` and `ObjectCard`, backed by a shared `concentration()` schema in `src/models/chat/common.ts` and stamped in `_createActivationCard`, onto the card types whose schema declares the field. This replaces the spell card's unread `duration` schema, whose `period` field was assigning a boolean into a `StringField`; nothing read either field.

**Pack data.** Removed the now-redundant `condition: concentration` node from the eight core spells that shipped one (Boisterous Winds, Electrocharge, Enchant Weapon, Fiery Embrace, Fly, Memory Veil, Seething Storm, Shadow Trap). Corrected Fly, Greater Windform, Lesser Windform and Radiant Bond from 1 minute to 10, verified against `CoreRules-2.0.3` ("Fly … Concentration: Up to 10 minutes") and the GMG secret-spell list.

**Migration058** clears those nodes from spells and objects in existing worlds, matched on the `concentration` property rather than on source ids so homebrew copies and inscribed scrolls are covered. `LATEST_SCHEMA_VERSION` bumped to 58.

## Acceptance criteria

- [x] A player casting any of the twelve concentration spells gets `concentration` on their own character, no GM involvement
- [x] Holds for a spell targeting an ally (Electrocharge): the condition lands on the caster
- [x] Holds for a spell cast with no target selected (Fly)
- [x] A non-concentration spell applies nothing
- [x] The cast card visibly indicates concentration, and the casting player can see it
- [x] A non-concentration spell's card shows no indicator
- [x] Casting from a scroll applies it to the caster and shows the same indicator
- [x] A second concentration spell leaves exactly one `concentration` condition
- [x] A homebrew spell with the property ticked behaves identically, no rules authored
- [x] No spell offers a GM-clickable `concentration` chip any more
- [x] A migration clears the removed nodes from existing worlds
- [x] Fly, Greater Windform, Lesser Windform and Radiant Bond store 10 minutes

## Tests

12 new tests: the util's behavior and its one-at-a-time replacement, the `_createActivationCard` wiring, card-data stamping on both item types, and a pack-data test pinning "no concentration-property spell carries a concentration node" plus the four durations.

The wiring tests were confirmed red before the fix (temporarily disabling the call fails 2 of 3), and the pack test would fail against `origin/dev`'s data.

## Gates

`pnpm check` exits 0: Biome format, lint, dependency-cruiser, `tsc --noEmit`, and 4594 tests across 296 files. The migration-count check passes (58 files / version 58). `pnpm docs:generate` produced no reference changes.

The one remaining Biome `info` is a pre-existing `useLiteralKeys` note in `shouldSuppressActivationCard.test.ts`, a file this branch does not touch.

## Out of scope

Breaking concentration (crit save, 0 HP, Incapacitated) stays with #950 / #949. The applied condition carries no duration: nothing asked for it, and with no auto-break there is nothing to expire against. The docs state plainly that ending concentration is still manual.

## Worth a reviewer's eye

A weapon-type object with **Concentration** ticked will now apply the condition on use, since objects and spells share the property and the brief called for property-derived behavior covering both data models. That reads as correct to me, but no acceptance criterion names it.

## Docs

New Concentration section in `docs/documentation/playing/conditions.md`, a note in `homebrew/spells.md`, and the new utility added to the Shared Code Inventory.

- When a spell or object has the `Concentration` property, activation now applies `concentration` to the caster.
- The existing `concentration` condition ends before the new condition applies.
- The rule applies to spells, scrolls, roll-less spells, ally-targeting spells, and homebrew items.
- Spell and object chat cards show **Concentrating** only when condition application succeeds.
- Failed or refused condition application does not show a false concentration result.
- Core pack data no longer adds redundant `concentration` effects.
- `Fly`, `Greater Windform`, `Lesser Windform`, and `Radiant Bond` now use a 10-minute duration.
- Homebrew documentation explains automatic concentration and manual concentration loss tracking.

## Engine changes

- Added `applyCasterConcentration()` to the shared activation path.
- Activation continues when concentration application fails.
- Existing activation hooks and area-of-effect placement remain active after concentration failure.
- Added concentration data to spell and object chat card schemas.
- Added `ConcentrationNode.svelte` to both card types.
- Removed spell card duration concentration fields.
- Updated `ConditionSourceDocument.uuid` to accept `null`.
- Added migration `058`, `Migration058ConcentrationAppliesToCaster`.
- Increased the latest schema version from `57` to `58`.
- Added tests for application, replacement, refusal, errors, caster immunity, failed scroll checks, card output, and pack data.
- Changed pack content by removing redundant concentration effects and correcting four spell durations.
- Added the `NIMBLE.chat.concentrationApplied` key to `public/lang/en.json`.
- No new settings or rule types were added.


src/documents/item/object.test.ts
```diff
@@ -192,6 +200,17 @@ describe('NimbleObjectItem.activate', () => {
 			expect(scroll.delete).toHaveBeenCalled();
 		});
 
+		it('starts no concentration on a failing roll', async () => {
```
Fronix, line 203: This test cannot fail. The `beforeEach` at line 105 mocks `NimbleBaseItem.prototype.activate`, and concentration is only applied inside that method. `createConditionEffect` is therefore never called, on a pass or on a fail. Assert `expect(baseActivate).not.toHaveBeenCalled()` instead, or remove the test.

src/view/chat/components/ConcentrationNode.svelte
```diff
@@ -0,0 +1,39 @@
...
+			border-bottom: 1px solid var(--nimble-card-border-color);
+		}
+	}
+
+	.nimble-concentration-indicator {
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		gap: 0.5rem;
+		background: var(--nimble-color-primary-alpha);
```
Fronix, line 34: `--nimble-color-primary-alpha` and `--nimble-color-primary` are not defined anywhere in the repo, and there is no fallback, so the banner gets no background and no colour. This block copies the upcast indicator (`SpellCard.svelte:144-145`), which has the same fault. Use colour tokens that are defined, or add fallbacks as the sibling nodes do.

src/utils/applyCasterConcentration.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export default async function applyCasterConcentration(
+	item: ConcentrationItem,
+): Promise<AppliedConditionEffect | null> {
+	if (!requiresConcentration(item)) return null;
+
+	const caster = item.actor;
+	if (!caster) return null;
+
+	await caster.toggleStatusEffect(CONCENTRATION_CONDITION_ID, { active: false });
```
Fronix, line 41: The old concentration is deleted before `applyConditionToActor` runs the `preApplyCondition` check and the create. If the check refuses, or the create fails, the caster ends with no concentration, not one. #23 says "ends with exactly one". Remove the old instance only after the new apply is allowed.

Separately, this line ends every earlier concentration. Master of Storm (Stormshifter, level 15) allows 1 lightning spell and 1 wind spell at the same time, so this feature needs an exception or a documented limit.

src/utils/applyCasterConcentration.ts
```diff
@@ -0,0 +1,47 @@
+import applyConditionToActor, {
+	type AppliedConditionEffect,
+	type ConditionTargetActor,
+} from './applyConditionToActor.js';
+
+const CONCENTRATION_CONDITION_ID = 'concentration';
```
Fronix, line 6: `STATUS_EFFECT_IDS.concentration` already exists in `src/config/registerConditionsConfig.ts:18`. Import that constant instead of this literal.

src/migration/migrations/Migration058ConcentrationAppliesToCaster.ts
```diff
@@ -0,0 +1,36 @@
...
+	override readonly version = Migration058ConcentrationAppliesToCaster.version;
+
+	override async updateItem(source: any): Promise<void> {
+		if (source.type !== 'spell' && source.type !== 'object') return;
+		if (!source.system?.properties?.selected?.includes(CONCENTRATION)) return;
+
+		const effects = source.system?.activation?.effects;
+		if (!Array.isArray(effects)) return;
+
+		const remaining = effects.filter(
```
Fronix, line 25: This filter only looks at top-level nodes. Activation nodes can have `children` (see `flattenActivationEffects`), so a homebrew concentration node nested under another node stays in place and applies concentration a second time. The shipped data has no nested nodes, so this matters only for homebrew.

src/migration/migrations/Migration058ConcentrationAppliesToCaster.ts
```diff
@@ -0,0 +1,36 @@
+import { MigrationBase } from '../MigrationBase.js';
+
+const CONCENTRATION = 'concentration';
+
+/**
+ * Drop the `condition: concentration` activation nodes the core spells shipped
+ * with, now that the property applies the condition on its own (issue #23).
```
Fronix, line 7: Nit: please remove "(issue #23)" from the comment.

src/documents/item/base.svelte.ts
```diff
@@ -178,6 +179,9 @@ class NimbleBaseItem<ItemType extends SystemItemTypes = SystemItemTypes> extends
 	/**
 	 * Create the activation chat card unless a rule suppresses it, then fire
 	 * the `useItem` hook. Shared tail of every activate() implementation.
+	 *
+	 * Concentration is applied here, not on the damage-applied path, which a
+	 * roll-less utility spell such as Fly never reaches.
```
Fronix, line 184: Nit: engine comments must stay generic. Write "a roll-less activation", not the name of a specific spell.

src/documents/item/base.svelte.ts
```diff
@@ -186,6 +190,15 @@ class NimbleBaseItem<ItemType extends SystemItemTypes = SystemItemTypes> extends
...
 	): Promise<ChatMessage | null> {
 		const suppressCard = this._shouldSuppressActivationCard(rolls, activation);
+
+		let concentrating = false;
+		try {
+			concentrating = (await applyCasterConcentration(this)) !== null;
+		} catch (error) {
+			console.error('Nimble | Could not apply concentration to the caster.', error);
+		}
+		foundry.utils.setProperty(chatData as object, 'system.concentration', concentrating);
```
Fronix, line 200: This line writes `system.concentration` into every card type, but only the spell and object card schemas have the field. On the other card types, Foundry removes the key without a message. Consider writing it only for the card types that have the field.

public/lang/en.json
```diff
@@ -94,6 +94,7 @@
 			"applyConditionToTargets": "Apply condition to this card's targets",
 			"applyDamage": "Apply Damage",
 			"applyHealing": "Apply Healing",
+			"concentrationApplied": "Concentrating",
```
Fronix, line 97: The key is `concentrationApplied`, but the text is "Concentrating". The chat card stays after concentration ends, so it then shows something that is no longer true. "Concentration applied" stays correct.

Fronix (changes requested):
## Review: changes requested

The main behaviour works. Every spell and scroll cast reaches the new call, the condition goes on the caster, and the card indicator is visible to the player. I request changes for one rules problem, one test that cannot fail, and a small number of standards problems. The inline comments give the details.

### Rules
- **Master of Storm (Stormshifter, level 15)** says: "You can concentrate on 1 lightning spell and 1 wind spell at the same time." (Heroes 2.0.2). `applyCasterConcentration` always ends the earlier concentration, so a Stormshifter who casts Fly while concentrating on Electrocharge loses the first one. Add an exception for this feature, or state the limit in the docs.
- The rulebook says "A character can only concentrate on one activity at a time", but it does not say that a new concentration ends the old one. The replacement is an inference. It is reasonable, but it is not in the text.
- Confirmed against the rulebooks: the condition goes on the caster, the set of 12 spells is correct, the four 10 minute durations are correct, and the docs text about breaking concentration is correct.

### Must fix
1. `object.test.ts:203`: this test cannot fail (see the inline comment).
2. `ConcentrationNode.svelte:34`: the colour tokens are not defined (see the inline comment).
3. `applyCasterConcentration.ts:41`: the old concentration is removed before the new apply is allowed (see the inline comment).

### Standards
- `applyCasterConcentration.ts` is a global utility with one caller (`_createActivationCard`). The AGENTS.md "Code Promotion Rules" keep a function with one caller in the file that calls it.
- `applyCasterConcentration.ts:6` repeats a literal that already has a constant, `STATUS_EFFECT_IDS.concentration` (`src/config/registerConditionsConfig.ts:18`). STYLE_GUIDE says "No magic numbers or repeated literals".
- `src/utils/concentrationSpellsPack.test.ts` tests pack data, not a module in `src/utils/`. `secretSpellsPack.test.ts` is a precedent, so this is minor. Lines 33 to 36 also throw on a spell that has no `effects` array, and the error does not name the spell.
- `ConcentrationNode.svelte` is the third copy of the `.nimble-card-section` style block. It also looks almost the same as the upcast indicator, so an upcast concentration spell shows two banners that look the same, next to each other.
- The new docs prose in `conditions.md` and `homebrew/spells.md` uses em-dashes. Please rewrite those sentences.

### Spec (#23)
- The four duration fixes change only the pack JSON. Copies of these spells already in worlds or on characters keep 1 minute, because Migration058 only removes nodes. The brief scoped this to pack data, so this is a gap, not a violation. Decide if Migration058 must also update `duration.quantity`.
- The card flag now means "the condition was applied", not "the item has the property". If the apply throws, the card shows no indicator and the only error goes to the console. Please show the user a warning (`ui.notifications.warn` with a `NIMBLE.` key).
- Concentration is applied even when a rule hides the card. The tests assert this, but #23 does not ask for it. Please confirm that this is intended.
- The PR body says the flag is set "in both `prepareChatCardData` implementations". The code sets it in `_createActivationCard`. Please update the body.

docs/documentation/homebrew/spells.md
```diff
@@ -28,7 +28,7 @@ Spell sheets have no Rules tab. Passive bonuses belong on features and objects,
 
 On the **Configuration** tab:
 
-- **Spell Properties**: toggle any of **Concentration**, **Range**, **Reach**, **Secret Spell**, and **Utility**. Selecting Range or Reach reveals the matching distance configuration. Marking a spell as **Utility** hides the tier and scaling sections entirely, because utility spells sit outside the tier system.
+- **Spell Properties**: toggle any of **Concentration**, **Range**, **Reach**, **Secret Spell**, and **Utility**. Selecting Range or Reach reveals the matching distance configuration. Marking a spell as **Utility** hides the tier and scaling sections entirely, because utility spells sit outside the tier system. Ticking **Concentration** is all it takes to make casting the spell apply the [Concentration condition](../playing/conditions.md#concentration) to the caster — don't add a condition effect for it as well.
```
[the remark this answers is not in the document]
trevlar: Fixed in `098c01c0` — the sentence now breaks at a period.

docs/documentation/playing/conditions.md
```diff
@@ -23,6 +23,16 @@ Some conditions come bundled together, and the system tracks the relationships f
 - **Automatic Hampered.** Applying Dazed, Grappled, Prone, Slowed, or Restrained automatically applies Hampered as well. When the last of those trigger conditions is removed, Hampered goes away on its own.
 - **Wounded stacks.** Unlike other conditions, Wounded can be applied multiple times, one stack per wound.
 
+## Concentration
+
+Concentration applies itself. Casting a spell whose **Concentration** property is ticked puts the Concentration condition on the caster as the chat card posts, and the card says **Concentrating** so the player can see it took. No GM click is involved, and it happens whether the spell targets an ally, several creatures, or nobody at all — the condition always lands on the caster, never on the targets.
```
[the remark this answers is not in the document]
trevlar: Fixed in `098c01c0` — the sentence now breaks at a period.

docs/documentation/playing/conditions.md
```diff
@@ -23,6 +23,16 @@ Some conditions come bundled together, and the system tracks the relationships f
...
+
+A caster holds one concentration at a time. Casting a second concentration spell ends the first: the old condition is removed and a new one applied, credited to the new spell.
```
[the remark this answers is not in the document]
trevlar: Fixed in `098c01c0` — the section now says the card shows nothing and the caster gets a warning when the condition cannot be applied.

src/documents/item/activationConcentration.test.ts
```diff
@@ -0,0 +1,143 @@
...
+	});
+
+	it('still fires the useItem hook when applying the condition throws', async () => {
+		const item = createItemStub(['concentration']);
+		item.actor.toggleStatusEffect.mockRejectedValue(new Error('no permission'));
+
+		await createActivationCard(item);
+
+		expect(Hooks.callAll).toHaveBeenCalledWith(
+			'nimble.useItem',
```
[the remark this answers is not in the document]
trevlar: Fixed in `098c01c0` — the assertion uses `systemHookName('useItem')`.

src/documents/item/object.test.ts
trevlar: Fixed in `098c01c0` — removed. `wastes the scroll without casting on a failing roll` directly above already asserts `baseActivate` was not called, which is the real boundary, so a second test on the same fact added nothing.

src/utils/applyCasterConcentration.ts
trevlar: Fixed in `098c01c0` — uses `STATUS_EFFECT_IDS.concentration`.

src/migration/migrations/Migration062ConcentrationAppliesToCaster.ts
```diff
@@ -0,0 +1,36 @@
+import { MigrationBase } from '../MigrationBase.js';
+
+const CONCENTRATION = 'concentration';
+
+/**
+ * Drop the `condition: concentration` activation nodes the core spells shipped
+ * with, now that the property applies the condition on its own (issue #23).
```
trevlar: Fixed in `098c01c0` — dropped.

src/documents/item/base.svelte.ts
trevlar: Fixed in `098c01c0` — reads "a roll-less activation".

src/view/chat/components/ConcentrationNode.svelte
trevlar: Fixed in `098c01c0` — both tokens were undefined repo-wide. Added `--nimble-chat-concentration-background` and `--nimble-chat-concentration-color` in `_variables.scss` and `_theme.scss` alongside the other `--nimble-chat-*` pairs, and pointed the node at them. The new colour is purple, so the banner no longer looks like the upcast indicator sitting next to it. SpellCard.svelte:144-145 still carries the same fault on the upcast indicator; left alone as out of scope here.

src/migration/migrations/Migration062ConcentrationAppliesToCaster.ts
```diff
@@ -0,0 +1,36 @@
...
+	override readonly version = Migration058ConcentrationAppliesToCaster.version;
+
+	override async updateItem(source: any): Promise<void> {
+		if (source.type !== 'spell' && source.type !== 'object') return;
+		if (!source.system?.properties?.selected?.includes(CONCENTRATION)) return;
+
+		const effects = source.system?.activation?.effects;
+		if (!Array.isArray(effects)) return;
+
+		const remaining = effects.filter(
```
trevlar: Fixed in `098c01c0` — the filter is now a recursive walk that prunes a concentration node at any depth.

src/documents/item/base.svelte.ts
trevlar: Fixed in `098c01c0` — the write is gated on the card type. `chatDataModels.test.ts` asserts the gate list matches the card models that actually declare the field, so the two cannot drift.

public/lang/en.json
trevlar: Fixed in `098c01c0` — the string is now "Concentration Applied".

src/utils/applyCasterConcentration.ts
trevlar: Fixed in `098c01c0`, both parts.

Transactional: `applyConditionToActor` takes a `replaces` option naming the effects this application supersedes. They are deleted only after `preApplyCondition` has allowed the new one, and their stored sources are re-created if creation throws. Tests cover hook refusal and creation failure with an existing concentration.

Master of Storm: added a `concentrationTrack` rule type that gives named spell schools a concentration of their own, and put it on `master-of-storm.json` with lightning and wind. The effect carries its track in a system flag, so a cast only ends concentration on its own track. Schools no rule names share the default track, which is the one-at-a-time behaviour every other caster has.

On the rulebook point: confirmed against the vault that the text says only "A character can only concentrate on one activity at a time" and never says a new concentration ends the old one. `conditions.md` now says the replacement is the system's reading rather than printed text.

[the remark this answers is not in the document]
trevlar: Fixed in `098c01c0` — `applyConditionToActor` takes a `replaces` option. The named effects go only after `preApplyCondition` has allowed the replacement, and their stored sources are re-created if creation throws. Tests cover hook refusal and creation failure with an existing concentration.

src/view/chat/components/ConcentrationNode.svelte
```diff
@@ -0,0 +1,39 @@
...
+
+	const isConcentrating = $derived(messageDocument?.reactive?.system?.concentration ?? false);
+</script>
+
+{#if isConcentrating}
+	<section class="nimble-card-section nimble-concentration-indicator">
+		<i class="fa-solid fa-brain"></i>
+
+		{localize('NIMBLE.chat.concentrationApplied')}
+	</section>
```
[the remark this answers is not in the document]
trevlar: Fixed in `098c01c0` — added `ConcentrationNode.test.ts` with a `testHarness.svelte`, matching `ConditionNode`. It renders the component and asserts the localized text is in the document when the value is `true`, and absent when it is `false` or missing.

trevlar:
## Fixed

- **Concentration lost when the replacement was refused** — `applyConditionToActor` takes a `replaces` option naming the effects an application supersedes; they go only after `preApplyCondition` allows the new one, and their stored sources are re-created if creation throws (`src/utils/applyConditionToActor.ts:183`).
- **Master of Storm lost a concentration it should keep** — new `concentrationTrack` rule type gives named spell schools a concentration of their own; the effect carries its track in a system flag, so a cast only ends concentration on its own track (`src/models/rules/concentrationTrack.ts:46`, `src/documents/item/base.svelte.ts:76`). Wired to `master-of-storm.json` with lightning and wind.
- **Global utility with one caller** — `applyCasterConcentration.ts` is gone; the path is module-scope functions in the file that calls it, and the STYLE_GUIDE inventory row went with it (`src/documents/item/base.svelte.ts:94`).
- **Repeated `'concentration'` literal** — uses `STATUS_EFFECT_IDS.concentration`.
- **Test that could not fail** — removed. The test above it already asserts `baseActivate` was not called, which is the real boundary.
- **Undefined colour tokens** — added `--nimble-chat-concentration-background` / `-color` to `_variables.scss` and `_theme.scss` and pointed the node at them. The new purple also stops the banner looking identical to the upcast indicator next to it (`src/view/chat/components/ConcentrationNode.svelte:34`).
- **Silent failure** — a failed apply now shows `ui.notifications.warn` on `NIMBLE.chat.concentrationFailed`, not just a console line (`src/documents/item/base.svelte.ts:285`).
- **Card flag written onto every card type** — gated on the card types whose schema declares the field, with `chatDataModels.test.ts` asserting the gate matches the models so the two cannot drift.
- **Nested concentration nodes survived the migration** — the filter is a recursive walk now.
- **Stale durations on existing copies** — Migration062 corrects `duration.quantity`/`type` for the four ten-minute spells. The table is exported and `concentrationSpellsPack.test.ts` asserts the pack data against the same one.
- **Pack test threw on a spell with no `effects`** — reads through optional chaining and reports by name.
- **Misleading label** — `concentrationApplied` reads "Concentration Applied", which stays true after concentration ends.
- **Em dashes in the new docs prose** — rewritten in `conditions.md` and `homebrew/spells.md`.
- **`(issue #23)` in a code comment, Fly named in an engine comment** — both dropped.
- **No rendered test for the indicator** — `ConcentrationNode.test.ts` with a test harness, matching `ConditionNode`.

## Not fixed

- **The third copy of the `.nimble-card-section` block** — two similar blocks is where AGENTS.md puts the DRY threshold, and extracting a card-section component touches every chat card. Worth its own pass over all of them rather than this PR.
- **A Foundry v14 integration test for condition persistence** — no integration harness exists for activating a real embedded item; that is infrastructure, not a fix.

## Approach

The two "must fix" items turned out to be the same mechanism. Making replacement transactional meant `applyConditionToActor` had to own the swap, and once a caller could name what it replaces it could also name *nothing*, which is exactly what a second concentration in another track needs. So the Master of Storm exception fell out of the same option rather than needing a parallel path.

Master of Storm is implemented as a rule, not a special case: a school named by a `concentrationTrack` rule gets a concentration of its own, everything else shares one. Confirmed against the vault that this is the only such feature in the published material, and that the schools are the spell's school field with no separate keyword. Two things the rulebook does not settle, so the code does not pretend to: it never says a new concentration ends the old one (replacement is our reading, and `conditions.md` now says so), and it never says what a caster with the feature does when they exceed their tracks.

On applying concentration when a rule hides the card: suppression means a rule posts its own output in place of the default card, not that the spell was not cast. The condition should land either way, so this is unchanged and the test stays.

The PR body is corrected: the flag is set in `_createActivationCard`, not in the `prepareChatCardData` implementations.

src/view/chat/components/ConcentrationNode.svelte
trevlar: Follow-up in `ad6678b9` — also fixed the upcast indicator rather than leaving it. It now uses `--nimble-chat-upcast-background` / `-color`, defined in both `_variables.scss` and `_theme.scss`. No `--nimble-color-primary*` reference is left in `src/view/chat/`.

src/view/chat/components/ConcentrationNode.svelte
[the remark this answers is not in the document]
trevlar: Follow-up in `ad6678b9` — added `tests/integration/concentration.test.ts` as well. It activates a real embedded spell and inspects the persisted effect on the caster: origin credited to the spell, the track flag, replacement leaving exactly one, and a `concentrationTrack` rule holding lightning and wind at once. I could not execute it: it needs a running Foundry instance and none is reachable here, and this suite never runs in CI.

trevlar:
## Fixed

- **`.nimble-card-section` copied into nine components** — the base block now lives once in `src/scss/components/_chat-card.scss:163`; components vary the padding through `--nimble-card-section-padding`, which `Targets.svelte` already did. The `--description` `:global` block was a third copy and moved with it. 107 lines of duplication gone, and `ObjectCard.svelte` no longer needs a `<style>` block at all.
- **Upcast indicator had the same undefined tokens** — now on `--nimble-chat-upcast-background` / `-color`, defined in `_variables.scss` and `_theme.scss`. No `--nimble-color-primary*` reference left under `src/view/chat/`.
- **No live coverage of condition persistence** — added `tests/integration/concentration.test.ts`: activates a real embedded spell and inspects the persisted effect (origin credited to the spell, track flag, replacement leaving exactly one, and a `concentrationTrack` rule holding lightning and wind at once).

## Not fixed

- Nothing.

## Approach

I was wrong twice in the last round and corrected both. AGENTS.md says extract *after* the same pattern appears three times across different files; `ConcentrationNode` was the third, so the threshold was met, not missed. And I said no integration harness existed when `tests/integration` and `pnpm test:integration` are right there.

The globalized rule could have fought the components on specificity, since a Svelte-scoped `.nimble-card-section.svelte-hash` and `.system-nimble .nimble-card-section` are both (0,2,0). Avoided that by leaving exactly one `padding` declaration in the codebase and having every component set the variable instead. Verified against the built CSS: one `padding` rule, and each surviving scoped rule sets only the custom property or an unrelated layout modifier.

`LevelUpSummaryCard` and `FieldRestCard` also style `.nimble-card-section` but render a single section each, so the inherited `:not(:last-of-type)` border cannot match them.

**The integration test is unexecuted.** It needs a running Foundry instance, none is reachable here, and that suite never runs in CI. Everything else is verified: lint, circular-deps, typecheck, `vite build`, and 4646 unit tests pass.

Fronix (changes requested):
## Review

All round 1 threads are resolved and I confirmed each fix landed.
The findings below are new, or are places where a round 1 fix is incomplete.

### Blocking

**1. A refused create still loses the held concentration.** `src/utils/applyConditionToActor.ts:264-268`

```ts
try {
	effect = await createConditionEffect(target, conditionId, options);
} catch (error) {
	await restoreReplacedEffects(target, replacedSources);
	throw error;
}
```

`createConditionEffect` ends `return created ?? null;` (`:175`). Foundry's `Document.create`
resolves `undefined` when a `preCreateActiveEffect` hook or `_preCreate` returns false. That
path returns `null` without restoring, and it also fires `conditionApplied` with `effect: null`.
This is the same loss round 1 asked you to close. It is closed for `preApplyCondition` and for
a throw, but not for a refused create. `docs/documentation/playing/conditions.md:32` promises
the opposite: "If the new one is refused, the first is left where it was."

Fix: restore and return when `effect` is `null`, before the hook call.

**2. Two concentration tracks cannot be ended one at a time.** `src/documents/item/base.svelte.ts:65-77`

`concentration` has no static `_id`, because only `petrified` is linked
(`src/config/registerConditionsConfig.ts:163`). V14's `Actor#toggleStatusEffect` therefore
deletes every single-status match in one call (`client/documents/actor.mjs:560-563`):

```js
for ( const effect of this.effects ) {
  const statuses = effect.statuses;
  if ( (statuses.size === 1) && statuses.has(status.id) ) existing.push(effect.id);
}
```

A Master of Storm caster holding lightning and wind loses both on one token HUD click. No other
part of the UI ends a single track. `conditions.md:36` tells that player to remove the condition
by hand.

**3. Master of Storm grants three concurrent concentrations, not two.** `src/documents/item/base.svelte.ts:57-62`

> **_LEVEL 15_ Master of Storm.** You can concentrate on 1 lightning spell and 1 wind spell at
> the same time. (`Heroes-2.0.2.md:1805`)

Naming lightning and wind gives each school its own track and leaves the default track free. The
holder can then hold lightning, wind, and one other school at the same time. A Stormshifter picks
up a third school at level 3 through Deepening Study (`Heroes-2.0.2.md:1789`).

Fix: a named school should claim the slot from the default track, not add one.

**4. The `concentrationTrack` reference partial is missing, so the doc link resolves to nothing.**

```md
Homebrewers get this through the [Concentration Track](../reference/rules-conditions.md#concentration-track) rule.
```

`docs/documentation/playing/conditions.md:34`. Partials are hand written
(`scripts/docs/generateReference.gen.ts:11`), not generated, and all 25 other rule types have one.
`pnpm docs:generate` cannot report a change here, because `docs/documentation/reference/*.md` is
gitignored (`.gitignore:18`).

**5. The hoisted card-section rule reaches a component that never carried it.** `src/scss/components/_chat-card.scss:166`

```scss
.system-nimble .nimble-card-section {
  padding: var(--nimble-card-section-padding, 0.5rem);
  &:not(:last-of-type) { border-bottom: 1px solid var(--nimble-card-border-color); }
}
```

`TemplateSection.svelte:51` renders `class="nimble-card-section nimble-card-section--template"`
and declares neither property. Svelte scoping kept the parent cards' copies off it. It now gains
padding and a border-bottom on the Spell, Object and Feature cards.

Fix: give `--template` an explicit `--nimble-card-section-padding`.

### Should fix

**6. A refusal shows no warning.** `src/documents/item/base.svelte.ts:291-296`. `conditions.md:28`
says the caster gets a warning when the condition cannot be applied. Only a throw warns. A
`preApplyCondition` refusal returns `null` in silence.

**7. Two concentration objects still apply nothing.** `packs/magicItems/core/wands/wand-of-fly.json`
and `packs/magicItems/core/gmGuide/cloak-of-lesser-windform.json` read "Concentration" in their
text, but `properties.selected` is `[]`. Migration062 skips them for the same reason. Wand of Fly
also stores 1 minute against its own "Up to 10 minutes". Pre-existing, but inside this PR's own
duration-correction scope.

**8. The delete loop has no guard.** `src/utils/applyConditionToActor.ts:189-191`

```ts
for (const effect of replaces) {
	const source = effect.toObject?.();
	await effect.delete();
```

A throw here escapes before the `try`. With two or more replaced effects, the earlier deletes are
lost and nothing is created. `deleteDocuments(ids, { parent })` makes it one atomic call.

**9. `replaces: []` turns off duplicate suppression.** `src/utils/applyConditionToActor.ts:245`

```ts
if (!options.replaces && targetAlreadyHasCondition(target, conditionId)) return null;
```

An empty array is truthy. `applyCasterConcentration` always passes an array, so it never gets the
guard. Test `options.replaces?.length`, or state the behaviour in the JSDoc.

**10. Concentration has no automation gate.** `src/documents/item/base.svelte.ts:292`. Every other
auto-applied condition goes through `isDerivedConditionsAutomationEnabled()`
(`src/hooks/automaticConditions.ts:23,90`). A table with derived conditions off still gets the
condition applied, and its previous one deleted.

**11. Inline array in pack JSON.** `packs/.../circle-of-sky-and-storm/master-of-storm.json:18`

```json
"schools": ["lightning", "wind"]
```

This is the only inline multi-entry array in the whole pack tree. `pnpm format` runs biome on
`./src/ ./types/ ./lib/` only, so nothing fixes packs. Expand it, as `deepening-study.json` does.

### Nits

- `src/models/rules/concentrationTrack.test.ts:12`: `disabled` is a dead factory parameter, and
  `predicate` is pinned to `{ size: 0 }`. `test()` checks both (`src/models/rules/base.ts:346-348`),
  and neither branch is exercised negatively. `speedBonus.test.ts:143` has the house pattern.
- `Migration062ConcentrationAppliesToCaster.ts:76`: durations are rewritten by `source.name`. A GM's
  own spell named `Fly` is overwritten, and a translated world matches nothing.
- `SafeRestCard.svelte:54` keeps a scoped `padding: 0.5rem` that now ties on specificity with the
  global rule. Injection order decides it. The values match today, so this is latent only.
- `CONTEXT.md` has no entry for "concentration track", which is a new domain term.

### Confirmed correct

- Concentration is one at a time (`CoreRules-2.0.2.md:352`). The rulebook does not say what happens
  on a second cast, and `conditions.md:32` says plainly that replacement is the system's reading.
- All four corrected durations: Fly (`CoreRules-2.0.2.md:1620`), Lesser Windform
  (`GMguide-2.0.2.md:732`), Greater Windform (`:736`), Radiant Bond (`:738`). The six spells left
  at 1 minute match their own printed text. Memory Veil's `special` matches `GMguide-2.0.2.md:740`.
- Property-driven application is behaviour-preserving: all eleven touched spells already carry
  `concentration` in `properties.selected`.
- Migration numbering: 62 files, `LATEST_SCHEMA_VERSION = 62`, contiguous.

trevlar:
## Fixed

- **A refused create lost the held concentration** — `Document.create` resolving `undefined` is now treated the same as a throw: restore the replaced sources and return `null` before the `conditionApplied` hook fires (`src/utils/applyConditionToActor.ts:286`).
- **Two tracks ended on one HUD click** — `NimbleBaseActor#toggleStatusEffect` intercepts `concentration` and, when two are held, asks which to end instead of letting core delete every single-status match (`src/documents/actor/base.svelte.ts:1171`, `src/documents/actor/endOneConcentration.ts`).
- **Master of Storm granted a third concurrent track** — a named school now claims the default slot rather than adding one: casting in a named school ends that school's track plus anything on the default, and casting in an unnamed school ends everything (`src/documents/concentration.ts:61`).
- **Missing `concentrationTrack` reference partial** — written by hand like the other 25 (`docs/documentation/reference/_partials/concentrationTrack.md`).
- **Hoisted card-section rule reached `TemplateSection`** — `&--template` sets `--nimble-card-section-padding: 0` and `--nimble-card-section-border: none`, and the border is now a token so a section can opt out (`src/scss/components/_chat-card.scss:175`).
- **A refusal was silent** — `applyCasterConcentration` returns `applied` / `refused` / `skipped`, and `refused` warns on the same string a throw does (`src/documents/item/base.svelte.ts:261`).
- **Wand of Fly and Cloak of Lesser Windform applied nothing** — both now carry `concentration` in `properties.selected`, Wand of Fly's duration reads 10 minutes, and Migration062 ticks the property on copies already on a character.
- **Unguarded delete loop** — the replaced effects go out in one `deleteDocuments(ids, { parent })` call, so a refusal partway cannot leave the target holding neither (`src/utils/applyConditionToActor.ts:216`).
- **`replaces: []` turned off duplicate suppression** — the guard tests `options.replaces === undefined`, so an empty array keeps it (`src/utils/applyConditionToActor.ts:259`).
- **No automation gate** — gated on `isRuleAutomationEnabled()` (`src/documents/item/base.svelte.ts:55`); see Approach.
- **Inline array in pack JSON** — `master-of-storm.json` expands `schools` one entry per line.
- **Dead test parameters** — `disabled` and a failing `predicate` each get a negative case (`src/models/rules/concentrationTrack.test.ts:94,102`).
- **Migration matched on `source.name`** — the correction table is keyed by compendium source id, so a GM's own `Fly` is left alone and a translated world still matches (`src/migration/migrations/Migration062ConcentrationAppliesToCaster.ts:23`).
- **`SafeRestCard` scoped padding tied on specificity** — it sets `--nimble-card-section-padding` instead of redeclaring `padding` (`src/view/chat/SafeRestCard.svelte:115`).

## Not fixed

- **`CONTEXT.md` entry for "concentration track"** — there is no `CONTEXT.md` in the repo and no domain-term glossary anywhere else (`docs/PROJECT_CONTEXT.md` is tech stack and coding rules). The term is defined in `docs/documentation/playing/conditions.md` and the new reference partial, which is where this repo documents terms. Starting a glossary is its own change.

## Approach

The automation gate is `applyRuleEffects`, not `derivedConditions`. Derived conditions are the ones implied by *other conditions* (Hampered while Dazed); concentration-on-cast is driven by an item property at activation, which is what `applyRuleEffects` covers. Its setting hint now says so explicitly.

Track arithmetic moved out of `item/base.svelte.ts` into `src/documents/concentration.ts` so the cast path, the toggle override and the end-one prompt all read the same definition of what a track is and what a cast displaces.

Ending one of two tracks is a real choice, so it prompts rather than picking. Ending the only one still goes straight through.

Pushed as `90390d7`.

trevlar:
## Fixed
- **Master of Storm granted three concentrations** — a named school now claims the caster's single concentration instead of adding one: an unnamed school ends every track, a named school ends its own plus anything on the default track (`src/documents/concentration.ts:60`). Max concurrent is now the number of named schools.
- **Two tracks could not be ended one at a time** — `NimbleBaseActor#toggleStatusEffect` intercepts concentration and asks which to end when two are held, so core's delete-every-match loop is never reached (`src/documents/actor/base.svelte.ts:1171`, `src/documents/actor/endOneConcentration.ts:47`). One held concentration still ends without a prompt.
- **Refused create lost the held concentration** — a `null` effect restores the replaced sources and returns before the `conditionApplied` hook (`src/utils/applyConditionToActor.ts:262`). `applyCondition.test.ts` pinned the old `effect: null` fire; it now asserts the hook does not fire.
- **Missing `concentrationTrack` partial** — added, with the Master of Storm worked example and the corrected semantics (`docs/documentation/reference/_partials/concentrationTrack.md`).
- **Hoisted card-section rule reached TemplateSection** — the border is now `var(--nimble-card-section-border, ...)` and `--template` sets both it and the padding to nothing, restoring what the scoped blocks gave it (`src/scss/components/_chat-card.scss:174`).
- **Refusal showed no warning** — `applyCasterConcentration` returns `applied`/`refused`/`skipped`, and the call site warns on `refused` (`src/documents/item/base.svelte.ts:53`, `:301`).
- **Two concentration objects applied nothing** — Wand of Fly and Cloak of Lesser Windform now carry the property, Wand of Fly's duration is 10 minutes, and Migration062 ticks the property on existing copies (`packs/magicItems/core/wands/wand-of-fly.json:57`, `src/migration/migrations/Migration062ConcentrationAppliesToCaster.ts:106`).
- **Unguarded delete loop** — one `deleteDocuments(ids, { parent })` call instead of a delete each (`src/utils/applyConditionToActor.ts:204`).
- **No automation gate** — concentration goes through `isRuleAutomationEnabled()`, and that setting's hint now says so (`src/documents/item/base.svelte.ts:55`).
- **Inline array in pack JSON** — expanded (`packs/classFeatures/core/stormshifter/stormshifter-subclasses/circle-of-sky-and-storm/master-of-storm.json:18`).
- **Durations rewritten by name** — the fix table is keyed by compendium source id, so a GM's own `Fly` is untouched and a translated world still matches (`src/migration/migrations/Migration062ConcentrationAppliesToCaster.ts:23`).
- **Dead test parameters** — `disabled` and a failing predicate each have a negative test, following `speedBonus.test.ts` (`src/models/rules/concentrationTrack.test.ts:95`).
- **SafeRestCard specificity tie** — routed through `--nimble-card-section-padding` (`src/view/chat/SafeRestCard.svelte:114`).

## Not fixed
- **`replaces: []` turning off duplicate suppression** — taken as the documented behaviour rather than a bug: passing `replaces` at all hands the caller the dedupe decision, and an empty array means "I checked, nothing needs removing". The test is now `options.replaces === undefined`, the JSDoc says the empty array counts, and a named test pins that a caster with a multi-status concentration effect still gets the condition applied (`src/utils/applyConditionToActor.ts:60`).
- **`CONTEXT.md` entry for "concentration track"** — there is no `CONTEXT.md` in this repo; `docs/PROJECT_CONTEXT.md` is the agent-rules file and carries no glossary. The term is defined in `conditions.md` and the new reference partial instead. Happy to add a glossary if you want one.

## Approach
The track logic moved to `src/documents/concentration.ts` because the actor toggle and the item activation path both need it now. The rules reading behind finding 3: the rulebook grants one concentration at a time and Master of Storm grants exactly the lightning-and-wind pair, so lightning-plus-ice is not a thing the feature allows either; that is why a named-school cast also clears the default track. The integration test that asserted three concurrent concentrations now asserts one, and there is a new live test for a tracked school displacing an untracked one.

Pushed as 90390d7. Full `pnpm check` green (4670 tests).

Fronix (changes requested):
## Review

All round 2 threads are resolved and I confirmed each fix landed: a refused create restores the held effect, the replaced effects go in one `deleteDocuments` call, a named school claims the default slot, the partial exists, `TemplateSection` opts out of the hoisted rule, a refusal warns, both magic items carry the property, the automation gate is in place, the migration is keyed by source id, and `SafeRestCard` sets the variable. The branch merges clean onto the current dev tip and migration numbering is contiguous.

The findings below are new, or are places where a round 2 fix is incomplete.

### Blocking

**1. The live suite cannot pass.** `tests/integration/concentration.test.ts:37`

```ts
system: { school, tier: 1, properties: { selected: ['concentration'] }, ...
```

The caster is a bare `Actor.create({ type: 'character' })` with no class, so `highestUnlockedSpellTier` is 0. `ItemActivationManager.getData` (`src/managers/ItemActivationManager.ts:205`) refuses a tier 1 spell on every path, fastForward included, so every `castSpell` returns `null` and every `waitFor` times out. This is the same failure #983 fixed on dev this morning.

Fix: `tier: 0`, or give the caster a class that unlocks tier 1. Then run the suite; it has not been executed yet.

**2. The Master of Storm rule never reaches existing characters.** `packs/classFeatures/core/stormshifter/stormshifter-subclasses/circle-of-sky-and-storm/master-of-storm.json:9-22`

Migration062 handles only `spell` and `object` sources, and `8gTUauVQdm1Se1X6` is not in its table. A level 15 Stormshifter already in a world gets no tracks, while `conditions.md:34` tells that player it works. Migration061 (`:147-170`) is the house pattern for adding rules to imported features by source id.

**3. Engine comment names the motivating feature.** `src/documents/concentration.ts:55-57`

```ts
 * Master of Storm reads "you can concentrate on 1 lightning spell and 1 wind
 * spell at the same time", so naming a school splits the caster's single
```

Engine comments describe the mechanism, not the feature that asked for it. The rulebook example already lives in the partial. Same in `src/models/rules/concentrationTrack.test.ts:21`, where the fixture is named `Master of Storm`; sibling tests use `Test Item`.

**4. Shared test infrastructure modified.** `tests/mocks/foundry.ts:34-40`

AGENTS.md: "Don't modify test infrastructure: `tests/setup.ts` and `tests/mocks/foundry.js`". The `toggleStatusEffect` stub is load-bearing for `concentrationToggle.test.ts`. The `deleteEmbeddedDocuments` stub is used by no test; each test supplies its own `vi.fn()`. Drop the unused one, and stub the prototype from the test if you can.

**5. A Foundry global reached through `globalThis` with a dead guard.** `src/documents/actor/endOneConcentration.ts:23-26`

```ts
const resolve = (globalThis as Record<string, unknown>).fromUuidSync as ... | undefined;
if (typeof resolve !== 'function') return null;
```

`src/utils/resolveContentLinks.ts:15` and `markTargetEffects.ts:174` call `fromUuidSync(...)` directly. The guard exists for the mock, which the STYLE_GUIDE's YAGNI rule forbids.

### Should fix

**6. A failed restore is only logged.** `src/utils/applyConditionToActor.ts:230-239`. If the create failed for lack of permission, every restore fails for the same reason, and the caster ends with no concentration while the toast says only "could not be applied". Surface the loss.

**7. Immunity is reported as failure.** `src/documents/item/base.svelte.ts:69`. `return effect ? 'applied' : 'refused'` treats a `preApplyCondition` refusal from condition immunity as a failure, so a concentration-immune actor gets the warning on every cast.

**8. Dismissing the prompt returns `undefined`.** `src/documents/actor/base.svelte.ts:1184`. Core's contract (`client/documents/actor.mjs:566-578`) returns `true` when the status stays active and `undefined` only when asked to deactivate an absent status. The prompt path also drops `options.overlay`.

**9. The tracks `Set` is never reset.** `src/models/rules/concentrationTrack.ts:77-79` writes a `Set` onto `actor.system` outside `actorAccumulatorPaths`, so the reset loop at `src/documents/actor/base.svelte.ts:334-338` never clears it. `conditionImmunity.ts:69` is the precedent, but this value is read at cast time.

**10. Migration gaps.** `src/migration/migrations/Migration062ConcentrationAppliesToCaster.ts:106,129`. The duration fix is gated on the property, so a copy whose GM un-ticked Concentration keeps 1 minute. An object source with no `selected` array gets neither property nor duration. `duration.details` is written to the pack but not by the migration. The docblock at `:83-85` says homebrew copies are cleared, which is untrue for the two `addsProperty` items.

**11. Unreachable `details` value.** `packs/magicItems/core/wands/wand-of-fly.json:19` sets `details: "Up to 10 minutes"` on `type: "minute"`. `ItemActivationCoreConfigTab.svelte:167` shows the field only for `special`, so it cannot be seen or cleared. Copies the pre-existing cloak.

**12. Upcast indicator ties the hoisted rule.** `src/view/chat/SpellCard.svelte:140`. `.nimble-upcast-indicator { padding: 0.5rem }` on a `nimble-card-section` ties `.system-nimble .nimble-card-section` at (0,2,0). Same latent issue the round 2 `SafeRestCard` fix removed. Set `--nimble-card-section-padding`.

**13. Concentration module placement.** Round 1 asked for the one-caller utility to be inlined. It now has three callers (item, actor, rule) and lives half in `src/documents/concentration.ts` and half at module scope in `src/documents/item/base.svelte.ts:18-69`. AGENTS.md's promotion rule points at `src/utils/` plus a Shared Code Inventory row.

**14. Card indicator depends on the setting, not the property.** `src/documents/item/base.svelte.ts:55`. With Apply Conditions and Effects from Rules off, the outcome is `skipped` and a concentration spell's card shows nothing. #23 ties the indicator to the property. The docs say this is intended; please confirm it is.

### Nits

- `src/scss/components/_chat-card.scss:173-174`: the comment narrates the diff's history ("carried neither property while this block was scoped per card"). State the intent.
- `src/utils/concentrationPack.test.ts:44-49`: matches on the last id segment only, so the pack-name segment that decides a real match is unverified. The spell glob also skips the two `object` items.
- `src/models/rules/applyCondition.test.ts:504`: `not.toHaveBeenCalledWith(name, objectContaining(...))` passes on any payload shape change. Assert the hook was not called.
- The PR body is stale: it still names `Migration058`, `applyCasterConcentration.ts` as a utility, "12 new tests" and "LATEST_SCHEMA_VERSION bumped to 58".

### Rules

- Wand of Fly and Cloak of Lesser Windform appear in the books only as adventure props by name (`GMguide-2.0.2.md:4379`, `:2722`), with no stat block, duration or Concentration tag. Marking them Concentration / 10 minutes is inference from the imbued spell. Reasonable, but the PR body should say it is authored, not sourced.
- Master of Storm's wording is confirmed (`Heroes-2.0.2.md:1805`). "Never a third" and "an unnamed cast ends everything" are inferences from that line plus `CoreRules-2.0.2.md:352`. `conditions.md:34` and the partial state them as fact, firmer than the source; hedge them the way the replacement sentence is hedged.
- Gaze of Two Minds says "for as long as you hold concentration" (`Heroes-2.0.2.md:1258`). It is a `feature`, and `FeatureDataModel` has no properties field, so the property mechanism cannot reach it. Not this PR's defect, but worth a docs note or a follow-up issue.
- The docs omit that ending concentration is a free action, once per turn (`CoreRules-2.0.2.md:486`).

### Confirmed correct

- All 12 concentration spells and their durations match the books (`CoreRules-2.0.2.md:1620`, `GMguide-2.0.2.md:732,736,738,740`).
- One at a time (`CoreRules-2.0.2.md:352`), with replacement correctly labelled as the system's reading.
- Every acceptance criterion in #23 traces to code and holds at the edges: no target, ally target, scroll, homebrew property, second cast leaves one, no `condition: concentration` node left in packs, migration walks embedded items.
- No other feature in the five books changes concentration capacity. Artificer and Hexbinder never mention it.
- Changed unit tests: 8 files, 78 tests pass. `tsc --noEmit` is clean.

PR #979, opened 2026-09-20
fix(chat cards): show a damage roll once when the node has a target disposition

Fronix:
## Summary

A homebrew spell showed its damage roll two times on the chat card. The cause is the Target Disposition dropdown: it stores "any" as a real value, and any stored disposition made the damage node put itself on the card in addition to the outcome child that already carries the same roll. Reported on Discord.

## Changes

- A top-level damage node stands alone on the card only when no outcome child surfaces it for that context. This also removes the same duplicate for Friendly, Neutral, Hostile and Secret, and for a homebrew On Miss outcome.
- "Any" now reads as no disposition at all, which is what the config sheet means by it. New helper `hasDispositionTarget`.
- The outcome child inherits `targetDisposition`, so the Apply Damage recommend and discourage hint still knows who the roll is meant for.

Searing Light (disposition, no outcome child), Shadow Trap (Roll Damage From Card) and the miss card keep their present behaviour. Each is pinned by a test.

## Test Plan

Automated, at three seams. Revert `findNodesByContexts.ts` and all three go red:

- `src/view/dataPreparationHelpers/effectTree/getRelevantNodes.test.ts` (new): node shapes in and out.
- `src/documents/chatMessage.test.ts`: `effectNodes`, the list the card renders from.
- `src/view/chat/components/ItemCardEffects.test.ts`: counts the damage boxes in the rendered DOM.

By hand in Foundry:

1. Make a spell with one Damage effect, formula `1d6`, and the default On Hit -> Damage Outcome (Full Damage).
2. Set Target Disposition to Hostile, then back to Any. Before this change either choice is enough to break it.
3. Roll the spell. The card shows one damage box and one Apply Damage button.
4. Check Searing Light still shows its damage box, and a Shadow Trap card still shows its Roll Damage button.

Note for the reviewer: I have not run this in a live Foundry world. The verification above is unit and rendered-component only.

## Checklist

## Related Issues

None filed. Reported on Discord.

## Player and GM impact

When a damage node has a target disposition, the chat card no longer shows the same damage roll twice. `"Any"` acts as no target disposition.

When damage is deferred, the Roll Damage button remains visible until the roll resolves.

Apply Damage recommendations and hints keep the correct target disposition.

## Engine changes

- Added the generic `hasDispositionTarget` type predicate.
- Suppressed a top-level damage node when a matching `damageOutcome` child shows the roll.
- Propagated `targetDisposition` to `damageOutcome` nodes.
- Updated `ApplyDamageButton` logic and related type declarations.
- Added tests for node selection, chat message effect nodes, and rendered damage box counts.

No new or renumbered migrations, pack content, settings, rule types, or `en.json` keys changed.


trevlar (dismissed):
**🔧 Push back — deferred damage loses its Roll Damage button**

**Must fix before merge**

- **src/utils/treeManipulation/findNodesByContexts.ts:37** — A deferred damage node that has an outcome child now disappears from the card, button and all. `createDamageNode` gives every top-level damage node an `on.hit` Damage Outcome by default, and the Deferred Roll checkbox is offered on exactly those top-level nodes, so ticking it on a homebrew damage effect makes `surfacedByOutcome` true and drops the root. The outcome child that stands in has no roll yet, and `DamageRoll` renders nothing without `roll?.class`, so the damage becomes unrollable. Shadow Trap escapes only because Migration046 writes `on: {}`. I ran `getRelevantNodes` on the builder-default shape against both branches: on `dev` the root node survives, here only `root-damage-hit` does. Mirroring `DamageNode.svelte`'s own `awaitingRoll` test covers it — `const awaitingDeferredRoll = node.deferredRoll === true && !node.roll?.class;` then `if (awaitingDeferredRoll || (!surfacedByOutcome && standsAlone)) result.push(node);`. Suppressing it after the roll lands is right, since the outcome child inherits the roll by then.

- **src/view/dataPreparationHelpers/effectTree/getRelevantNodes.test.ts:132** — This is the one deferred shape that can't reach the new branch: `on: undefined` means `surfacedByOutcome` is false no matter what. Pin the shape the builder actually produces instead, `deferredRoll: true` with the default `on.hit` outcome child. It goes red on this commit.

- **src/view/chat/components/ApplyDamageButton.svelte:50** — `pnpm check:svelte` now fails here: `Type '"any"' is not assignable to type 'TargetDisposition | undefined'`. Widening the prop to `EffectNodeDisposition` introduced it, and `pnpm check` doesn't run `check:svelte`, which is why CI stayed green. Turning `hasDispositionTarget` into a type predicate (`node is T & { targetDisposition: Exclude<EffectNodeDisposition, 'any'> }`) makes the filter narrow and the error goes away. Fine as a follow-up PR if you'd rather keep this one to the fix.

- **src/view/dataPreparationHelpers/effectTree/getRelevantNodes.test.ts:89** — This comment repeats `findNodesByContexts.ts:25` nearly word for word, and the test name already carries it. Drop it.

**What's good**

- Three seams, and the source revert really does turn all three red. I checked.
- Reading `"any"` as no choice matches what the config sheet already does with `node.targetDisposition ?? 'any'`.

---

1. Keep the Roll Damage button on deferred damage
2. Pin the builder-default deferred shape in a test
3. Clear the new svelte-check error
4. Drop the duplicated comment in the test

trevlar (dismissed):
**🔧 Push back — an explicitly stored "Any" now loses the damage roll**

**Must fix before merge**

- **src/utils/treeManipulation/findNodesByContexts.ts:38** — A damage node with `targetDisposition: 'any'` and no `damageOutcome` child under the active context now puts nothing on the card. I probed both branches: `getRelevantNodes` on `{type:'damage', targetDisposition:'any', on: undefined}` with `['hit']` gives `['root-damage']` on `dev` and `[]` here. The dropdown stores the string, because the `onchange` at `ItemActivationEffectsConfigTab.svelte:239` writes `target.value` raw, so "Hostile, then back to Any" (step 2 of your own test plan) is enough to reach it. Searing Light is that exact shape with `hostile`, and your test at line 78 says it has to keep showing, so the same shape with Any shouldn't disappear. Put `node.targetDisposition != null` back in `standsAlone` and keep `hasDispositionTarget` where it earns its place, on `ApplyDamageButton.svelte:21`. `surfacedByOutcome` is what fixes the duplicate, so nothing is given up.

- **src/view/dataPreparationHelpers/effectTree/getRelevantNodes.test.ts:46** — This passes whether or not `'any'` reads as no disposition. I swapped line 38 to `node.targetDisposition != null` and ran the five files around this change: 198 green, this one among them. `surfacedByOutcome` already catches the duplicate it's aiming at, which leaves the `'any'` behaviour with no coverage at all. Add the case above, where nothing else surfaces the roll.

- **src/documents/chatMessage.ts:1230** — This comment says disposition-targeted damage nodes are surfaced alongside their own outcome children, which is the thing you just stopped happening. The `surfacedOutcomeParentIds` guard under it is unreachable now as well: the one shape that still puts both on the card is deferred-and-unrolled, and there neither node has a `roll.class` to get past line 1247. It's pre-existing, from #847. Delete both, or the next reader will believe the duplicate is still live. Fine as a follow-up PR.

- **src/view/dataPreparationHelpers/effectTree/getRelevantNodes.test.ts:96** — `conditionType` isn't a field on `ConditionNode`. It's `condition` (`types/effectTree.d.ts:56`), and this is the only place in the repo spelling it the other way. Use the real name and the `as unknown as EffectNode` on line 99 can go.

- **src/view/dataPreparationHelpers/effectTree/getRelevantNodes.test.ts:135** — `toContain` hides that the result is `['root-damage', 'root-damage-hit']`, where every sibling test uses `toEqual`. The render is fine, I checked it: one Slashing box, one Roll Damage button. Asserting both ids would show a reader why there are two.

- **src/utils/treeManipulation/findNodesByContexts.ts:34** — "deferred and not yet rolled" now sits in three files, here plus `DamageNode.svelte:26` and `buildDeferredDamagePatch.ts:22`. That's the third file, which is where AGENTS.md's Rule of Three bites, and your comment on line 32 already names the drift. A one-liner beside `hasDispositionTarget.ts` would hold all three.

- **docs/documentation/homebrew/activations.md:41** — This lists the dropdown as "Any / Friendly / Neutral / Hostile / Secret" without saying what Any does, and you've just given it a meaning a homebrewer can act on. One sentence, per CLAUDE.md's rule on homebrewer-facing changes.

- **src/view/chat/components/HealingNode.svelte:37** — The cast asserts a type that excludes `'any'`, which the sheet can store. It lands on 'neutral' either way so nothing's broken, but `hasDispositionTarget` now exists to say it properly.

**What's good**

- All four items from last round landed, and I re-ran the one I was least sure of: `check:svelte` reports nothing in any file you touched. The 440 errors it prints are spread across 72 files that predate this branch.
- No shipped content changes behaviour. Searing Light and Shadow Trap are the only pack damage nodes carrying a disposition or a deferred flag, and neither has an outcome child, so both still surface. This reaches homebrew only.

---

1. Let a stored "Any" keep a bare damage node on the card
2. Cover the "Any" path with a test that can fail
3. Drop the stale duplicate-damage guard in chatMessage
4. Use `condition`, not `conditionType`, in the fixture
5. Assert both ids in the deferred test
6. Extract the awaiting-deferred-roll check
7. Say what Any means in the activations docs
8. Narrow HealingNode's disposition with the new helper

Fronix:
All eight items are in, pushed as `f8090ef6`, `11ccdbc9`, `ae48d017` and `092f75a1`.

**1 and 2, the blocking one.** `standsAlone` reads `node.targetDisposition != null` again, so a stored Any keeps a bare node on the card, and `hasDispositionTarget` now lives only where it earns its place: the Apply Damage hint and, per item 8, the healing node. New test `still shows a bare damage node whose Target Disposition is Any` goes red on the previous commit.

Your finding was wider than stated. I built a differential grid this time: every combination of disposition (absent, `any`, `hostile`), deferred flag, rolled or not, eight child layouts and three card contexts, 288 shapes, through `origin/dev`'s copy of `findNodesByContexts` and this branch's. Before the fix 35 cells differed: 16 lost the roll (your bare-`any` case plus the condition-only, `on: {}` and miss-outcome-on-hit siblings), and 19 were non-deferred nodes with no roll, which `ItemActivationManager.ts:419` rules out on a real card. After the fix: 0 regressions, 46 intended, 19 unreachable, and every intended cell is the root dropped in favour of an outcome child carrying the same roll. The grid and the dev snapshot are deleted, not shipped.

**3.** Guard and comment removed. The existing test that pins 42 for a disposition-targeted node with a child still passes, so the dedupe now happens upstream of it.

**4.** `condition`, and the cast is gone.

**5.** `toEqual(['root-damage', 'root-damage-hit'])`, with a line saying why both are there.

**6.** `isAwaitingDeferredRoll` in `src/utils/treeManipulation/`, used by `findNodesByContexts`, `DamageNode.svelte` and `buildDeferredDamagePatch`.

**7.** One sentence in `docs/documentation/homebrew/activations.md`: Any means no preference, the Apply Damage button gets no hint.

**8.** `HealingNode.svelte` narrows with the helper instead of the cast.

Gates: 4635 tests, `tsc` clean, no dependency violations, `check:svelte` still 440 with nothing in a touched file.

One thing the grid surfaced that is pre-existing and out of scope here: a Damage Outcome under both On Critical Hit and On Hit applies the roll two times on a crit. Filed as #980 with a card-seam repro.

trevlar (dismissed):
**🔧 Push back — dropping the `chatMessage` guard doubles damage one level down**

**Must fix before merge**

- **src/documents/chatMessage.ts:1230** — Item 3 from last round was my mistake. That guard wasn't stale, and without it this loop counts a nested damage packet twice. `surfacedByOutcome` lives inside the `!node.parentNode` branch at `findNodesByContexts.ts:21`, so it never sees a damage node hanging under another damage node's On Hit. Build Damage → On Hit → Damage → On Hit → Damage Outcome and `findNodesByContexts` returns `['nested', 'nested-hit']`; `processNodes` copies the parent roll onto the child, so both carry it. I ran that shape through this branch: a `1d4` that rolled 3 produces two entries of 3 and applies 6. On `dev` the guard skipped `nested` and applied 3. It reaches real HP through `#applicableDamagePackets`, and it inflates the `getDamageBreakdownForTarget` preview the same way. The builder allows every step — `getNodeOptions` offers Damage under a damage node, and `getValidActionConsequences` gives the nested node its own On Hit bucket — and `activations.md:53` documents the shape as supported. No shipped pack content has it, so this is homebrew only. Either put the guard back, or lift `surfacedByOutcome` out of the root-only branch so the child push at `findNodesByContexts.ts:51` skips a damage child that already has an outcome child for the context. Whichever you pick, pin it: nothing today tests a nested damage node.

- **src/view/dataPreparationHelpers/effectTree/getRelevantNodes.test.ts:143** — "The child has no roll yet and so draws nothing" is the opposite of what happens. `DamageOutcomeNode.svelte:11` hands the child straight to `DamageRoll`, whose `{:else}` branch at `DamageRoll.svelte:67` renders a `RollSummary` with `total = Math.ceil((roll?.total ?? 0) * multiplier)`. I rendered an unrolled deferred node with its default On Hit child and got a `0 Necrotic` box beside the Roll Damage button. That's the placeholder `DamageNode.svelte:22-26` says must not appear, and every deferred node built in the UI hits it, because `createDamageNode.ts:19-23` always gives a root node that child. The bug predates you and I'd leave it to a follow-up, but the comment is new and says the reverse, so drop it and pin what the card really draws.

- **docs/documentation/homebrew/activations.md:41** — The new sentence scopes Any to the Apply Damage hint, but `findNodesByContexts.ts:39` reads `!= null`, so picking Any is also what keeps a damage entry with no outcome child on the card. Two homebrewers who both mean "no preference" get different cards depending on whether they opened the dropdown. That's the behaviour I asked for last round, so I'm not asking you to change it, just to say it here.

**What's good**

- The three seams do what the description says. `getRelevantNodes.test.ts:52` goes red on the previous commit, and the card-level and DOM-level tests pin the same claim where a regression would actually surface.
- Splitting the crit double-outcome into #980 and stacking #981 on this branch, rather than letting this PR grow to cover it.

---

1. Stop double-applying nested damage after the guard removal
2. Add a test for the nested damage shape
3. Fix the wrong comment about the unrolled deferred child
4. Document that Any also keeps a bare damage entry on the card

Fronix:
All four confirmed, and all four fixed. Thanks for catching the nested case, that one was removing real HP.

I reproduced every claim before changing anything, at unit level, differentially against `dev`, and live on a v14 world.

## 1. Nested damage double-applies

Confirmed, exactly as you described. Live, activating a `Damage(1d8) -> On Hit -> Damage("3", necrotic) -> On Hit -> Damage Outcome` feature against a linked token and clicking the card's real Apply Damage button:

```
surfaced: ["nested","nested-hit"]   card totals: ["3","3"]   breakdown total: 6   hp lost: 6
```

I pinned causation rather than inferring it: putting the guard back, rebuilding and relaunching the world gave `hp lost: 3` on the same feature. The unit-level differential agrees, 6 on this branch and 3 on `dev`.

I took your second option rather than restoring the guard, because **restoring the guard fixes the total but not the card**. With the guard in place the card still drew two `3` boxes, and it does the same on `dev` today. The guard only ever ran at the point of applying damage. `findNodesByContexts` now applies the "an outcome child already carries this roll" rule to a nested node too, through a shared `needsItsOwnEntry` helper, so the roll is drawn once and applied once.

After the fix, live:

```
surfaced: ["nested-hit"]   card totals: ["3"]   hp lost: 3
```

One correction to the severity note, in your favour. `packs/spells/core/ice/shatter.json` does nest a damage node under `criticalHit`, so shipped content is not entirely free of the shape. It carries `on: {}` with no outcome child, so it produces one entry and never hit the bug. I added it as a test case in both directions, and verified live that the shape still shows and applies both packets (4 + 5 = 9 HP).

Because this is a pure function with two boolean flags, I ran a differential grid before pushing: 18 tree shapes x 4 context sets x both flags, through the `dev` copy, the branch tip copy and the fixed copy. Every difference against the tip falls in the two nested-with-outcome shapes. Nothing else moved.

## 2. A test for the nested shape

Added at all three seams, since a regression surfaces differently at each:

- `findNodesByContexts.test.ts` for node selection
- `chatMessage.test.ts` for the packets that reach real HP
- `ItemCardEffects.test.ts` for the boxes drawn

Both bug-pinning tests go red on the previous commit and green on the fix; I checked that explicitly rather than assuming. Five more pin the shapes the fix must not break: a nested node with no outcome child, an unrolled deferred nested node, `Any`, Hostile, and base damage nodes on a miss.

## 3. The wrong comment

Confirmed. Live, from a real deferred card:

```
{"totals":["0"],"labels":["Necrotic"],"hasRollButton":true}
```

Comment replaced with what actually happens, and pinned with a DOM test that asserts the `0` box beside the Roll Damage button. It is recorded as the current behaviour, not the intent, so it fails when someone fixes the placeholder, which is the right moment to rewrite the comment again. Left the underlying defect to a follow-up as you suggested.

## 4. Documenting Any

Confirmed and documented. A bare damage entry with no disposition is dropped on a hit; the same entry with `Any` is kept. The page now says so, and also states the nested rule: a damage entry inside an outcome bucket shows its roll once, through its own Damage Outcome, the same as at the root.

## Unrelated, found while verifying

`tests/integration/chatCards.test.ts` has been red on `dev` since 2026-09-12 and is nothing to do with this branch. It picks the first non-AoE spell in the compendium index, which is now Living Inferno at tier 7, and the blank test character cannot cast above tier 0, so no card is posted. Filtering to tier 0 alone is not enough either, because a blank character has no `@key`. Fixed separately in #983.

## Verification

Unit 4641 passed, type-check clean, no circular deps, svelte-check unchanged at its existing baseline. Live: 30 files, 161 tests, all green on #983's branch, which is the first time that suite has been fully green in a while.

Three commits: the fix and its tests, the deferred-placeholder pin, and the docs.

Note that #981 is stacked on this branch and touches the same five files, so it will need a cascade with conflicts before it can merge.

trevlar (approved):
**✅ Approved**

All four items from last round landed, and I checked each rather than taking the description's word for it. Reverting the line-63 skip turns three tests red across three files, so the nested fix is pinned where a regression would actually show up. Full suite 4641 green, `tsc` clean, and `svelte-check` reports nothing in any file you touched; its 440 errors sit in 72 files that predate this branch.

This is an approval after three rounds of push-back, so to be clear about what turned it: fixing the nested double-apply at node selection rather than at application fixes the card as well as the total. Restoring the guard would have fixed only the total.

**Worth fixing (non-blocking)** — none of these gate the merge.

- **src/utils/treeManipulation/findNodesByContexts.ts:63** — The skip also runs on a `savingThrow`'s bucket children, but `traverse` only descends into `on` children when `node.type === 'damage'` (line 77). Under a saving throw the skip can only lose a roll, never dedupe one. `savingThrow.on.failedSave = [damage{ on.failedSave: [damageOutcome] }]` with contexts `['failedSave']` returns `['save']` here and `['save', 'dmg']` on `dev`; the roll disappears, outcome child and all. Nothing builds that shape today, since `getValidActionConsequences` gives a damage node with `parentContext: 'failedSave'` only the hit/miss/criticalHit buckets, so no content moves. Adding `node.type === 'damage' &&` to the condition makes the skip match where the traversal actually goes.

- **src/documents/chatMessage.test.ts:1507** — The root node in this fixture has no outcome child under On Hit, so its own `1d6` never reaches the card and the total is 3 instead of 8. `createDamageNode.ts:19-23` gives every root damage node a default On Hit outcome, so the builder can't produce this tree. Give `root` its own outcome child and the assertion also pins that the root is counted exactly once, which is the half a regression is most likely to break. Same fixture at `ItemCardEffects.test.ts:283`.

- **src/utils/treeManipulation/findNodesByContexts.test.ts:117** — Nothing covers the crit path, and it's the only caller that passes two contexts into `isSurfacedByOutcomeChild`'s `contexts.some(...)`. It's a behaviour change you're making, too: `root{ on.hit: [outcome], on.criticalHit: [damage with its own outcome] }` gives `['crit-bonus', 'root-hit', 'crit-bonus-hit']` on `dev` and `['root-hit', 'crit-bonus-hit']` here. Worth a case beside the nested ones.

- **src/view/dataPreparationHelpers/effectTree/getRelevantNodes.test.ts:127** — The literal above already matches `TextNode`, so `as unknown as EffectNode` buys nothing; `tsc --noEmit` is clean without it. Same cast you dropped from the condition fixture two rounds ago.

**Follow-up, not this PR**

- **src/utils/treeManipulation/findNodesByContexts.ts:77** — Pre-existing and identical on `dev`, so nothing here caused it. This loop recurses into every `on` bucket while line 58 pushes only for the active contexts, so a damage node under On Critical Hit that carries its own Damage Outcome surfaces that outcome on an ordinary hit card. `ItemActivationManager.ts:415` rolls the whole flattened tree, so the roll is real, `processNodes` copies it onto the child, and it applies. Worth an issue next to #980.

**What's good**

- The three seams do what you say they do. I reverted line 63 and got one failure per seam, in three separate files.
- `shatter.json` is the only shipped content with the nested shape, and you pinned it in both directions rather than asserting it was safe.

